PR #2, opened 2015-05-18
Copy instructions from make.py and format .md

daleathan:


Nick-Hall:
@dsblank Doug, Is the make.py run from this repository now?  Should the make.py examples be in the README for the addons repository instead?

bpisoj:
As i see it and use it!
Addons repo is for downloads only. Source code is on addons-source. You change code there and build it there but result (tgz and listing) ends up in addons repo. So we have two group of plugins developers: one which can change existing one or upload new plugins and second one which have power to release them (push builds to addons repo)

bpisoj:
I would like it prepend with something like:


Source code for Third-party plugins for Gramps genealogy program
If You are interested in changing, enhancing or creating new plugins , You are welcome to fork this repository and submits your changes thru pull-request.

To have write access to this repository you must be member of Gramps project Addon Developers team.
...above instructions...

dsblank:
I think we need to split make.py up into two parts: functions for transaltions, and functions for packaging. I haven't thought too much about that yet. But it sounds like these suggestions all make sense.

PR #7, opened 2015-08-12
Resolve nosetests conflicts in DynamicWeb report

belissent:
This PR should allow to pass the CI tests.
All the DynamicWeb tests were excluded from the nostests:
The example reports generation does not work because it seems that in the Travis test environment, the addons are not rebuild and not loaded in the USER_PLUGINS directory.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,4 +30,4 @@ install:
 
 script:
     -  mkdir -p /home/travis/.gramps/grampsdb/
-    -  GRAMPS_RESOURCES=../gramps PYTHONPATH=../gramps nosetests3 -vv --exclude=DynamicWeb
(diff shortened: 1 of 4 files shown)


belissent:
I don't understand why the checks have failed.
The failure is inside "test_dynamicweb.py", which has been renamed "run_dynamicweb.py"

dsblank:
Looks like:

```python
from dynamicweb import *
```

is failing. Perhaps you need to add the parent directory to sys.path, or do a relative import?

belissent:
Thanks for fast answer (as always).
I agree with you for relative import, but the file "test_dynamicweb.py" is discarded. So how could it raise an error?
I ran nosetests3 in my environment before committing, result was OK (no tests performed for DynamicWeb).

dsblank:
Looks good.... I suspect that it may have been a Travis/PR issue. Thanks for this!

PR #8, opened 2015-08-13
DynamicWeb: JSON used internally + sort objects + nosetests

belissent:
Thanks to the .travis.yml update, my addon tests pass:
https://travis-ci.org/belissent/addons-source/builds/75473784


dsblank:
Oh, another difference in Travis testing: there is no DISPLAY set, and so we can't import Gtk.

Can you try to conditionally import, or even better would be to fix gramps master so that we can use this functionality without trying to import Gtk.

dsblank:
Let me see if I can move thumbnails around a bit...

belissent:
Is it completely impossible to have a display in Travis?
With the following configuration in .travis.yml, I managed to get rid of Gtk errors:

```
install:
    - travis_retry sudo apt-get install xvfb libgtk-3-dev

before_script:
    - export DISPLAY=:0
    - sudo Xvfb :0 -ac -screen 0 1024x768x24 &
```
(tried in my fork: https://travis-ci.org/belissent/addons-source/builds/75515109)

dsblank:
Yes, that is one way. But, that doesn't help solve the issue that a DISPLAY is needed, even when running command-line code. I have a different solution... just committed. Let's see if this works.

dsblank:
BTW, your method will be useful when we want to start testing the GUI. We can have a special section for that, too. I think that there are libraries for clicking on Gtk things from Python...

dsblank:
Getting a strange error: https://travis-ci.org/gramps-project/addons-source

belissent:
This is due to the /home/travis subdirectories.
On my side, the tests now pass with Gtk.
I will propose you another PR

dsblank:
Ok, got it, and we can avoid the DISPLAY issues for the time being.

PR #10, opened 2015-08-13
Add display feature in Travis CI

belissent:


.travis.yml
```diff
@@ -25,12 +27,12 @@ install:
     - cd addons-source
     - python3 -m compileall *
 
-#before_script:
-#    - sudo Xvfb :99 -ac &
-#    - export DISPLAY=:99
+before_script:
```
dsblank: Does it work without the xvfb trick? I'd rather not put that in yet, if we don't need it.

dsblank:
We'll add the X display stuff when we have to.

PR #12, opened 2015-08-19
A few DynamicWeb report improvements

belissent:
There are a lot of files because the Bootstrap CSS and JS are now built from Bootstrap source.
This eases the tedious task of building the styles templates.

This CSS build is *NOT* to be included in the GRAMPS build.
All the files are pre-built before committing.


sam-m888:
Great report :+1:

belissent:
Thanks!
Do you have any improvement idea / suggestion?

PR #19, opened 2015-10-08
DynamicWeb: Update German translation

sebschub:
Small update.


romjerome:
Thanks.

New version on gramps42 and master.

PR #21, opened 2015-11-29
Make sure the surname handle is unique. Connect surname to name through link table

killes:
I tested this with the gramps example data and the output appears to be correct.

Note that the Import also needs to be adjusted.

### DIFF
--- a/Sqlite/ImportSql.py
+++ b/Sqlite/ImportSql.py
@@ -251,7 +251,7 @@ class SQLReader(object):
         return [self.pack_media_ref(sql, result) for result in results]
 
     def get_surname_list(self, sql, handle):
-        results = sql.query("""select * from surname where handle = ?;""", handle)
+        results = sql.query("""select s.* from surname s inner join link l ON l.to_handle = s.handle where l.from_handle = ?;""", handle)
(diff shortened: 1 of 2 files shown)


killes:
I've updated the PR to include a change to ImportSql.py that let me re-import the created database.

What would be the best tool to compare the two trees?

Sqlite/ImportSql.py
```diff
@@ -251,7 +251,7 @@ def get_media_list(self, sql, from_type, from_handle):
         return [self.pack_media_ref(sql, result) for result in results]
 
     def get_surname_list(self, sql, handle):
-        results = sql.query("""select * from surname where handle = ?;""", handle)
+        results = sql.query("""select s.* from surname s inner join link l ON l.to_handle = s.handle where l.from_handle = ?;""", handle)
```
dsblank: Do these need to be ordered? That is, will they be imported/exported and assembled in the correct order?

killes: I don't know :(
I guess we could order by primary?

dsblank: I think you can use `.get_links()` instead of the join. `links` should probably have an explicit ordering. But that can be a separate fix. Thanks!

PR #29, opened 2016-04-29
Ancestor fill

grocanar:
I just updated my ancestorFill Addon.
i make some cleanup in the code and i made an explanation in french of what i did.
if it necessary i could try to translate it in english.


sam-m888:
I believe you've committed the wrong files? Both GedcomforGeneanet and  AncestorFill are in this Pull Reqest.

grocanar:
Hi Sam


Indeed i m still new at git.

i wasn't aware i was including my previous modifications.

I guess i should have create a branch before making my new modifications.

I m wondering now what is the better thing to do?

sam-m888:
@grocanar I suggest closing this pull request. Creating a new branch and resubmitting.

grocanar:
after reading more on git i agree with you

i will do that

PR #31, opened 2016-05-20
GraphVeiw: fix 0009460 issue. Multiple spouses.

vantu5z:
After changes in "build_graph" definition of "self.person_handles = set()", we can't Include spouses from other marriage not selected by filter in "__add_family".
And I try to find all spouses in "add_descendant" function. I don't know if it's right but It works for me.

### DIFF
--- a/GraphView/graphview.py
+++ b/GraphView/graphview.py
@@ -1179,6 +1179,38 @@ class DotGenerator(object):
                             num_generations - 1,
                             person_handles)
 
+                # add all his spouses recursively
+                sp_person = self.database.get_person_from_handle(father_handle)
+                if sp_person:
+                  for sp_family_handle in sp_person.get_family_handle_list():
+                     sp_family = self.database.get_family_from_handle(sp_family_handle)
+                     if sp_family.get_mother_handle() and sp_family.get_mother_handle() not in person_handles:
+                        self.add_descendant(
+                          self.database.get_person_from_handle(sp_family.get_mother_handle()),
+                          1,
+                          person_handles)
+                     if sp_family.get_father_handle() and sp_family.get_father_handle() not in person_handles:
+                        self.add_descendant(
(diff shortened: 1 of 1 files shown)


tecknicaltom:
In the bug in Mantis you say you're using GraphView ver: 1.0.55. Have you tried to reproduce this issue with the newest version of the graphview source in git? https://github.com/gramps-project/addons-source/blob/master/GraphView/graphview.py I'm not sure but this may have been fixed after 1.0.55 was released.
If the version from git still exhibits the same problem, could you please describe the scenario in a bit more detail? I'm not claiming my changes have no bugs, I'm just trying to understand what I may have messed up :) Thanks!

vantu5z:
I use version from git, where my last localisation  commit.
You change type of 'self.person_handles' from 'list' to 'set' and when we try add spouses in "__add_family" that don't add in "add_descendant" python give error that 'self.person_handles' changed while iteration.
To reproduce just make second family with one father or mother and click on it in graphview. Or make it active at people view and you get an error what I mentioned.

vantu5z:
F - father, M - mother, fam - family 

M1 - - -fam1
F - - - {
M2 - - - fam2

If click on M1 or M2 you get the error.

vantu5z:
Oops,I did it just for 'fmm..' it needs to add for 'mff..'.
Hm.. what happens if 'fmmfm..' or something?

tecknicaltom:
oh, that's the source of my confusion. it looks like I already fixed this issue and forgot to do a pull request. I'll make a pull request of my fix, and you can look it over and see if you like your solution or mine better.

tecknicaltom:
Pull request for fix referenced in previous comment: https://github.com/gramps-project/addons-source/pull/32

sam-m888:
Thank you

sam-m888:
Not sure which one of these broke the addon will investigate over weekend!

PR #33, opened 2016-05-27
Fix 6816 GraphView: dash lines

vantu5z:


sam-m888:
Thank you

sam-m888:
Not sure which one of these broke the addon will investigate over weekend!

vantu5z:
Sorry, I didn't make it right. It include not needed commits. If click on changes in this PR - only that diffs should be added.
And PR #32 need to be revert. Because it was a one of solution for fix a bug.

PR #34, opened 2016-05-30
0009482: [Related Relatives Gramplet] duplicate titles

ennoborg:
This is my 1st pull request. The only change is in the Dutch title, but when I build, I also see a new version number being generated. Does that mean that I must include the new gpr.py file too? Is the new version number needed to ensure that users see the new Gramplet?


sam-m888:
Thank you

PR #36, opened 2016-08-11
0009627: DynamicWeb: Google map API now requires a key

phejl:
Fix for https://gramps-project.org/bugs/view.php?id=9627


belissent:
Thanks a lot for this correction.
Do you think that we should provide a default valid key in order to pass the tests, and easily generate examples ?

phejl:
I think it would be useful. Though I would not share it with wider audience. For tests it might be risky to push the "localhost" key as it may be abused by others. For examples (I suppose the deployed ones) it should be ok as far as the key usage is limited to your domain. I'm pretty new to gramps development (long term user) so I'm not sure if it is doable.

phejl:
You are right. Fixed in 94c986e17dd65bd338ec3c265b2fe74005dc30ba.

PR #40, opened 2016-08-17
[GraphView] Improvements

vantu5z:
- fix non handle error
- fix flicking when drag tree with zoom less then 0,5


vantu5z:
It's already fixed in commit 396df8f49f644c8a03421bf81abf8dab9534afef

PR #42, opened 2016-08-22
DynamicWeb localisation update for Czech

phejl:


belissent:
Thanks you for the update

PR #43, opened 2016-08-22
DynamicWeb: Update german translation

sebschub:


belissent:
Thanks for the update!

PR #44, opened 2016-08-22
DynamicWeb: Update german translation

sebschub:
For gramps42. Equal to the changes in master.


belissent:
Thanks for the update!

PR #52, opened 2016-09-05
[GraphView] fix flicking when drag[Gramps51]

vantu5z:
Seems that when calculates shift of dragged canvas scale is not taken into account.
And so, when scale less than 0,5 - I have flicking effect when dragging.
PR fix this behavior.


enlite93:
Don't see this on Windows 10, is this specific to linux?

macolquhoun:
Have you only tested this on linux?

vantu5z:
Yes, maybe this is linux specific. Anyway I have this bug in master (or gramps50) on fresh Arch Linux with Gnome.
@macolquhoun , can you test this patch on Windows (may be it broke something there)?

sam-m888:
Closed as made part of PR #88

PR #55, opened 2016-11-08
Remove Right-click menu icons etc.

prculley:
This is a continuation of the code sweep done in Gramps main code [PR268](https://github.com/gramps-project/gramps/pull/268),  Removed right-click menu icons.

While debugging, I noted several other issues:
1) added parent=self.uistate.window to several popup dialogs
2) in HtreePedigreeView, the `window.get_hadjustment()` was pointing to the wrong level of the Gtk hierarchy, a ViewPort rather than the ScrolledWindow.
3) in HtreePedigreeView, the window drag cursor was using deprecated syntax;
`Gdk.Cursor.new(Gdk.CursorType.FLEUR)` -> 
`Gdk.Cursor.new_for_display(Gdk.Display.get_default(), Gdk.CursorType.FLEUR)`
4) removed deprecated menu titles
5) a few pylint items (mostly trailing whitespace)
6) in HtreePedigreeView, an attempt to change the config options slider for number of generations didn't work because the slider was too small for the control to move.  I increased the size of the slider.
7) a fault (crash) when trying to use the phototagging gramplet with no photo selected
8) When comparing TimeLinePedigreeView with PedigreeView, I noted that the menu text was different for "chistianized" -> "christened"
9) in TimeLinePedigreeView, the cgi.escape was deprecated; I change to html.escape (like PedigreeView)
10) in TimeLinePedigreeView, several place positional arguments to Gtk methods were deprecated, replaced with labeled arguments.
11) When comparing TimeLinePedigreeView with PedigreeView, i noted that we should use the the is_right_click() method, so I changed TimeLinePedigreeView to match.
12) When comparing TimeLinePedigreeView with PedigreeView, i noted that `label.set_alignment(0, 0)` was deprecated and changed to `label.set_halign(Gtk.Align.START)`
13) in TimeLinePedigreeView, I was unable to get a block of code to be executed; I determined that the code could never be executed so commented it out.
14) in TimeLinePedigreeView, I changed a large space in the menu to a seperator.


prculley:
Looks like Travis script has a bug when I try to use a local branch (rightclick) off of gramps-project/addons-source...

sam-m888:
The Travis script for addons I believe has to be updated to be inline with the script from the main Gramps repository after @kulath updated it a few months back?

Thanks for updating all these :+1:    

I'll rebuild the addons.

Testing took a while :)

prculley:
Thanks for the testing.  And the assistance in merging and Bug tracker work.  I was unsure about building the addons; did not know if we had some sort of automated script that might do a build based on a commit...  Or perhaps they were rebuilt when we do a release Or...

PR #60, opened 2017-01-05
Language-sensitive Type and priority

GaryGriffin:
Replaced Type with language-sensitive Type. Prioritized Place Name to be in correct language

### DIFF
--- a/GetGOV/getgov.py
+++ b/GetGOV/getgov.py
@@ -330,7 +330,6 @@ class GetGOV(Gramplet):
         visited = {}
 
         type_dic = dict()
-        
         type_url = 'http://gov.genealogy.net/types.owl/'
         response = urlopen(type_url)
         data = response.read()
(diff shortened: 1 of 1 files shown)


sam-m888:
[Travis build failed](https://travis-ci.org/gramps-project/addons-source/builds/189053167#L2080) due to:

ERROR - Trailing whitespace found in source file(s) 

On line 333 and 443, resubmit with trailing whitespace removed please.

GetGOV/getgov.py:333:        
GetGOV/getgov.py:443:                    place_type.set_from_xml_str(type_dic.get(tuple([type_code,'en']),'No Type'))

GaryGriffin:
fixed blank space

sam-m888:
Thank you.

PS:GetGOV addon rebuilt for Gramps master as https://github.com/gramps-project/addons-source/commit/a84124f782a17954dc846a8311102d4136fa2ca6

GaryGriffin:
Is it possible to also back-port to the 4.2 tree? I would like to write the documentation but I don’t think I can provide the link if it is on master (5.0).

Gary

sam-m888:
@GaryGriffin 
> Is it possible to also back-port to the 4.2 tree?

You can give it a try ( create it on the [maintenance/gramps42](https://github.com/gramps-project/addons-source/tree/maintenance/gramps42)  branch)

>  I would like to write the documentation but I don’t think I can provide the link if it is on master (5.0).

Yes you can as it is an addon which is listed in a different part of the wiki than the user manual.

You may have to create a wiki page for the addon  (Assuming you already have a wiki account?)

The addon will also need a (Help) button added (see one of the other addons) that goes to the link on the wiki.

PR #65, opened 2017-03-14
Fixed bugs in center_person routine and graph URL.

zorkmark:
NetworkChart ver_0.0.3 release notes.

Fixed bug in center_person routine
Modified graph URL to locale URL to be Wiki Image security compliant
Switched order of trim_descendants and trim_ancestors with trim_groups
Modified marriage round to decade to round to year
Modified birth round to decade to round to year
Modified change middle names to initials to remove middle names
Modified such that databases with more than 1500 individuals will always have the following defaults (but can be changed): 
   Graph Style = Curved
   Rank Direction = Left to Right
   Rank Separation = 3
Modified such that databases with more than 5000 individuals will have the following initial defaults:
   Center Person = First Person in Database
   Highlight Center Person = True
   Use Center Person = True
   Radius = 10
Different values can still be selected if desired and will be remembered after the initial run.

### DIFF
--- a/NetworkChart/NetworkChart.gpr.py
+++ b/NetworkChart/NetworkChart.gpr.py
@@ -17,11 +17,13 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, 
 # Boston, MA 02110-1301  USA
 #
-# version 0.0.2
+# version 0.0.3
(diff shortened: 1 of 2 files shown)


zorkmark:
All taken care of in PR66.

PR #66, opened 2017-03-19
Added dependency dialog with link to wiki

zorkmark:
NetworkChart ver_0.0.4 release notes.

Added Dialog box to gpr.py file to alert user to missing python modules.  
Link is also provided in dialog to plugin wiki website for instructions.
Fixed highlight bug related to center_person.
All fixes from 0.0.3.

### DIFF
--- a/NetworkChart/NetworkChart.gpr.py
+++ b/NetworkChart/NetworkChart.gpr.py
@@ -17,11 +17,13 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, 
 # Boston, MA 02110-1301  USA
 #
-# version 0.0.2
+# version 0.0.4
 #-------------------------------------------------------------------------
 import importlib
-module = importlib.find_loader("networkx")
-if module is not None:
+from gramps.gen.const import URL_WIKISTRING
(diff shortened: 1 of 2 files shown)


uxcq:
> All fixes from 0.0.3.

So ignore [PR65](https://github.com/gramps-project/addons-source/pull/65) ? If so you may want to close it so it does not get applied.

sam-m888:
> So ignore PR65 ? If so you may want to close it so it does not get applied.

@zorkmark  Good question?

Paul-Franklin:
https://gramps-project.org/bugs/view.php?id=9993

sam-m888:
Thank you

zorkmark:
Thanks Sam.

It appears that installing pygraphviz on Windows7, under python 3.x has
some issues.  I assume that this is the case for 8 and 10 as well but
haven't confirmed.  I am investigating alternatives to pygraphviz and
just wanted to give a heads up not to try to package this for Windows
until I can find a better import dependency for cross-platform.
##SELECTION_END## 

Cheers,

Mark



-----Original Message-----

PR #68, opened 2017-04-19
make.py: support several xml files (eg Form) and update Form's template.pot

sebschub:
In order to update the translation of the Form gramplet, `make.py` has to support several xml files. This is done in the first commit. Before the Form's `template.pot` can be updated, an non-ASCII character has to be substituted from the US definitions (second commit). Finally, `template.pot` is updated.

Please check carefully whether these changes make sense. I am neither a python programmer nor a regular translator.

The change of `make.py` probably has to be cherry-picked to the master branch.

### DIFF
--- a/Form/form_us.xml
+++ b/Form/form_us.xml
@@ -3107,7 +3107,7 @@
                    <_attribute>Months unemployed during census year</_attribute>
                </column>
                <column>
-                   <_attribute>(Day of the Enumerator’s visit) sick or temporarily disabled</_attribute>
+                   <_attribute>(Day of the Enumerator&apos;s visit) sick or temporarily disabled</_attribute>
                </column>
                <column>
                    <_attribute>Blind</_attribute>
@@ -6232,7 +6232,7 @@ North Dakota Census
                 <_attribute>Can Read</_attribute>
             </column>
             <column>
-                <_attribute>Can't Read</_attribute>
+                <_attribute>Can&apos;t Read</_attribute>
             </column>
             <column>
                 <_attribute>Can Write</_attribute>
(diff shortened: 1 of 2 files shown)


Form/form_us.xml
```diff
@@ -2403,7 +2403,7 @@
                    <_attribute>Months unemployed during census year</_attribute>
                </column>
                <column>
-                   <_attribute>(Day of the Enumerator’s visit) sick or temporarily disabled</_attribute>
+                   <_attribute>(Day of the Enumerator's visit) sick or temporarily disabled</_attribute>
```
Nick-Hall: Do you really want to do this?  Changing an attribute will cause problems for existing users of the form.

sebschub: Sorry, @Nick-Hall , I missed your comment. I just had the feeling that a non-ASCII character is not optimal (producing error message) but more importantly, I find it hard to distinguish between " ` ", " ' " and " ´ ". Usually, I stick to " ' " and let the software do the right thing. In this case, I don't care because I will probably never use this entry. Please just say what you think and I update the PR accordingly.

Nick-Hall: Kurt is the maintainer of the US form definitions, so he should make the decision.

Caulley: The changes being requested will not effect any data that has already been entered. After sebschub confirms his issue has been resolved the new file can be released. The corrected file is at https://www.dropbox.com/personal/Public/Gramps%204.2?preview=form_us.xml

Caulley: The address to the corrected file should be https://www.dropbox.com/s/8u5hwjuu17agxga/form_us.xml?dl=0

sam-m888: @sebschub Can you confirm please?

sam-m888:
- [ ] TODO Add to master

PR #70, opened 2017-05-19
[ChangeGivenNames]Fix for bug 10044 exception on Windows

prculley:
Also, fixed trasient parent issues

### DIFF
--- a/ChangeGivenNames/ChangeGivenNames.py
+++ b/ChangeGivenNames/ChangeGivenNames.py
@@ -264,7 +264,7 @@ class ChangeGivenNames(tool.BatchTool, ManagedWindow):
                         fname = capitalize(name.first_name)
                         name.set_first_name(fname)
                 if change:
-                    self.db.commit_person(person, transaction=self.trans)
+                    self.db.commit_person(person, self.trans)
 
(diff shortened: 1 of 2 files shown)


sam-m888:
Testing the existing addon without patch showed the following error occurs when you "select" any of the resulting lines and then press "Accept changes and close":

33134: ERROR: grampsapp.py: line 143: Unhandled exception

>   File "/home/grampsdev/.gramps/gramps50/plugins/ChangeGivenNames/ChangeGivenNames.py", line 266, in on_ok_clicked
>     self.db.commit_person(person, transaction=self.trans)
> TypeError: commit_person() got an unexpected keyword argument 'transaction'

Did you get the same result without the patch for testing?

prculley:
@sam-m888 I did not get the error you mentioned.  I wonder if we have some difference in our setups; perhaps you were using dbapi?  Also I note that the error traceback above says line 266, however the line in my file (and on Github) is line 267...

sam-m888:
> perhaps you were using dbapi?

@prculley You are correct, sorry for not mentioning that, personally I've only been running gramps master with dbapi for the past year.

Thanks for the fix

PR #71, opened 2017-05-19
[ChangeGivenNames]Fix for bug 10044 exception on Windows

prculley:
Also, fixed transient parent issues

For Gramps42 branch


sam-m888:
Worked on Linux.

PR #72, opened 2017-05-25
fix bug 10053 PhotoTaggingGramplet exception

prculley:
when starting new selection with a person selected.

I have not built this for download/listing.  I also expect that our 'senior developer' will do the merges to master branch...

I also note that the PhotoTaggingGramplet source file is different than the one in the download list for Gramps master. Seems someone made changes and has not rebuilt the plugin since.

If accepted this should also be back-ported to Gramps42 branch...

### DIFF
--- a/PhotoTaggingGramplet/PhotoTaggingGramplet.py
+++ b/PhotoTaggingGramplet/PhotoTaggingGramplet.py
@@ -618,7 +618,7 @@ class PhotoTaggingGramplet(Gramplet):
 
         # populate the context menu
         persons = self.all_referenced_persons()
-        if selected.person is not None:
+        if selected is not None and selected.person is not None:
             persons.remove(selected.person)
         if persons:
             self.additional_items.append(Gtk.SeparatorMenuItem())


Nick-Hall:
> I have not built this for download/listing. I also expect that our 'senior developer' will do the merges to master branch...

Third-party addons don't follow the standard release cycle, so the committing policies for core Gramps don't apply.  In particular, new features are allowed in maintenance branches.  This is left to the discretion of the addon author.

For security reasons, only a few _Addon Publishers_ are allowed to push updates to the _addons_ repository.  However, any developer can build an update and submit a PR.

@prculley Let me know if you would like to be added to the _Addon Publishers_ team.

prculley:
@Nick-Hall So I guess I want to make it very clear, does the individual doing an addon or change to one need to do the work for BOTH gramps50 and master branches (and possibly gramps42 at his discretion)?  Or is some sort of 'merge' going to get done every so often to update master branch from gramps50?

I note that you made some fixes to the gramps50 branch, but not the master branch already so I infer that you expect the 'merge' to happen later, but I want to make sure.

Regarding the _Addon Publishers_; I'm not unwilling, but not very enthusiastic.  I just went through the whole process with another patch, only to discover I could not push to **addons**.  I suspect I would do it rarely enough that it becomes rather error prone, what with two repos to make sure I've pulled from, the various make commands (which only seem to work on my linux machine), etc.  This makes updating a bit difficult, maybe I need to make up a script or checklist if I'm going to do this.

I'm not too clear on whether we should be doing PRs for the Addons, and addon-source repos, either.

Nick-Hall:
@prculley For addons only:

1. Make changes to the _gramps50_ branch in the _addons-source_ repository and test.
2. Build the addon.
3. Commit the changes.
4. Update the listings.
5. Commit the new download and listings to the _addons_ repository.
6. Push the changes or make pull requests.
7. Cherry-pick your changes to the _master_ branch.

Check with the addon author for their committing policy.

prculley:
@Nick-Hall Back on May26 you asked if I would like to be a member of Addon Publishers team.  Since Sam seems to be busy these days, I will volunteer to move things along by dealing with some of these if you make me a member of that list.

PR #73, opened 2017-05-28
Fix bug 10058 Timeline View shows infinite lifespans for some persons

Aerilius:
When one date (birth or death) is not known, the lifespan would give a huge negative or a huge positive number, leading to excessive lifespan boxes.


sam-m888:
Thank you

PR #74, opened 2017-06-01
Fix bug 9970 ExtendedAttributes Plugin HandleError

prculley:
when attempting to edit the Person or Family using double-click


prculley:
Pushed commits directly to maintenance/gramps50

PR #75, opened 2017-06-17
Fixes for Gramplet signalling issues

prculley:
This PR is intended to match the main Gramps gramps50 branch repo https://github.com/gramps-project/gramps/pull/424, that is it does the same things.

Fixes #10089, #10090
I'm pretty sure that none of these changes will hurt compatibility with gramps50 alpha2 so it should be OK to merge this before alpha3 comes out.

I discovered that doing a Family merge that also merged at least one parent, would still get HandleErrors. Root cause, a lot of Gramplets were getting connected to Person history active-changed signal, and thus were doing unnecessary updates. In this case, the Person update occured before the Family Gramplets got shifted off of the family deleted by the merge, so the Gramplets tried to show the (now deleted) family again. And got HandleErrors.

I changed the base Gramplet class to only make default connections to the Person History active-change signal when the Gramplet had a navtype of 'Person' from the .gpr file. So unless a Gramplet has navtype=['Person'] in .gpr it will not get an automatic connection to Person active-changed.

A few Addon Gramplets that should have had navtype set to 'Person', did not.

It turns out a few Gramplets knew they were getting the 'extra' Person connection, and had a 'disconnect' which now failed with the prior change. So these extra disconnects are removed.

I also noticed that on db close/open, a lot of signals were getting connected when they already were connected. This because the connection was made in a db_changed method, but never disconnected.

To prevent multiple connections after db close/open, I decided to make all connections through Gramplet class 'connect', rather than directly to db, and to implement a disconnect on db_close and on db_changed, prior to making the connections again.

In the sweep though all our Gramplets I spotted a few other smaller issues like a wrong signal connect ('person-edit', rather than 'person-update'). And a few signals were missing from some of the Gramplets.

### DIFF
--- a/BirthdaysGramplet/BirthdaysGramplet.py
+++ b/BirthdaysGramplet/BirthdaysGramplet.py
@@ -41,14 +41,10 @@ class BirthdaysGramplet(Gramplet):
         self.set_text(_("No Family Tree loaded."))
         self.max_age = config.get('behavior.max-age-prob-alive')
 
-    def post_init(self):
-        self.disconnect("active-changed")
-
     def db_changed(self):
-        self.dbstate.db.connect('person-add', self.update)
-        self.dbstate.db.connect('person-delete', self.update)
-        self.dbstate.db.connect('person-update', self.update)
-        self.update
+        self.connect(self.dbstate.db, 'person-add', self.update)
+        self.connect(self.dbstate.db, 'person-delete', self.update)
+        self.connect(self.dbstate.db, 'person-update', self.update)
 
     def main(self):
         self.set_text(_("Processing..."))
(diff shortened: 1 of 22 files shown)


sam-m888:
Rebuild addons for 
- [ ] gramps50

Cherry-pick and rebuild for:
- [ ] master

.............................
- [ ] BirthdaysGramplet
- [ ] DataEntryGramplet
- [ ] DeepConnectionsGramplet
- [ ] DescendantCount
- [ ] EditExifMetadata
- [ ] ExtendedAttributes
- [ ] FaceDetection
- [ ] ImportGramplet
- [ ] LastChange
- [ ] MediaBrowser
- [ ] NoteGramplet
- [ ] PhotoTaggingGramplet
- [ ] PythonGramplet
- [ ] Query
- [ ] RelatedRelativesGramplet
- [ ] SourceReferences
- [ ] WordleGramplet
- [ ] lxml/etreeGramplet.py
- [ ] lxml/lxmlGramplet.py

sam-m888:
Thank you

PR #81, opened 2017-08-15
[PhotoTagging] update Russian translation

vantu5z:


sam-m888:
Thank you for merging 👍 

Note that usually you leave it open so that when it is merged it can also be rebuilt and be made available (which I've now done for this and the gramps42 branch, do you intend to add it to gramps master also?).

PR #82, opened 2017-08-15
[PhotoTagging] try to fix face detection

vantu5z:
https://gramps-project.org/bugs/view.php?id=10164


sam-m888:
Thank you

Rebuilt  https://github.com/gramps-project/addons-source/commit/708651088f76d9bef8f33eb8501e0ff71ea17191

PR #83, opened 2017-08-16
Fix DenominoVisio to use full automatic place name generation

prculley:
Bug From Gramps-users email 

http://gramps.1791082.n4.nabble.com/Place-names-in-DenominoVisio-web-page-td4680631.html


sam-m888:
Thank you

PR #84, opened 2017-08-24
[LastChanges] Add Russian translation

vantu5z:


vantu5z:
Don't know that I have write access :)

PR #86, opened 2017-08-30
[GraphView] another way to display tags

vantu5z:
Tags display by color lines in the bottom of person box.
It looks like this:

### DIFF
--- a/GraphView/graphview.py
+++ b/GraphView/graphview.py
@@ -147,6 +147,9 @@ class GraphView(NavigationView):
         self.graph_widget = None
         self.dbstate.connect('database-changed', self.change_db)
 
+        # dict {handle, tooltip_str} of tooltips in markup format
+        self.tags_tooltips = {}
+
         self.additional_uis.append(self.additional_ui())
         self.define_print_actions()
 
@@ -898,6 +901,8 @@ class GraphvizSvgParser(object):
         else:
             line_width = 1  # Thin box
 
+        tooltip = self.view.tags_tooltips.get(self.handle)
+
         # Highlight the home person
         # stroke_color is not '#...' when tags are drawing, so we check this
         # maybe this is not good solution to check for tags but it works
@@ -905,13 +910,20 @@ class GraphvizSvgParser(object):
             home_person = self.widget.dbstate.db.get_default_person()
             if home_person and home_person.handle == self.handle:
                 fill_color = self.home_person_color
+                tooltip = None
 
         item = GooCanvas.CanvasPolyline(parent = self.current_parent(),
                                         points = points,
                                         close_path = True,
                                         fill_color = fill_color,
                                         line_width = line_width,
-                                        stroke_color = stroke_color)
+                                        stroke_color = stroke_color,
+                                        tooltip = tooltip)
+        # turn on tooltip show if have it
+        if tooltip:
+            item_canvas = item.get_canvas()
+            item_canvas.set_has_tooltip(True)
+
         self.item_hier.append(item)
 
     def stop_polygon(self, tag):
@@ -938,6 +950,8 @@ class GraphvizSvgParser(object):
             stroke_color = attrs.get('stroke')
             fill_color = attrs.get('fill')
 
+        tooltip = self.view.tags_tooltips.get(self.handle)
+
         item = GooCanvas.CanvasEllipse(parent = self.current_parent(),
                                        center_x = center_x,
                                        center_y = center_y,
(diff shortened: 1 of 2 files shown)


vantu5z:
I can't understand why tags not shown on home person (when it highlight).

SNoiraud:
The second commit doesn't work for me.
If I change 'transparent' by 'none', it's OK.
I don't know where none is loaded in gramps!

SNoiraud:
Do you use styles? perhaps this is the difference. Neither transparent or none are present in gramps sources.
I think these values come from graphviz.

vantu5z:
Yes, I think it come from graphviz.
I just insert print():
```
if self.highlight_home_person:
            print(attrs)
```
and get strings like this:
```
...
{'fill': '#b9cfe7', 'stroke': '#204a87', 'points': '4853,-974 4608,-974 4608,-811 4853,-811 4853,-974'}
{'fill': '#efb60c', 'fill-opacity': '0.156863', 'stroke': 'transparent', 'points': '4620.5,-818.5 4620.5,-828.5 4841.5,-828.5 4841.5,-818.5 4620.5,-818.5'}
{'fill': '#b9cfe7', 'stroke': '#204a87', 'points': '4474,-455.5 4379,-455.5 4379,-419.5 4474,-419.5 4474,-455.5'}
...
```

vantu5z:
For me works this:
`if self.highlight_home_person and (not attrs.get('fill-opacity')):`

SNoiraud:
OK. You have 'stroke' in the attrs. This is the reason you have 'transparent'.
I don't know the reason you have that.
For me, colors always begin with '#'. Yours too. Perhaps we could change the if statment by : 
if self.highlight_home_person and stroke_color[:1] == '#':

SNoiraud:
I don't have fill-opacity. I only have the following : 
attrs : {'points': '9497,-981.5 9497,-991.5 9714,-991.5 9714,-981.5 9497,-981.5', 'fill': '#edd400', 'stroke': 'none'}

vantu5z:
OK, that we should to check if it is a color like you suggested.

SNoiraud:
As we are trying to add tags to people, it could be usefull to do the same thing for families.
I already have something for that.

SNoiraud:
Here is my first try. Does it work for you.
I added the possibility to refresh the screen when we change the show tags button.

[show_tags.txt](https://github.com/gramps-project/addons-source/files/1268564/show_tags.txt)

vantu5z:
Yes, it works. But I think it should be like this.

vantu5z:
And look like this:

SNoiraud:
Perhaps we could add this to gramps 5.0. It's a really nice feature.

vantu5z:
I would like to see it in gramps 5.0 too.

vantu5z:
Should we show this PR for some one else or can merge it?
And who decide if it go to gramps 5.0?

SNoiraud:
I think you can merge it.
As this is a addon module, I think this is to the developper to decide.
Perhaps Sam ( @sam-m888 ) could confirm this.

sam-m888:
> As this is a addon module, I think this is to the developper to decide.

@SNoiraud Yes Addons are handled by the original developer or whoever contributes to it.  So if a gramps50/[42](https://github.com/gramps-project/addons-source/pull/86#issuecomment-326538632) pr are made I will rebuild the addon for you @vantu5z 

Very good enhancement 👍 

Suggestion is it possible to:

* have a legend of what each color tag is?
* or display a tooltip when you hover over the color ?
* or possibly partially display the tag name in the tag box?

vantu5z:
Make PR #87 for gramps50 branch

sam-m888:
Rebuilt for Gramps master as https://github.com/gramps-project/addons/commit/2cbceca0f8fa9c0be1ebf20cb8f0c880b31a129d

sam-m888:
Thanks you for merging, really would have preferred that you let someone review that it works as advertised in case of a major issue so that it does not get pushed out to normal users.

So thanks for adding the legend for tags, unfortunately after looking a bit closer (image below) you can see that the colors for the tags don't match those in the legend table.

img_20170908_130447727

vantu5z:
I don't know why colors of tags at graph differ from legend (in legend correct). But, if you open tag in tag manager and resave it, it will be shown correctly.

sam-m888:
> Maybe gramps example tree have tags colors in other format that used befor.

Not as far as my testing shows so far!

PR #88, opened 2017-09-05
[Graphview] graph zoom and navigation[Gramps51]

vantu5z:
**PR description**
New:
- zoom by mouse wheel;
- move graph by middle mouse;
- add button to best fit;
- add button to center on active person;
- remove zoom panel from toolbar.

Changes:
- coef for scale when graph lager than paper applied to operations with graph, so zoom=1 is the similar for large tree and for small;

Fixes:
- positioning on active person;
- flicking when move graph (also fix speed of moving). (Was in PR #52 )

### DIFF
--- a/GraphView/graphview.py
+++ b/GraphView/graphview.py
@@ -747,12 +747,6 @@ class GraphWidget(object):
         self.scale = value
         self.canvas.set_scale(value / self.transform_scale)
 
-    def zoom_changed(self, adj):
-        """
-        Zoom the canvas widget
-        """
-        self.canvas.set_scale(adj.get_value() / self.transform_scale)
-
     def select_node(self, item, target, event):
         """
         Perform actions when a node is clicked.
--- a/GraphView/graphview.py
+++ b/GraphView/graphview.py
@@ -529,6 +529,7 @@ class GraphWidget(object):
         self.vadjustment = scrolled_win.get_vadjustment()
 
         self.canvas = GooCanvas.Canvas()
+        self.canvas.connect("scroll-event", self.scroll_mouse)
         self.canvas.props.units = Gtk.Unit.POINTS
         self.canvas.props.resolution_x = 72
         self.canvas.props.resolution_y = 72
@@ -553,6 +554,24 @@ class GraphWidget(object):
         hbox.pack_start(self.scale, True, True, 0)
         self.vbox.pack_start(scrolled_win, True, True, 0)
         self.change_max_zoom()
+
+    def scroll_mouse(self, canvas, event):
+        """
+        Try to zoom with mouse wheel.
+        """
+        scale_coef = self.canvas.get_scale()
+        if scale_coef < 2:
+            step = 0.1
+        else:
+            step = 1
+        if event.direction == Gdk.ScrollDirection.UP:
+            scale_coef += step
+            self.canvas.set_scale(scale_coef)
+        elif event.direction == Gdk.ScrollDirection.DOWN:
+            scale_coef -= step
+            if scale_coef < 0.2:
+                scale_coef = 0.1
+            self.canvas.set_scale(scale_coef)
 
     def populate(self, active_person):
         """
--- a/GraphView/graphview.py
+++ b/GraphView/graphview.py
@@ -636,8 +636,11 @@ class GraphWidget(object):
             self.scale1.destroy() # destroy the Scale if it exists.
         except:                   # we can't change the max value
             pass                  # then recreate a new scale
+        nb_persons = int(self.dbstate.db.get_number_of_people()+5)
+        zoom = log(nb_persons,10)*log(nb_persons,5)
+        max_zoom = 2.0 if zoom < 2.0 else zoom
         # (value, lower, upper, step_increment, page_increment, page_size)
-        adj = Gtk.Adjustment(1.00, 0.01, 2.00, 0.01, 0.05, 0)
+        adj = Gtk.Adjustment(1.00, 0.01, float(max_zoom), 0.01, 0.05, 0)
         adj.set_value(1.0)
         self.scale1 = Gtk.Scale(adjustment=adj,
                                 orientation=Gtk.Orientation.HORIZONTAL)
--- a/GraphView/graphview.py
+++ b/GraphView/graphview.py
@@ -33,7 +33,6 @@
 #
 #-------------------------------------------------------------------------
 import os
-from math import log
 from xml.parsers.expat import ExpatError, ParserCreate
 from gramps.gen.const import GRAMPS_LOCALE as glocale
 try:
@@ -176,7 +175,7 @@ class GraphView(NavigationView):
         Set up callback for changes to the database
         """
         self._change_db(db)
-        self.graph_widget.change_max_zoom()
+        self.scale = 1
         if self.active:
             self.graph_widget.clear()
             if self.get_active() != "":
@@ -557,17 +556,11 @@ class GraphWidget(object):
         hbox.pack_start(self.goto_active_btn, False, False, 1)
         self.goto_active_btn.connect("clicked", self.goto_active)
 
-        zoom_label = Gtk.Label(label=_("Zoom:"))
-        hbox.pack_start (zoom_label, False, False, 1)
-
-        self.scale = Gtk.Box(False, 4, orientation=Gtk.Orientation.HORIZONTAL)
-        hbox.pack_start(self.scale, True, True, 0)
         self.vbox.pack_start(scrolled_win, True, True, 0)
         # if we have graph lager than graphviz paper size
         # this coef is needed
         self.transform_scale = 1
-
-        self.change_max_zoom()
+        self.scale = 1
 
     def goto_active(self, button):
         """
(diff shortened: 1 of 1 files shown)


SNoiraud:
I don't like the zoom change. I have more than 8000 people in my database.
I need to have the possibility to zoom at more than 20.
The scale to fit is unusable in this case.
This is why the zoom has been set depending on the database size.

SNoiraud:
Perhaps we could only change the maximum value (5) : 
max_zoom = 5.0 if zoom < 5.0 else zoom

vantu5z:
hm.. I think that default (1) zoom let to see normal scale of person boxes and text on it. And we need to zoom out to see more people.
If scale=1 depends on the database size it's not good I think.
Do you have test tree with large count of persons so I cat make test on it?

vantu5z:
And the scale to fit can be used on large tree if counts of generations is set.

vantu5z:
Yes, you are right!
Found description of graphviz size attribute http://www.graphviz.org/doc/info/attrs.html#d:size
We have set size to 100 inches and when graph is larger than it will be scaled.

vantu5z:
I think we can look how graph was scaled and then set max zoom depending on that scaling. This allow to set max zoom depending on real graph on screen and not all db.

vantu5z:
And another idea to set coef to zoom scale, so graph is shown  similar for large db and not for zoom=1.

SNoiraud:
The original commit I made was : ef6ec4297c83cc78a55aa87748df946ce4a221f3
I think now the max zoom should depend on the database size and the number of generations shown.
Or a better idea : remove this scaling and replace this by the wheel mouse.

vantu5z:
I think it should stay, maybe someone need it.
About wheel - I want to present it in next PR with drug and zoom. But can't catch wheels events now :( Can you help me in that? :)

SNoiraud:
I can try it.

SNoiraud:
first try. I think we can get it better.
[scroll.patch.txt](https://github.com/gramps-project/addons-source/files/1278801/scroll.patch.txt)

vantu5z:
Thank you to show how it works!
Try this it should work with large and small tree.

vantu5z:
I don't have large tree, but I test it on gramps example with set graphviz paper size to 5 instead of 100.
And it works for me.

SNoiraud:
Some comments. I added images from my database.
zoom_to_best_feet
This part of my tree is unreadable which contains more than 2000 people.
zoom-2 0
With a zoom of 2 max, I can't follow the lines.
What is the purpose of GraphView_drug_by_middle_mouse ? I can't use it.
The go to active is not good for me. Something is missing.
I think we could select goto one people in the bookmark list or in the go list.

vantu5z:
Ok.
1. We can set min zoom to some value in which we can see something  (it was static). But I think 0.01 looks better then 0.04 for example. 
2. For casual use we don't need zoom more then 1. But sometimes it needed to show node for old people quickly or on TV screen.
3. Move by middle mouse is useful feature. It allow to move graph and don't worry in what place you click. It go to motion even we click on node.
4. Seems that I set not good icon name that you don't have.
5. Good idea about goto person (but I think better to separate it from this PR). Will be good to add anomation to this operation.

SNoiraud:
> 1. We can set min zoom to some value in which we can see something (it was static).

The problem is not the zoom. Best fit with many people (> 2000 in the given case) is the problem.

> 2. For casual use we don't need zoom more then 1

It depends of number of generations you want to see.
I know this is true now with the possibility to have numbers of ascendant and descendant generation.
If you want to have for example 8 ascendant generations and 9 of descendant, 2 is not sufficient. This is why I used the zoom like it was before

> 4. Seems that I set not good icon name that you don't have.

It's not the 'go-jump' icon, but the 'zoom-fit-best' icon you see on the first snapshot.

> 5. Good idea about goto person.

I'm working on this. I'll make a new PR.

SNoiraud:
Personaly I will remove all what is needed by the scales and the problem  will disapear.
There will be no more limits and all people will be happy.
remove change_max_zoom function, the zoom_label, self.scale1. We only need to save the current scale.

vantu5z:
I have a feeling that we talking about different things.

Try to explain what I suggest.
How said I don't have large tree and all operations will be done with gramps example tree.

Normal behaviour for small tree with or without proposed changes
(only difference is that we have different range of zoom)
1. For zoom=1:


2. For zoom=2:


Behaviour for reduced graphviz page size (100 -> 5), that simulate large tree
1. For zoom=1 (without proposed changes):


2. With proposed changes result will be the same as for small tree.

How can we see, we have reduced tree with the scale that graphviz apply to page.
I don't like this behaviour: than tree is lager -> than picture is smaller for zoom=1.

With proposed changes we get the same size of nodes and text on it (with same zoom) for large and small tree. If you want to see reduced tree when you open View - we can add default zoom factor to config dialog, but I think zoom=1 should be with expected font and thumbs size.
This needed even we remove the zoom panel and change_max_zoom function I think.

If you don't see the same size of nodes for gramps example and your large tree with zoom=1, that something go wrong...

About best fit button: yes, it's not for fitting huge tree, it's for quickly fit to window not large tree (or with generations set tree) or to see observed topology of tree and zoom-in to needed place. For me it's very useful.

> With a zoom of 2 max, I can't follow the lines.
1. Why do you have so scale widget? Look at my screens - they are different.
2. And what do you mean? Do you need more or less? Or this is another issue that have not solution for now?

> What is the purpose of GraphView_drug_by_middle_mouse ? I can't use it.

Strange. What is your value for button in button_press function when you press middle mouse?

SNoiraud:
I agree. This works correctly with small trees.
The problem is with big trees and  if you change the number of ascendant and descendant generation.
If you look at the second image I attached above, with a zoom of 2, I can't follow the lines.
I don't mind if I see a node partially, I need to follow the lines.
If we suppress this limit to 2 (the scale), we can have all I want.
I attach a zoom of 21 with my database. You can see it is insufficient.
zoom-21

vantu5z:
I don't understand you.
Do you checkout the branch and try it on your tree or you apply some commits?
Zoom=1 should show you nodes with text size 14 (that is not small). And if you zoom=2 you have to see partially nodes.
If it works like I expected and you want follow the lines you should zoom-out (how you like) to do this.

Can you show zoom=1 on your tree with checkout branch (included all the changes to addon)?
It should be lager than in your last screenshot with zoom=21.

Do you use Linux or Win or Mac, and what is Graphviz and Gtk versions?
Maybe problem in this case (that we see different result)
I have this versions and use linux:
graphviz 2.40.1-7
gtk3 3.22.19-2

SNoiraud:
I suppressed the sacle. Now I can do what I want.
Here is another example : The upper line is for this mariage and the two other below are for another mariage. Impossible to detect with a zoom of 2.
For this image, the zoom is 15.8
zoom-15
The patch I used is the following to suppress the scale : 
[suppress_scale.patch.txt](https://github.com/gramps-project/addons-source/files/1283951/suppress_scale.patch.txt)

SNoiraud:
I checked the branch and the last commit is : "stop scroll by wheel - just zoom"
I use linux with graphviz 2.38 and gtk3 3.20.9

vantu5z:
Ow, now I understand you :)

vantu5z:
Changes was applied with little changes, you can try it.
Looks nice :)

What about middle mouse and go to active button?

SNoiraud:
OK for the middle mouse.
OK for the active button (useful when the active person is not visible).
All seems good to me.

vantu5z:
Thanks for your help! 
So, can I merge it?

sam-m888:
Looks good apart from my zoom and navigation issues from as quick attempt to use this only showed me that I can not use it as I only have a [trackpoint keyboard](https://en.wikipedia.org/wiki/Pointing_stick) eg :

* I do not have a mouse with a scroll wheel (so could not test)
* or a middle mouse (so could not test)
*I don't have a mouse (due to RSI issues)

Which means as a keyboard user when I select the "Zoom to best fit" button I am stuck at that zoomed best fit with no easy way to return to normal scale (none of the three possible buttons do anything useful) 

2017-09-08 15_57_30-ubuntu-16 04 1-grampsdev-amd64 running - oracle vm virtualbox


To return to normal scale I selected a single individual and because its microscopic(10000 people) 

2017-09-08 15_53_31-ubuntu-16 04 1-grampsdev-amd64 running - oracle vm virtualbox


I then need to hit  "Zoom to best fit"  again

2017-09-08 15_53_45-ubuntu-16 04 1-grampsdev-amd64 running - oracle vm virtualbox

 and then select "Go to the default person"(Home button) , please bring back some type of scaler(Zoom panel!) or add keyboard shortcuts for those of us that don't have a mouse!


PS: Is anyone annoyed that Graphview uses the same icon as the pedigree view?

vantu5z:
@sam-m888 , thank you for testing. I think we should add buttons (zoom-in, zoom-out, zoom-1).

Also I try to make animation to positioning. And don't know include to this PR or not.
You can test it: 969ac275af47656a1af26b0894abd3117518c9df
Say me if I do this in not right way.

SNoiraud:
Animation seems good to me.

sam-m888:
> I think we should add buttons (zoom-in, zoom-out, zoom-1).

@vantu5z Thanks for adding the zoom buttons, that fixes the usability issue 👍 

2017-09-09 08_26_41-ubuntu-16 04 1-grampsdev-amd64 running - oracle vm virtualbox

sam-m888:
Thank you

Addon rebuilt as https://github.com/gramps-project/addons/commit/e241117737b94df4e7b122ff300e08f86fd6835f

If you want you can create a backport/PR for Gramps50 also?

SNoiraud:
FYI : I got the following today : 
2017-10-05 11:52:18.163: ERROR: grampsapp.py: line 143: Unhandled exception
  File "/home/serge/.gramps/gramps50/plugins/GraphView/graphview.py", line 789, in fit_to_page
    scale_h = (height / height_canvas)
ZeroDivisionError: float division by zero

vantu5z:
I don't know how you get this, but fix this in commit 73e70d8e76898d99f9161e8e9ea78f3178d7ca49.

PR #89, opened 2017-09-12
Graph view animation [gramps51]

vantu5z:
- class for animation added;
- apply animation for moving to active person (on button click);
- add ability to develop "move to person" feature;
- add combobox with bookmarked persons to navigate to them (if in current tree);
- add shake animation after moving to person.

We need someone who know about threads to look how it realized.
Testing is needed.

### DIFF
--- a/GraphView/graphview.py
+++ b/GraphView/graphview.py
@@ -585,6 +585,9 @@ class GraphWidget(object):
         self.transform_scale = 1
         self.scale = 1
 
+        self.animation = CanvasAnimation(self.canvas, scrolled_win,
+                                         self.transform_scale)
+
     def goto_active(self, button):
         """
         Go to active person
@@ -605,9 +608,13 @@ class GraphWidget(object):
         bounds = self.canvas.get_root_item().get_bounds()
         height_canvas = bounds.y2 - bounds.y1
 
-        # Centre the active person
-        self.canvas.scroll_to(self.active_person_x - h_offset,
-                              height_canvas + self.active_person_y - v_offset)
+        # check if animation is needed
+        if button:
+            animation = True
+        else:
+            animation = False
+
+        self.animation.move_to_person(self.active_person_handle, animation)
 
     def scroll_mouse(self, canvas, event):
         """
@@ -647,6 +654,8 @@ class GraphWidget(object):
         # save transform scale
         self.transform_scale = parser.transform_scale
         self.set_zoom(self.scale)
+
+        self.animation.update_items(self.transform_scale)
 
         # Save position of the active person
         if parser.active_person_item:
--- a/GraphView/graphview.py
+++ b/GraphView/graphview.py
@@ -186,7 +186,6 @@ class GraphView(NavigationView):
             self.dirty = True
         # update combobox with bookmarks on db changes
         self.graph_widget.load_bookmarks()
-        self.dbstate.connect('active-changed', self.graph_widget.load_bookmarks)
 
     def get_stock(self):
         """
@@ -528,8 +527,6 @@ class GraphWidget(object):
         self.dbstate = dbstate
         self.uistate = uistate
         self.active_person_handle = None
-        self.active_person_x = 0
-        self.active_person_y = 0
 
         scrolled_win = Gtk.ScrolledWindow()
         scrolled_win.set_shadow_type(Gtk.ShadowType.IN)
@@ -681,11 +678,7 @@ class GraphWidget(object):
 
         self.animation.update_items()
 
-        # Save position of the active person
-        if parser.active_person_item:
-            self.active_person_x = parser.get_active_person_x()
-            self.active_person_y = parser.get_active_person_y()
-
+        # Scroll to activ person without animation
         self.goto_active(None)
 
         # Update the status bar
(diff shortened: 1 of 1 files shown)


vantu5z:
@SNoiraud , look this PR. Now you can use "move_to_person" function to move (with animation) to specified person by handle. We just need some friendly person chooser to realize "go to person" feature.

SNoiraud:
Seems good to me.
I have tested with bookmarks and If the bookmark is in the tree, we can animate to this person.
The patch is the following : 
[goto_other.patch.txt](https://github.com/gramps-project/addons-source/files/1297071/goto_other.patch.txt)

vantu5z:
@SNoiraud , patch applied with little changes.
And shaking animation after moving to person was added. It helps to see this person, because we don't highlight it anyhow and it can be not in center of screen.

SNoiraud:
When we add a new bookmark, we need to change the combobox. So, we need to add the following in the change_db method : 
        self.dbstate.connect('active-changed', self.graph_widget.load_bookmarks)

SNoiraud:
Last patches doesn't work : 
TypeError: move_to() missing 1 required positional argument: 'animated'
Warning: DbBsddb: Exception occurred in callback function.
  File "/home/gramps/gramps50/gramps/gen/utils/callback.py", line 405, in emit
    fn(*args)
  File "/home/serge/.gramps/gramps50/plugins/GraphView/graphview.py", line 258, in goto_handle
    self.graph_widget.populate(self.get_active())
  File "/home/serge/.gramps/gramps50/plugins/GraphView/graphview.py", line 698, in populate
    self.goto_active(None)
  File "/home/serge/.gramps/gramps50/plugins/GraphView/graphview.py", line 641, in goto_active
    self.animation.move_to_person(self.active_person_handle, animation)
  File "/home/serge/.gramps/gramps50/plugins/GraphView/graphview.py", line 2145, in move_to_person
    self.move_to((x, y), animated)
TypeError: move_to() missing 1 required positional argument: 'animated'

vantu5z:
@SNoiraud , fix missing argument. And make separated counters for shaking - it needed to prevent incorrect shaking of items on quickly target switching.

SNoiraud:
In some case, move_to doesn't go to the good person.
it happens with "go to the active person" or "go to to the bookmark person".
Is it related to shake? I don't know.

vantu5z:
I don't have test alot. But known issue: if zoon when it scroll to person - you go to incorrect coordinates, because track calculated before zooming. I think about dynamic calculation.

SNoiraud:
I don't scroll.

SNoiraud:
I didn't reproduce the move_to problem until now.  I suppose the last patch solves the problem.

vantu5z:
Now trace to person calculated dynamically. We can move, scroll or .. when animation in motion.
Maybe we need stop animation when we try to move tree by mouse?

Testing is needed.

SNoiraud:
It's OK for me.
try to zoom when moving on long distances : OK
try to move during scrolling the window : OK
I try to add a video to show this (if it's not too big 6.2MB)
[PR89.mp4.txt](https://github.com/gramps-project/addons-source/files/1306740/PR89.mp4.txt)
You can remove the .txt to view it.

vantu5z:
Good! :)
But i found if we add new bookmark - combobox not updated as expected. And you?

And some ideas:
Maybe don't add bookmarks that not in current tree?
If we select one from combobox - we can't select it again while it selected.
It would be nice to have filterable combobox which filter list when enter name. And not for bookmarks but for persons in the current tree.

SNoiraud:
I have the same problem. When I tested it it worked.
We need to add self.graph_widget.load_bookmarks()
in to goto_handle method.

SNoiraud:
> Maybe don't add bookmarks that not in current tree?

The best way to do this should be to have a method : Is_this_person_present_in_the_tree
If False, we don't add it to the combobox.
Is the get_item_by_title method the solution ?

> If we select one from combobox - we can't select it again while it selected.

I look at this.

SNoiraud:
I have one solution : 
[bkmarks.patch.txt](https://github.com/gramps-project/addons-source/files/1308310/bkmarks.patch.txt)

vantu5z:
I test it after weekend :)

sam-m888:
Very nice animation 👍  Will test more next week.

SNoiraud:
After last patch (bkmarks.patch.txt), we need to remove line 596 in __init__ : 
self.load_bookmarks()

  File "/home/serge/.gramps/gramps50/plugins/GraphView/graphview.py", line 596, in __init__
    self.load_bookmarks()
  File "/home/serge/.gramps/gramps50/plugins/GraphView/graphview.py", line 617, in load_bookmarks
    present = self.animation.get_item_by_title(bkmark)

vantu5z:
Combobox updating works for me :)

SNoiraud:
OK for me. very nice.

SNoiraud:
I use it since september. I had no problem.
The shaking is very fast and short. I have a patch to modify the speed and count.
If you apply it, I think we could merge it.
[count.patch.txt](https://github.com/gramps-project/addons-source/files/1629257/count.patch.txt)

vantu5z:
Add your patch.
For me animation speed is very fast, so I set `GLib.usleep(20 * self.speed)` between steps.
Fix shaking animation when switch to other person and shaking in progress.

sam-m888:
Thank you 👍 

Addon rebuilt in https://github.com/gramps-project/addons/commit/0fdfe642e5397157b3ca109c6d97d5583efe9c47

PR #92, opened 2017-10-07
GetGOV fix for 0010220

GaryGriffin:
Fix in 5.0 tree for non-English query to gazetteer. If no place returned by gazetteer, do not create a blank place


sam-m888:
Thank you

PR #93, opened 2017-11-04
Fix NetworkChart and TMGimporter to work with Travis (no display)

prculley:
The .gpr files were importing Gtk, which doesn't work for Travis
because there is no display.


sam-m888:
Thank you

PR #94, opened 2017-11-10
Update Descendant Count Gramplet for performance etc.

prculley:
[Fixes #9981](https://gramps-project.org/bugs/view.php?id=9981)
This Gramplet puts up a list of all people and the number of their descendants.

The original version had several problems, the first and most important was that it took a long time to execute with large dbs and during that time Gramps was unresponsive.

Under v4.2 the Gramplet also updated with every active-change of the Person view, and due to some non-optimal GUI code was updated multiple times for every change.

The algoritm used exhibited a people^2 time function of some sort.  I modified it to operate in approximately linear time with the number of people.  I also modified it to operate only during Gtk 'idle' periods, so the GUI does not go unresponsive, even if it takes some time to run and finish updating the result pane.

One downside is that it takes more memory during the update; I spotted a 3.1MB temporary jump with the example.gramps db.  Shouldn't be a problem with modern systems.

In testing with the example.gramps db I found the following times:
Gramps50 modified: 3.6 sec
Gramps50 original: 45.5 sec
Gramps42 original: 227.5 sec

Under the same conditions (debugger), the #9981 db (6828 dogs) took 14.8 seconds, rather than the 12+ hours reported.

I also modified it to update with changes of family; the original did not do this.

### DIFF
--- a/DescendantCount/DescendantCount.gpr.py
+++ b/DescendantCount/DescendantCount.gpr.py
@@ -1,16 +1,21 @@
-register(QUICKREPORT,
-         id = 'Descendant Count Quickview',
-         name = _("Descendant Count"),
-         category = CATEGORY_QR_MISC,
-         runfunc = "run",
-         status = STABLE,
-         description= _("Display descendant counts for each person."),
-         fname="DescendantCount.py",
-         authors=["Douglas S. Blank"],
-         authors_email=["doug.blank@gmail.com"],
-         version = '1.0.24',
-         gramps_target_version = "5.0",
-         )
+#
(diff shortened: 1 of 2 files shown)


sam-m888:
Worked for my 13000 person family tree.

prculley:
@sam-m888 When I redid it, I compared the counts to the original version as running under Gramps4.2.6.  They are the same.  I would update the wiki to match the results.

sam-m888:
@prculley Ok, I've made the wiki description simpler and removed that line.

sam-m888:
Thank you

PR #97, opened 2017-11-19
Update LastChange Gramplet and Report

prculley:
Fixes #10267
Adds some performance enhancements
Adds Notes and Citations to Report


sam-m888:
Thank you

PS: Only pushing without check as you've already [committed](https://github.com/gramps-project/addons/commit/d6887737fe0ecd9bddaf007e55632110b8ee7c52) the rebuild of the addon,  usually it works the other way round, so it can be tested first. Also the released version number shows as '0.1.2' instead of '0.1.1' that is in this PR please try to keep them synchronized.

prculley:
@sam-m888 Sorry about that, I got interrupted before I could upload my post-build source, that's why the version numbers got out of sync.  I really intended to merge the PR myself, but after I made sure that Travis etc was happy.  You beat me to it.

PR #98, opened 2017-12-31
Norwegian Bokmål translation for BirthdaysGramplet

GitHaarek:


Nick-Hall:
Thanks.  I don't think we have a translator for Norwegian (Bokmål) at the moment.

PR #99, opened 2018-01-06
DescendantLines updates [gramps50]

prculley:
* Only display events with PRIMARY or FAMILY role [# 10347](https://gramps-project.org/bugs/view.php?id=10347)
* prevent output corruption when run twice.  Output file was opened for write several times, when the report was run twice this caused a file corruption the second time.  This also allowed the output file to be deleted before Gramps was closed (it was open).
* Fix incorrect filename extension.  When changing output type, the filename extension was not being updated so file could not be opened by associated program.  Note; this is NOT updated in GUI, only in actual file.  Too lazy to completely rework addon menu system to fix this.
* Prevent and warn user not to use SVG output type for supplemental report (doesn't support text operations). #10201 #9785
* Pylint


Nick-Hall:
Bug fixes should be made in the _gramps42_ branch.

I have already fixed bug [10347](https://gramps-project.org/bugs/view.php?id=10347) in commit d289b5cbb653ebd4557f065ac47919941540d031.

prculley:
Yes, I copied that fix to this branch while working on some other issues.  I can retrofit the other fixes to gramps42 (probably not the pylint).  But I thought your policy was to let the developer decide what branches to update for addons?  I.E. you are not merging commits from gramps42 to gramps50 and gramps51 like in the main Gramps repo.

I keep hoping that we will eventually move gramps50 to a real release...

Nick-Hall:
I tend to leave decisions relating to third-party addons to their author.  When fixing a bug in someone else's addon, I would fix it in the current release so that it is immediately available to users.

You seem to have merged my fix in with another.  Please keep to one commit per fix if possible.

prculley:
Fixes separated, and your fix removed, as it was uploaded since I started.  I have a build/listing commit ready to upload when this is accepted.

I still think Travis test failure is due to some change in Gramps around sqlite or dbapi...

Nick-Hall:
This needs a rebase.

sam-m888:
Thank you

Addon rebuilt for Gramps50 in https://github.com/gramps-project/addons/commit/f9002928dc31b4b9c67ce5b697dc86da32907145

PR #100, opened 2018-01-06
DescendantLines fixes for gramps42

prculley:
* prevent output corruption when run twice.  Output file was opened for write several times, when the report was run twice this caused a file corruption the second time.  This also allowed the output file to be deleted before Gramps was closed (it was open).
* Fix incorrect filename extension.  When changing output type, the filename extension was not being updated so file could not be opened by associated program.  Note; this is NOT updated in GUI, only in actual file.  Too lazy to completely rework addon menu system to fix this.
* Prevent and warn user not to use SVG output type for supplemental report (doesn't support text operations). [#10201](https://gramps-project.org/bugs/view.php?id=10201) [#9785](https://gramps-project.org/bugs/view.php?id=9785)


sam-m888:
Thank you

Rebuilt addon for Gramps42 in https://github.com/gramps-project/addons/commit/058b3c78d116d641af1a9a61e17fc02065fe01c2

sam-m888:
for gramps50 in PR https://github.com/gramps-project/addons-source/pull/99

PR #101, opened 2018-01-13
GraphView; some printing issues[gramps42]

prculley:
* Fix GraphView for file write errors; use a message, not exception.    Fixes [#10363](https://gramps-project.org/bugs/view.php?id=10363)
* Fix GraphView; Encoding exception when printing on non utf-8 system when unicode chars are present in file.  Found on Windows system with normal cp1252 encoding.  Fixes [#10371](https://gramps-project.org/bugs/view.php?id=10371)
* Fix GraphView for bad transient parent on print file dialog and 'File already exists' dialog.  Fixes [#10372](https://gramps-project.org/bugs/view.php?id=10372)

### DIFF
--- a/GraphView/graphview.py
+++ b/GraphView/graphview.py
@@ -401,11 +401,13 @@ class GraphView(NavigationView):
                     self.printview(obj)
                     return
             svg = val.replace('.gv', '.svg')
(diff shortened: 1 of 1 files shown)


sam-m888:
Thank you

PR #102, opened 2018-01-16
Get Form Gramplet running on gramps50

prculley:
Bug https://gramps-project.org/bugs/view.php?id=10228 inspired be to look into the Forms Gramplet.
* I copied from gramps42 branch, and fixed up gpr for Gramps50 (note that navtypes=[Person] is now necessary to get the 'active-changed' signal for persons).
* I noticed that the EditForm popup started a long way from Gramps on my multi-monitor screen, and would not remember its position, so I added code to deal with that.
* I noticed that after adding or editing a Form, the changes did not appear on the main Gramplet view, so I added an update callback to fix that.

Otherwise it seems to work as well as on gramp42, tried adding new forms, and editing old ones on several different people.

@Nick-Hall I made this a PR instead of just pushing to gramps50, as this is your code and I figured you would want to review it.


PQYPLZXHGF:
It'd be nice to be able to test this with the beta.

adamvazquez82:
Is this safe to use on the 5 beta ?

prculley:
It should work fine as I developed and tested the changes on gramps5.  I
don't know why Nick (author) is taking so long to review and commit it for
general distribution.

Paul C.

Form/editform.py
```diff
@@ -77,7 +77,7 @@ class EditForm(ManagedWindow):
     """
     Form Editor.
     """
-    def __init__(self, dbstate, uistate, track, citation):
+    def __init__(self, dbstate, uistate, track, citation, callback):
```
Nick-Hall: Instead of using a callback, the gramplet should connect to the appropriate database signals.  In this way the gramplet will update if changes are made using the standard editors.

prculley: I looks to me like it would need to connect to signals for all events, sources and citations as well as the current person (already connected) to catch the case of someone manually adding the right things to make a form appear.

I judged the callback (which API seems to be there for a reason) to be simpler and to cover the common case where someone adds a new form, and would like to actually see it in the list afterword.

It's your code, feel free to accept or change my suggested changes.  I was just trying to help a user.

daleathan:
>   I only actively maintain the current version of this addon.

@Nick-Hall So what happens when you ask people to test the new version of Gramps and they can't even setup like the existing version! Works great from my side, but not everyone will download or even no how to down from here, the only thing they see is beta version fantastic, installs it and downloads the addons and goes what no form addon, back to the old version and we lose out on them testing!  I accept that you maintain only one version and most people using the test version would also, but at some stage and I would believe during a beta or release candidate an exception would be made?  Thanks for the Form addon 🥇

I'm curious about when the next release will be, what needs to be done?

Form/editform.py
Nick-Hall: @prculley OK.  It looks like I found a related bug that you didn't attempt to fix.  Some people do edit forms, but are less likely to create new forms, using the standard editors.

Nick-Hall:
@daleathan I really don't want to maintain more than one version, so I will apply the fixes to the _gramps42_ branch.  I have no objection to putting a copy in the _gramps50_ branch for testing.

The relaese of 5.0.0-beta1 was on 18th February.  The next step is to look at the bug reports and add any release blocking issue to the roadmap.  Then we need to enter a string freeze and make an announcement to the translators.  After that we should be close to a final release.

adamvazquez82:
> I have no objection to putting a copy in the gramps50 branch for testing.

Doesn't show for me as an option to install, is another step required?

sam-m888:
Thank you

@daleathan @adamvazquez82 Initial build of this addon has been added to
gramps50 please test via the builtin "check for updated addons" button.

PR #103, opened 2018-01-20
Gramps42 some bug fix commits from gramps50 branch

prculley:
These were already committed to gramps50, and are back-ported to gramps42.  A couple of minor tweaks in some of them to deal with the gpr gramps_target_version change, and to deal with other differences between gramps42 and gramps50.

Each plugin seems to work and to fix the bug associated with it in my testing.

Will need rebuilding when accepted...


sam-m888:
Thank you

Gramps42 addon rebuilt:
- [x] DenominoViso
- [x] DescendantsLines
- [x] DynamicWeb
- [x] NetworkChart
- [x] PhotoTaggingGramplet

PR #104, opened 2018-01-31
Fixes display options for dates/places in DescendantsLines reports

agnescb:
These now work as advertised.
Fixes #10392


Nick-Hall:
Merged in commit 2802ddb2a4f2ccb83c0fa343e373594d6bd6610d.

PR #106, opened 2018-02-05
Fix NetworkChart gpr to display missing dependencies correctly

prculley:
Fixes #10408
Turns out gramps42 branch doesn't provide 'uistate' to plugins, so there is no easy way to test for CLI mode.  So if dependencies are missing, will end up with a possible GUI dialog even in CLI mode.

Once I got that fixed, there was another stupid bug...


sam-m888:
Thank you

rebuild addon as https://github.com/gramps-project/addons/commit/818452300ed2a020bd25819b688c2f67a84afa12

PR #107, opened 2018-02-06
Fix PlaceCompletion for some issues

prculley:
Fixes  [#10409](https://gramps-project.org/bugs/view.php?id=10409)
Changed re compile options to avoid deprecated 'L' locale option.
Changed Glade file to allow correct transient parent.
Changed Glade file to allow keyboard control over result tree.
Changed Gdk event checks to correctly recognize click and keys.
Fix double change attempt for county/state.
Changed attempt to change county/state to a printed message instead of an error dialog; this doesn't do anything useful with our current enclosed places.  This avoids a click for every place during the scan when using the US state files.

I started this due to bug 10409, however as I began testing I discovered several problems.  The most concerning is that this tool is quite outdated in our current place structure.  While it works, the title change capability is of little utility for those users who use the automatic place title function.

The whole tool is useless unless the user has already manually enclosed his places.

The user can still use it for Lat/Long creation and update.

It may be time to rethink this tool and/or replace it entirely with something better, if someone is feeling ambitious.

### DIFF
--- a/PlaceCompletion/PlaceCompletion.py
+++ b/PlaceCompletion/PlaceCompletion.py
@@ -279,8 +279,8 @@ class PlaceCompletion(Tool.Tool, ManagedWindow):
         elif group == ('zip'):
             place.set_code(val)
         else:
-            ErrorDialog(_("Error in PlaceCompletion.py"),
-                        _("Non existing group used in set"))
+            print(_("PlaceCompletion is unable to create %s %s"),
+                  group, val)
         return place
 
     def fill_combobox(self, cmb, namelistopt, default) :
@@ -401,7 +401,7 @@ class PlaceCompletion(Tool.Tool, ManagedWindow):
         if findregex :
             try:
                 #compile regex aware of locale and unicode
-                self.matchlatlon = re.compile(findregex,re.U|re.L|re.M)
+                self.matchlatlon = re.compile(findregex, re.M)
                 latlongroup = ['lat', 'lon']
                 for group in latlongroup :
                     if findregex.find(r'(?P<'+group+r'>') == -1 :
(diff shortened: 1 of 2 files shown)


Nick-Hall:
The new place hierarchy largely eliminates the need for this tool.  Information is no longer duplicated, so we no longer get inconsistencies which need fixing.  Also changes to all titles lower in the hierarchy can now be made by changing the name of a single place.

Even the latitude/longitude functionality is not as necessary with improvements to the geography views and creation of the GetGOV gramplet.

You may be interested in looking at a useful data migration tool for place titles that Jérôme created in feature request [9907](https://gramps-project.org/bugs/view.php?id=9907).

prculley:
The last time I used this tool was in the 3.x time frame; it seemed quite useful then as I often load Gedcom data and want to mass update my places.  It seems no longer useful for that.  I think something that looks at places en mass and uses GetGov or other data to build up properly enclosed places with lat/lon etc. is what we need.  But this conversation is kind of off topic.

Regarding this tool; maybe we should stop listing it?

GaryGriffin:
Personally, I still use PlaceCompletion for bulk (per US state) Lat/Lon. The US GeoNames data files have more details (cemeteries and churches) which GOV database does not have. None of the Actons in the tool work for me on my Mac [10004](https://gramps-project.org/bugs/view.php?id=10004) , so I cannot speak for their usefulness.

Nick-Hall:
Yes.  I expect that quite a few people still use it for Lat/Lon.

sam-m888:
Nick also created in [8812](https://gramps-project.org/bugs/view.php?id=8812) an Experimental place gazetteer for gramps 42.

sam-m888:
Edited the Place completion user page to mention the issues with using this in Gramps50 

https://gramps-project.org/wiki/index.php?title=Place_completion_tool#Issues

sam-m888:
Thank you

addon rebuilt in https://github.com/gramps-project/addons/commit/1aef34456b38765b4dd7664e45f4d829369904f7

PR #108, opened 2018-02-11
[DescendantsLines]Centered lines are not centered and can overrun the container[gramps42]

sam-m888:
[DescendantsLines graphical report] Centered lines are not centered and can overrun the container

Fixes: #[10398](https://gramps-project.org/bugs/view.php?id=10398)


sam-m888:
Thank you :)

PR #111, opened 2018-02-15
GoogleEarthWriteKML output log Missing coordinates[gramps50]

sam-m888:
Provide feedback in log that coordinates are missing for place


adamvazquez82:
Turns out all the places I was testing did not have them!

Thanks

PR #113, opened 2018-02-23
Fix Descendant Book crash

prculley:
Fixes [#10463](https://gramps-project.org/bugs/view.php?id=10463)
Addon was not properly updated for change in gramps50 filters apply calling parameters.


sam-m888:
Thank you.

Addon rebuilt for gramps50 as:
https://github.com/gramps-project/addons/commit/545291c3e8960b2f391e3de037bed66ce850841b

PR #114, opened 2018-02-24
Fix Calculate estimated dates addon[gramps50]

prculley:
fixes #10462;  Issue with closing the Progressmeter twice
During testing I found other issues, decided to fix in one PR...

- Issue with HandleError when 'other' person was None
- Db corruption due to wrong deletion of a source if user used "Remove previously added ..."
- Db corruption due to NOT deleting Citations for a deleted source if user used "Remove previously added ..."
- Issue with calling for db rebuild during a transaction (two places)

I suspect that the db corruption issues probably messed up a few db's over the last 5 years...
This should be backported to gramps42 branch...

### DIFF
--- a/CalculateEstimatedDates/CalculateEstimatedDates.py
+++ b/CalculateEstimatedDates/CalculateEstimatedDates.py
@@ -255,6 +255,7 @@ class CalcToolManagedWindow(PluginWindows.ToolManagedWindowBatch):
                 self.results_write(_("Removing old estimations... "))
                 self.progress.set_pass((_("Removing '%s'...") % source_text),
                                        num_people)
+                supdate = None
                 for person_handle in people:
                     self.progress.step()
                     pupdate = 0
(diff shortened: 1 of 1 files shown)


sam-m888:
Tested and works well with all the setting changes I tried.

> I suspect that the db corruption issues probably messed up a few db's over the last 5 years...
> This should be backported to gramps42 branch...

Yes that would be a good idea!

Only issue I could spot had nothing to do with your changes.  When you  try to expand out the **Select** tab the result section does not fill the all the space.
calculate-estimated-dates-select-tab-50

sam-m888:
Thank you

Addon rebuilt for gramps50 as:
https://github.com/gramps-project/addons/commit/2046d7907eceb5cc070c8f52dcf818a61d058f47

prculley:
I committed the same change to gramps42, if you want to rebuild the addon there too...

sam-m888:
Thanks

rebuilt addon for gramps42
https://github.com/gramps-project/addons-source/commit/83b6ac294eec286665b4e6765c022d687b2a6fea

PR #115, opened 2018-02-24
Add View specific icons to four addons[gramps50]

prculley:
Sam mentioned that it would be nice if the Icons for the various addon charts were different...
Ask, and ye shall receive...

If you want this on gramps42, it is almost the same, but you have to drop the "if locals().get('uistate'):" as uistate is not available on gramps42 branch.  This means that if started in cli mode on gramp42 branch with the plugin present, you may get some (I think harmless) side effects due to the gtk imports.

### DIFF
--- a/GraphView/MANIFEST
+++ b/GraphView/MANIFEST
@@ -1,0 +1,1 @@
+GraphView/gramps-graph.svg
--- a/GraphView/gramps-graph.svg
+++ b/GraphView/gramps-graph.svg
@@ -1,0 +1,244 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:xlink="http://www.w3.org/1999/xlink"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   width="51.200001"
+   height="51.200001"
+   id="svg1771"
+   sodipodi:version="0.32"
+   inkscape:version="0.92.2 (5c3e80d, 2017-08-06)"
+   version="1.0"
+   sodipodi:docname="gramps-graph.svg"
+   inkscape:export-filename="/home/benny/programms/gramps/myicons/gramps_try4.png"
+   inkscape:export-xdpi="90"
+   inkscape:export-ydpi="90">
+  <title
+     id="title13529">Graphview Icon</title>
+  <defs
+     id="defs1773">
+    <inkscape:path-effect
+       effect="bspline"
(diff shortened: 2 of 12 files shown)


sam-m888:
@prculley You're a star 🥇 🥇 👍 

No more [mystery meat navigation](https://en.wikipedia.org/wiki/Mystery_meat_navigation) with choose the the right view!  Thank you.

new-addon-icons-normal-toolbar-50

new-addon-icons-text_and_icon-toolbar-50

@prculley Can you silence the following outputs from  "print(fname, pixbuf, file=sys.stderr)" please?

> C:\Users\me\Documents\gramps-test-profile\gramps\gramps50\plugins\GraphView\gramps-graph.svg <GdkPixbuf.Pixbuf object at 0x0cfa6738 (GdkPixbuf at 0x09a7f8f8)>
> C:\Users\me\Documents\gramps-test-profile\gramps\gramps50\plugins\TimelinePedigreeView\gramps-timelinepedigree.svg <GdkPixbuf.Pixbuf object at 0x0cfa6f30 (GdkPixbuf at 0x09a7b5f8)>
> 

Tested on Ubuntu 16.04 & Win 10(32bit)

 - [ ] Separate issue when building the addon it does not pickup the presence of the ***.svg** file! I looked at the other addons and to include extra files the [DynamicWeb addon](https://github.com/gramps-project/addons-source/tree/master/DynamicWeb) list them in the **MANIFEST** file. Ok tested this and it works. If you add a file called that with the contents shown below; it allow me to build the addon correctly.(only checked on 5.0)

Filename:
`MANIFEST`

File Contents:
`GraphView/gramps-graph.svg`

prculley:
@sam-m888 Lets see if this works now...

sam-m888:
> Lets see if this works now...

@prculley Had two issues 

 - [ ] The version numbers of both addons where lower than the current versions (so I adjusted this locally)

 - [ ] Initially I believed that the lower version number was why the listing was not created but after adjusting only the ***.tgz** was created, it seems the build script able to find the addon to rebuild it but not find it when the listing needs to be made!

sam-m888:
Checking the other addons to see how they handle it.

PQYPLZXHGF:
36635763-f04bd8d8-1a0e-11e8-9dcd-1a6b2e84927d

@sam-m888 Where do I find the option to show text under the icon like in your image(above)? It's one of the reasons I'm still on 3.49 and gramps 4 made it more difficult to navigate like you mention! 

Even shown in the manual for [gramps 3.4](https://gramps-project.org/wiki/index.php?title=Gramps_3.4_Wiki_Manual_-_Main_Window#Introduction):

mainwinannotated

adamvazquez82:
> @sam-m888 Where do I find the option to show text under the toolbar icons like in your image(above)? It's one of the reasons I'm still on 3.49 and gramps 4 made it more difficult to navigate like you mention!

@PQYPLZXHGF I did this aIso to my toolbar for much the same reason.

Look in the following file:
 [gramps/gui/viewmanager.py](https://github.com/gramps-project/gramps/blob/ef395f49fe62ab6a69642108e57596c36f0e1049/gramps/gui/viewmanager.py#L462) 

and find 

`if self.show_toolbar:` 

then add following line beneath it:

`        self.toolbar.set_style(Gtk.ToolbarStyle.BOTH)`

when you restart gramps and you should have text under the icons. 🦄🌈

If you upgrade you'll need to reapply.

sam-m888:
Thank you

sam-m888:
@adamvazquez82 @PQYPLZXHGF  Spot on with how I did it.  I'll raise a feature request and add detail about how on it.  👍

PR #119, opened 2018-02-28
Fix Clipboard Gramplet exception with data present when changing db.

prculley:
Issue [#10473](https://gramps-project.org/bugs/view.php?id=10473)

This is the second part of bug10473 (first is https://github.com/gramps-project/gramps/pull/566).
This bug occurs when there is data in the clipboard Gramplet and the db is changed.  An exception occurs because the data was in bytes mode instead of str.


sam-m888:
Thank you

PR #122, opened 2018-03-04
[CollectionsClipboardGramplet]Fix for db change and other issues[gramps50]

prculley:
Issue [#10475](https://gramps-project.org/bugs/view.php?id=10475)
* Clipboard Gramplet was overriding the 'refresh_objects' method of the listview, I think the original idea was to show clipboard items that were not valid for the current db.  Unfortunately, if a user tried to drag on of these invalid objects (they had a red X for icon) to a reasonable destination, it would create a corrupted db, and usually crash Gramps with HandleError.  The original Clibboard (not Gramplet) was removing these clips from the view entirely, so they could not be erroneously dropped.
* I noted that with some types of clips, the save/restore function of the Clipboard Gramplet would fail because the configure routines don't properly handle a generalized bytes element.  In particular, I saw a '%' in the bytes stream that was messing with the parsing of the .ini file.  I changed the data storage of the pickled data to a hex string to avoid this.  I also put the entire load block in a try/except to cover the remote case of loading an older versions bytes string and failing to decode it as hex.  This also covers some other possible corruptions.

* The Clipboard Gramplet was not restoring the db name and id to Text clips, which was confusing; after a Gramps restart, the text data all appeared to be coming from the current db.

There were other issues related to the original Clipboard (not Gramplet) code, these are dealt with in https://github.com/gramps-project/gramps/pull/569.  I think this stands alone, but obviously won't fix everything without the other PR.

### DIFF
--- a/ClipboardGramplet/ClipboardGramplet.gpr.py
+++ b/ClipboardGramplet/ClipboardGramplet.gpr.py
@@ -6,7 +6,7 @@
 
 register(GRAMPLET,
          id="Clipboard Gramplet",
-         name=_("Clipboard Gramplet"),
+         name=_("Collections Clipboard Gramplet"),
          description = _("Gramplet for grouping items"),
          status = STABLE,
          version = '1.0.31',
--- a/ClipboardGramplet/ClipboardGramplet.py
+++ b/ClipboardGramplet/ClipboardGramplet.py
@@ -26,7 +26,7 @@
 #
 #-------------------------------------------------------------------------
 import pickle
-
+from binascii import hexlify, unhexlify
 #-------------------------------------------------------------------------
 #
 # Gramps modules
@@ -39,9 +39,8 @@ except ValueError:
     _trans = glocale.translation
 _ = _trans.gettext
 from gramps.gen.plug import Gramplet
-from gramps.gui.ddtargets import DdTargets
 from gramps.gui.clipboard import (MultiTreeView, ClipboardListModel,
-                           ClipboardListView, ClipText)
+                                  ClipboardListView, ClipText)
 
 #-------------------------------------------------------------------------
 #
@@ -68,25 +67,6 @@ def unescape(data):
         data = data.replace("\\n", "\\x0a")
     return data
 
-#-------------------------------------------------------------------------
-#
-# ClipboardListView2 class
-#
-#-------------------------------------------------------------------------
-class ClipboardListView2(ClipboardListView):
-    """
-    Subclass ClipboardListView to override refresh_objects.
-    """
-    def refresh_objects(self, dummy=None):
-        def update_rows(model, path, iter, user_data):
-            """
-            Update the rows of a model.
-            """
-            model.row_changed(path, iter)
-        # force refresh of each row:
-        model = self._widget.get_model()
-        if model:
-            model.foreach(update_rows, None)
 
 #-------------------------------------------------------------------------
 #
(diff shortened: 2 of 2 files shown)


wroldwiedbwe:
Attempted to use this along with the other pr and sorry to say my as you call them previous "clips" disappeared and each of my 4 clipboard gramplets(that have different name titles and different groups of "clips") became identical, lucky I was only testing this on a copy!  Did you intend to change the behavior of the gramplet?

IMPO naming it a clipboard gramplet is misleading and confusing especially as Gramps has one builtin, it would be better to think of this gramplet as a collection gramplet that allows you to group together "clips" for use in data entry that I use for each of my family lines and it took a while to see that possibility especially handy with attribute clips and source and citations that keep getting reused.

You can [read](http://gramps.1791082.n4.nabble.com/Updated-Scratchpad-new-Clipboard-Gramplet-td2306505.html) from the original developer and see his demo [video](http://www.youtube.com/watch?v=hoTvwvteJjk) about how the gramplet is used.

Thanks for fixing this for everyone. 🌞🌞🌞 GRAMPS is a great program. Looking forward to 5.0 being released.

prculley:
I had regarded the different data in different Gramplets as a bug, not a feature, not realizing that someone would use multiple Gramplets in this fashion.  But I broke that change out as a separate commit, so it is easy to leave it out of the changes.

The loss of saved data from your older system is expected, as I changed the format the data is saved in.  This would only happen one time.  This was due to an issue where, depending on the data contents, it was possible to crash Gramps so that it would not restart, and not recover except by deleting the appropriate .ini file.  Not very user friendly. 

It is possible to keep your previous data, but not easy.  You would have to open Gramps in debug mode with the previous version of the Gramplet installed, then install the new version of the Gramplet, then use the Addon manager and its (debug only) 'Reload' button to 'register' the updated Gramplet, then quit Gramps.  The data should be there the next time you start Gramps.

Of course, since the data is saved in .ini files, if you ever used the newer Gramplet version, then these files would be overwritten, so your data may already be lost, unless you made a backup of those files with an external tool...

sam-m888:
@wroldwiedbwe Thanks for the example of how you use it, I'll update the [support page](https://gramps-project.org/wiki/index.php?title=ClipboardGramplet) with an image showing that when I get a moment.

> IMPO naming it a clipboard gramplet is misleading and confusing especially as Gramps has one builtin, it would be better to think of this gramplet as a collection gramplet

@wroldwiedbwe @prculley If we were to rename the gramplet then I believe the name should still include the word clipboard so existing users can find it again.   So how does renaming it to the "**Collection Clipboard**" gramplet sound? I suggest before the release of Gramps 5 would be the best time to do this as nobody would have it installed for that version yet.  What do you think?

wroldwiedbwe:
> I had regarded the different data in different Gramplets as a bug,

@prculley Understandable if you've never used it, thanks for changing it back to what it was 👍 Did the fact the gramplets help button links to the user manual instead of [here](https://gramps-project.org/wiki/index.php?title=ClipboardGramplet) make you believe it was just another clipboard in the form of a gramplet?

@sam-m888  Cool & thanks for the link to the support page!    

> So how does renaming it to the "Collection Clipboard" gramplet sound?

@sam-m888 Sounds good but I'd like to suggest "Collection**s** Clipboard" with an "s" as it identifies the purpose a bit more 👍

sam-m888:
Thank you

PR #124, opened 2018-03-06
Update ImportGramplet to work with gramps50

prculley:
Fixes [#6752](https://gramps-project.org/bugs/view.php?id=6752)

I suspect that this Gramplet hasn't worked in a long time; I'm not even sure it really has much utility, but @sam-m888 seems to think it might need to work...

I did not test if it can be back-ported to gramps42, although I suspect it can.

### DIFF
--- a/ImportGramplet/ImportGramplet.py
+++ b/ImportGramplet/ImportGramplet.py
@@ -24,8 +24,9 @@
 #
 #------------------------------------------------------------------------
 import time
-from io import StringIO
+from io import BytesIO, StringIO
 from xml.parsers.expat import ParserCreate
+from gi.repository import Gtk
 
 #------------------------------------------------------------------------
 #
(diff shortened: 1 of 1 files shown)


sam-m888:
> @sam-m888 seems to think it might need to work...

Just testing each of the addons and raising issues if needed along with updating the images for the wiki, all the while trying to decide if the issue is bad enough to delist/hide the addon if needed, but you've been fixing them as I've been finding them, Thank you for fixing them.

As to whether it has much utility , who's to say as we have no real telemetry for Gramps addon usage!

kulath:
I came here to see what this does, and it looks as though the comment in the run method has just been copied from somewhere else.

sam-m888:
@kulath Spot on identical text located in gramps  [ageondategramplet.py
](https://github.com/gramps-project/gramps/blob/cf42e5a4b847e6881f180cd1b776fb02234768be/gramps/plugins/gramplet/ageondategramplet.py#L80)

prculley:
@sam-m888 I'm surprised Github doesn't have a way to tell the administrator the number of downloads of a particular repository, or even a portion thereof.  Some of their graphs imply at least some metrics are collected.

PR #125, opened 2018-03-08
Fix Gedcom Extensions Addon for exception on gramps50

prculley:
Fixes [#10472](https://gramps-project.org/bugs/view.php?id=10472)


sam-m888:
Thank you

PR #126, opened 2018-03-08
Fix DynamicWeb for exception on config setting

prculley:
Fixes [#10483](https://gramps-project.org/bugs/view.php?id=10483)

Bug was side effect of the changes for dark theme support...
2nd bug was side effect of changes for place editor...


sam-m888:
Tested and does not work see bug report.

adamvazquez82:
**preferences.place-lang** was removed in https://github.com/gramps-project/gramps/pull/368

prculley:
@sam-m888 @adamvazquez82 Thanks for testing.
Seems I should wipe out my gramps.ini once in a while; I was not seeing the bug because I still had the old 'preferences.place-lang' in my gramps.ini.

sam-m888:
@prculley @adamvazquez82  Thank you

PR #127, opened 2018-03-14
[GraphView] fix generation spinners and popup [gramps50]

vantu5z:
- fix generation changing (toolbar spinners)
When user change generation spinners values - tree rebuild (populate) will be called with a little delay. So we have more GUI responsiveness and don't have freeze while change values.
I don't know if this is right way to add delay, may widget already have this feature.

- fix empty submenu on person which doesn't have children.


sam-m888:
> fix generation changing 

Yes the difference with the patch makes it easier to change as with larger generations it's more noticeable.

sam-m888:
Thank you

PR #128, opened 2018-03-17
[Quilt View]Add popup menu and other request.

SNoiraud:
Fixes [#10488](https://gramps-project.org/bugs/view.php?id=10488)

Add popup for people
Add popup for Families
The popups are copied from Graphview and adapted to this view. Thanks to Ivan.
Be conform to the scheme colours defined in preferences->colors
Add a tooltip when hover a family
Add a search field to facilitate navigation on big trees
Use the defined color for the home person.
Add a progess meter


johnbickmore:
You're a star 🌟🌟🌟🌟🌟🌟🌟.

Thanks 👍

SNoiraud:
I don't see anything else to do.
Do you think 3 seconds is a good value to see the tooltip ?

SNoiraud:
I am trying to add a search bar to select peope easily in the tree.

sam-m888:
Family tooltip on hover.
familytip

sam-m888:
Family Edit - context menu.
familyedit-context-menu

Person Edit - Context menu
personedit-context-menu

sam-m888:
Search field to facilitate navigation on big trees

seachfield

sam-m888:
> Be conform to the scheme colours defined in preferences->colors

Uses colors from defined in the colours tab.  Shown here using the "dark colours" on a light theme desktop.

darkcolours

sam-m888:
> Use the defined color for the home person.

@SNoiraud Spotted an issue, when I switched to the "dark colours" theme the home person colour did not change to the one defined in it.

darktheme-home-person

SNoiraud:
I understand that. Is there a signal emitted when we change the colours ?
If you only change the colours, I must rebuild the tree. I look at this.
We have the same problem with Graphview. Using the prior and next bookmark solves the situation.

fxtmtrsine:
Imported a ged file into a new tree, then switched to quilt view and this happens, probably because no home person set?

> 136678: ERROR: grampsapp.py: line 143: Unhandled exception
>   File "C:\Program Files (x86)\GrampsAIO32-5.0.0-beta1\gramps\gui\viewmanager.py", line 1093, in view_changed
>     self.__change_page(page_num)
>   File "C:\Program Files (x86)\GrampsAIO32-5.0.0-beta1\gramps\gui\viewmanager.py", line 1106, in __change_page
>     self.active_page.set_active()
>   File "C:\Program Files (x86)\GrampsAIO32-5.0.0-beta1\gramps\gui\views\navigationview.py", line 164, in set_active
>     self.goto_active(None)
>   File "C:\Program Files (x86)\GrampsAIO32-5.0.0-beta1\gramps\gui\views\navigationview.py", line 199, in goto_active
>     self.goto_handle(active_handle)
>   File "C:\Users\me\Documents\gramps-test-profile\gramps\gramps50\plugins\QuiltView\QuiltView.py", line 479, in goto_handle
>     if handle not in self.people.keys():
> AttributeError: 'list' object has no attribute 'keys'
> 

By the way if I press the Home icon on the toolbar should the quilt jump to that person?

fxtmtrsine:
That search bar button is very weird, should it actually be two separate buttons  like "clear" and the "results" for the other one?

When you attmpt to use the person menus, the person jumps to the top left corner while the menu stays where it is, took a little while to notice! How about when you use the menu the quilt stay where it is until you exit the menu?

SNoiraud:
I was adding a Progress meter dialog to see if something is working when we have a big database.
The load was done twice the first time in the current view. problem ?
So if the quiltview was the last opened and it is the current at the gramps start, we load the data twice.
If quitview is not the current view when we load the database, the data is loaded once.
To avoid that, I added some test. I think this is a problem when we load or change the database.

SNoiraud:
So, we now only bookmark a person if we call the person editor.

SNoiraud:
I think it's finished. I change the version number.
Any comments?

sam-m888:
On 19 March 2018 at 08:58, Serge Noiraud <notifications@github.com> wrote:

> I understand that. Is there a signal emitted when we change the colours ?
> If you only change the colours, I must rebuild the tree. I look at this.
> We have the same problem with Graphview. Using the prior and next bookmark
> solves the situation.
>
>
Possibly monitor the colours and if changed trigger a rebuild?

SNoiraud:
> Possibly monitor the colours and if changed trigger a rebuild?

We have 17 colours to check at each time we move the mouse. Too expensive in time and cpu.
For my 8600 people tree, it takes aproximately between 8 and 10 seconds to load it.
When I move the mouse hover my tree, Draw the path is already expensive (between 1 and 5 seconds).
Once we choose the colours, I think we practicaly never change them after that.
As I said, the solution is to go back then go forward in the bookmarks or change the view then return to this view.

johnbickmore:
> Any comments?

Looking at that message to the right of the "Clear" button I think on the example gramps database the total should be 1853  and the filtered list should show the total matching the filter eg for "nas" it would be 5.  


q1


q2


Thanks for improving this 👍

SNoiraud:
> on the example gramps database the total should be 1853.this represents the total number of connected people in the quilt right?

Yes. This is for the active person. For the exemple database, if you select القاسم, you'll get : 

quiltview message

> as the total number of people in the example gramps is 2103 and the filtered list should show the total
> matching the filter eg for "nas" it would be 5.

Yes

SNoiraud:
> Possibly monitor the colours and if changed trigger a rebuild?

The best way could be to emit a signal when one of the colours change. The view would connect to this signal. But this is a gramps feature request.

SNoiraud:
For the message, we could show : 
You have X filtered people, you have Y people shown on the tree and Z people in your database.
like that : 
quiltview message 2

sam-m888:
Thank you

PR #131, opened 2018-03-24
[GedcomExtensions]fix for filtered data[gramps50]

prculley:
Fixes [#10515](https://gramps-project.org/bugs/view.php?id=10515)

This should be cherry-picked to Gramp50 etc. as well.


sam-m888:
Thank you.

Meant to write [gramps42] & not [gramps50] in commit message!

PR #133, opened 2018-04-23
[DynamicWeb]Fix to avoid duplicate citation link buttons on events[gramps50]

prculley:
Changes the report so that the individual person page doesn't get duplicate citations in the Event column of the Events tab.  From a complaint on the gramps-users mailing list https://sourceforge.net/p/gramps/mailman/message/36301205/

Should be cherry-picked to gramps42 when accepted...


sam-m888:
Thank you

Addon cherry picked to gramps42 and rebuilt for both.

PR #137, opened 2018-05-01
Fixes for ListeEclair addon to work with Gramps50 branch

prculley:
Fixes [#10559](https://gramps-project.org/bugs/view.php?id=10559)
Several fixes as already applied to gramps42 and a couple more to make it work under gramps50.

Progressmeter converted to use 'user' version (to avoid 'no transient parent' warning).
'Entire Database' filter option added.
A couple more placetypes added to report output 'Town', 'Village' etc.
Fixed HandleError crash when trying to lookup non-Event handles as Events.
Fixed crash for persons with no surname.
Fixed issue with invalid dates.


sam-m888:
Thank you

PR #109, opened 2018-02-14
Add new PreRequisites Checker Gramplet[gramps50]

sam-m888:
A diagnostic tool to help work out if users have the [prerequisites](https://github.com/gramps-project/gramps/blob/master/README.md) installed as well as the latest version of Gramps

Also helps on OS's like windows and mac where it's difficult for users to use the command line to get information on gramps via ( gramps -v )!

- Retrieved and updated for Gramps 5

- Added simple Gramps version check

- Displays additional requirements for development & translation.

- Expanded to check all known third-party addons prerequisites

- Tested with Gramps versions: 3.3 to master (Only checked that it loads and displays on older versions)

- Correct min versions for GTK & pygobject (prculley)

- Add check for editexifmetadata addon (prculley)

- Add check for PIL to latexdoc type addons:genealogytree(prculley)

- Add check for PhotoTagging & lxml gramplets (PQYPLZXHGF)

- Add MongoDB (Dale)

context-menu-option

50prerequisites checker gramplet01
50prerequisites checker gramplet02

PR #144, opened 2018-05-13
[PrerequisitesChecker]Fixes UnboundLocalError[gramps50]

sam-m888:
Fixes [#10592](https://gramps-project.org/bugs/view.php?id=10592)

Spelling of [successs](https://github.com/gramps-project/addons-source/pull/109#pullrequestreview-119643008) > success  ( @adamvazquez82 )


SNoiraud:
Has I said in #10592: 
I found another problem with the HTMLView. This view works for me but the tool report WebKit is not installed.
We need to change the line 1113 :
The version for this module is not "__version__" but "_version"

We have the same problem with pydotplus near line 1040:
The version for this module is not "__version__" but "pyparsing_version"

sam-m888:
@SNoiraud fix pushed and addon rebuilt, Thanks can you check it works for you please?

PR #149, opened 2018-05-25
[GetGOV]Fix config setting name[gramps50]

sam-m888:
Fixes [#10603](https://gramps-project.org/bugs/view.php?id=10603#c54658)

https://gramps-project.org/bugs/view.php?id=10603#c54658


sam-m888:
@GaryGriffin Thank you

PR #151, opened 2018-05-27
fix travis: addons tests not being run because make.py is executable

azrdev:
from travis logs at e.g.
<https://travis-ci.org/gramps-project/addons-source/builds/384343076>:

> nose.selector: INFO: /home/travis/build/gramps-project/addons-source/make.py is executable; skipped


prculley:
Gentlemen; the make.py file is NOT a test, and SHOULD be skipped.  There is currently one actual test in this repo; (sqlite has one), so it doesn't look like much has been run.  But the row of '---' and the "ran 1 test" line of the Travis log indicates that it is working.

I believe that this PR should be closed.

PR #152, opened 2018-06-03
get_from_name_and_handle doesn't exist anymore

SNoiraud:
Fixes #10622


SNoiraud:
I found another problem when trying to merge databases with gramps ID ?%05d
The import file is also ?%05d.
The result is incorrect: 
importmerge 1

SNoiraud:
The problem is in the import_as_dict method: We should use the preferences gramps ID format.
I propose the following patch in gramps core
[import_as_dict.patch.txt](https://github.com/gramps-project/addons-source/files/2065893/import_as_dict.patch.txt)

PR #156, opened 2018-06-17
Thisdayinfamilyhistorygramplet

ThariusDawnbreaker:
Gramplet tested with the example family tree as well as my own under Gramps 5.0 beta 1.3 Windows AIO64

An empty tree or a tree with no events on the current date should produce:
this day in family history gramplet - sample output without content

A tree with at least 1 event on the current date should produce something like:
this day in family history gramplet - sample output with content

You can see in the above image that missing parents in families will not result in a crash as it did previously.  Other missing data items should similarly be handled.

The configure dashboard settings should be persistent.  I tested them against my data and all appears to be functional, let's hope your mileage is similar :)
this day in family history gramplet - configure dashboard default settings

Tested to refresh on database updates and on configuration updates.


ThariusDawnbreaker:
It was a good idea.  I had been cutting and pasting the output to my family
tree group on facebook and once I added that date in there it looked so
much better almost immediately.

Thanks!

wroldwiedbwe:
Gramps needs https://github.com/gramps-project/gramps/pull/629 applied if you want to use the configure dashboard settings easily.

sam-m888:
Looks to be a very useful gramplet 👍 

Thank you.

daleathan:
@ThariusDawnbreaker @sam-m888 Notice on your help page that you say it was built for gramps 5.0, any chance this can be added and built for gramps50 branch?

prculley:
Now that the other PR that fixes the config options is accepted, this CAN
be built for gramps50.  But it should probably wait until another release
of the Gramps50 branch, so users will not have issues with the ugly config
options.  We should be pretty close to a final release of 5.0.0, just
waiting on a few more translations and permission to go ahead.

Paul C.

daleathan:
>  But it should probably wait until another release of the Gramps50 branch, so users will not have issues with the ugly config options

Good choice 👍

PR #159, opened 2018-08-01
ADD: houses organised by number/locality, residents with multiple rec…

todef96:
…ords listed as date range


sam-m888:
@todef96 Added to gramps50 .

PR #163, opened 2018-08-30
Gary griffin geo name creation[gramps50]

GaryGriffin:
Initial checkin for GeoName. Bug reference https://gramps-project.org/bugs/view.php?id=10776

### DIFF
--- a/GeoName/geoname.gpr.py
+++ b/GeoName/geoname.gpr.py
@@ -1,0 +1,44 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2015      Nick Hall
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+#
+#
+
+#------------------------------------------------------------------------
+#
+# GetGOV Gramplet
+#
+#------------------------------------------------------------------------
+
+register(GRAMPLET,
+         id = "GeoName Gramplet",
+         name = _("GeoName"),
+         description = _("Gramplet to get places from the GeoNames database"),
+         status = STABLE,
+         version = '1.0.0',
+         gramps_target_version = '5.0',
+         fname = "geoname.py",
+         gramplet = 'GeoName',
+         height = 375,
+         detached_width = 510,
+         detached_height = 480,
+         expand = True,
+         gramplet_title = _("GeoName"),
+         help_url="GeoName Gramplet",
+         include_in_listing = True,
+        )
--- a/GeoName/geoname.py
+++ b/GeoName/geoname.py
@@ -1,0 +1,272 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2015      Nick Hall
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+#
+
+"""
+GeoName Gramplet.
+"""
+
+#------------------------------------------------------------------------
+#
+# Python modules
+#
+#------------------------------------------------------------------------
+from urllib.request import urlopen, quote
+from xml.dom.minidom import parseString
+
+#------------------------------------------------------------------------
+#
+# GTK modules
+#
+#------------------------------------------------------------------------
+from gi.repository import Gtk
+
+#------------------------------------------------------------------------
+#
+# Gramps modules
+#
+#------------------------------------------------------------------------
+from gramps.gen.plug import Gramplet
+from gramps.gen.db import DbTxn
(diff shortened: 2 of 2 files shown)


prculley:
A few comments based on a quick code read;
Are you as author really an alias for Nick Hall and copyrighting this as of 2015 ;-) ??

Do you really want all the Gramps users to be using your GeoNames account?  I suspect if more than a few users were trying to work at once, the account limits would kick in, causing annoyance at best.

I don't see any code to handle errors from GeoNames or the urlopen command; no network connection, or too many queries to GeoNames, for instance.

As it happens I am in the middle of coding up a somewhat more ambitious 'Place Cleanup' addon, one of which features is to use GeoNames to fill in places like you are doing here.  So I'm happy this is an open source submission, (as all Gramps related code is) so I can take advantage of some of your work!!!  Perhaps we should collaborate on ideas?

Nick-Hall:
Both of you could also start to think about integrating this and the GetGOV gramplet in the core code.

GaryGriffin:
I guess I am busted for using GetGOV as a basis for GeoName add-on . I didn’t thoroughly check the copyright area.

Account name: it is required to have an account name to use the REST API. I didn’t see (practically) an option of not providing this in the code and requiring each user to sign-up and use their own. I did think about creating another login, say ‘gramps5’ but didn’t yet. But I am open to ideas. I tried to use concurrently and never saw a clash, but didn’t hit the limits.

Error handling: Yes, you are completely right. I viewed this as the first version and wanted to get feedback. 

My problem with GeoNames is that it does not provide the Type for the entity. I pondered adding ‘ADM1’ and ADM2 as the Place Type but felt that was a worse option than Unknown. The benefit of GOV gazetteer is that it does provide a Type (multi-lingual). And occasionally the GeoName entity has the same name for ADM1 and ADM2 - I chose to maintain this duplicity. I didn’t see an easy way to have some data from one gazetteer and some from another (and still be able to tie back). Maybe if the gazetteer ID’s were attributes of the Place rather than the Place ID.

I am not a Python programmer (all of my code looks like 70’s Fortran) but decided that this was a project I could potentially swallow in my spare time. I am more than happy to share ideas. If this add-on doesn’t go forward due to a better solution, that is fine with me. I have used PlaceCompletion add-on w GeoNames (US state datafile) and that did a lot of place cleanup for me (just adding Lat/Lon) but many cases of duplicate names it didn’t work well. I had to exclude lots of matches to keep data correct.

GaryGriffin:
Some initial concerns on Place and its tools as it exists today
1) One of the current challenges with getGOV is that it uses its own naming for grampsID. If the user reorders, it loses sync with GOV gazetteer
2) This challenge is also present with this GeoName add-on. 
3) There is no way to have a single place that has data from both gazetteers given this conflicting use of the gramps ID
4) PlaceCompletion value seems to only be to update lat/lon. But it is not for casual user since they have to upload datasets.
5) If GeoName is released, then it may lead to more migration work ahead, depending on how you implement your Place Cleanup. 

Can you tell me what is your goal of your cleanup process. I can imagine various levels of engagement
A) For existing places, add place structure based on a gazetteer (GeoNames, for instance). This could be a valuable tool for those people who have a flat place list and have not built based on the hierarchy structure. This assumes the existing title has the fully qualified name that can then be parsed to create the hierarchy. It could also add data from the gazetteer (lat/lon, code)
B) When adding a new place, reach out to a gazetteer for date on the place (lat, lon, code). This may require the user to identify the enclosing entity
C) When adding a new place, reach out to a gazetteer to get the higher structure. Tie this in to any existing structure (add a new city to an existing county, for instance). It would have to be able to identify how the existing places tie to the new place structure. This will require leveraging the gazetteer ID within the gramps place data, I think.

Is your intention to add a place attribute for gazetteer ID and not use a unique grampsID? If so, then GeoNames should not be released as it would not align. And GetGOV would need updated. 
Is your intention to change the Place Add operation to query GeoNames gazetteer? If so, then GeoNames should not be released as it is redundant.

When I look at the 2 gazetteers, I see very different applications. GOV provides Europe with historical hierarchies and accurate types. GeoNames provides administrative hierarchy but no type info and is richer for US than Europe. All of the examples I compared it would be non-trivial to match entries in the 2 different gazetteers into a single hierarchy in gramps. Not sure how you would plan to resolve this. The city matches well, but the entities enclosing are not easy to match.

prculley:
The 5.0.0 reorderID tool has been fixed up to make it possible to NOT reorder GrampsIDs that don't match the usual pattern, which allows for the preservation of GOV or (potential) GeoNames IDs.

I've already submitted a request for 5.1.x time-frame to add place attributes (actually dated attributes) which would make dealing with the GOV and GeoNames IDs easier; they would not have to overload the GrampsID.

As it is, I'm trying to design my new addon such that it doesn't depend on the GrampsID values.  I'd be doing all searches by name.  It would not matter if IDs were changed. I'm tentatively thinking to at least make new places with the GOV/GeoNames style IDs, at least until Gramps supports place attributes.

My major goals, are to 1) allow for easier find and merge of places already in your db; the find would look for 'complete' places to use as the primary merge with the current place.
2) Provide a Geonames search, although I intend to let the user pick from potential multiple search results of the GeoNames search.  And allow the user to modify the search terms if necessary to reduce the results list.  (Try a search for 'Paris', if you don't add 'France' or 'TX' etc., you get a lot of results.)
3) allow GeoNames to provide place data, hierarchy, lat/lon, and alt names as your addon does.
4) when a Gazetteer search is done, allow the user to select primary name, and which alt names to keep, which of multiple postal codes to keep, and also to select from place types in the ComboEntry box like the place editor uses today.  The preselected value would be a best guess based on algorithm similar to yours for guessing the place type.  I was thinking of recording the users choice against the country and ADMx values so that the user did not have to make changes every time.  The recorded data would be maintained across sessions.
5) I hope to support some language functionality, so that displayed results appear in the users language, where it is available.

A future goal is to also integrate GOV in a similar way.  As you note, GOV data includes better place type data, as well as dated names, types etc.  But GOV doesn't have much of a web api, I have not found one to do searches, for example.  So I'd have to use a standard URL and then scrape the returned web page for data (GOV IDs) and place names.  Not a great long term solution since they might change web page layout or URL at any time, rendering that part of the addon useless.

By allowing the user to select from search results at each step, I expect that most issues created by using two gazetteers would be left up to the user.  More work for the user, but if the search works well, in most cases the user would only have to click a few times to complete a place.

GaryGriffin:
It sounds like your add-on would completely subsume this add-on. So there would be no value in releasing this.

Please let me know if you want anything from me. 

Couple last thoughts
1) A common place type in GeoNames is Township. This is not in the PlaceType._DATAMAP. It might be useful to add.
2) What to do when the user does not want to use all of the hierarchy entities from a GeoNames search. For instance, many cases where Populated Place and parent ADMx have the same name. Or ADMm and ADMn have the same name (state capital, for instance).
3) GOV - I was not able to find an API for the search either.

sam-m888:
@GaryGriffin Should this be closed?

GaryGriffin:
Yes, this should be closed. Paul extracted what he needed for a much better add-on.

PR #166, opened 2018-09-23
[NoSurnameCheckTool]Helps people with issue #10078[gramps50]

sam-m888:
Helps people who have issue [10078](https://gramps-project.org/bugs/view.php?id=10078)


sam-m888:
@prculley Fixed locally and never pushed to github!

PR #172, opened 2018-10-30
PhotoTaggingGramplet fixes and updates

vantu5z:
- fixes #10860 #10866
- cleanup
- update haarcascade file
- add option to configure sensitivity
- add 'face detect' button icon gramps-face-detection
- add popup menu item 'set person as active'
- open popup menu on treeview row right click

Before:

After:
after


vantu5z:
Add icon for "face detect" button:
gramps-face-detection
First try to load '`gramps-face-detect`' icon from theme, else use included to addon.

prculley:
Tested, looks like it works ok on at least some of my photos.  Something odd when right clicking in the list at right; if I just right click, it appears to reassign the person or something.  I have to hold down the right mouse button to select one of the options.

I rebuilt and listed before noticing this, so please take a look and see if that is correct, (I don't think it is!).

vantu5z:
Do you have this issue only at right list or on regions click too?
And which gtk version do you use?

vantu5z:
@prculley can you try if this https://github.com/gramps-project/addons-source/commit/8cf76ffdc25600566787f3bf931ee1380f19d84f fix issue with popup?

PR #173, opened 2018-11-02
Update selectform.py

romjerome:
Tweak improvement for handling the translation on "form_type" string.

As most event types are already loaded from gramps translation domain, 
we could use existing translations for displaying the localized 'form_type' string.

This minor translation issue is visible when we select a form from the list : the model ignores
the top level string (event type as parent). e.g., "census" instead of "recensement" under the french locale.


romjerome:
select_form_fr

romjerome:
select_form_fr_patch

PR #174, opened 2018-11-21
DynamicWeb Swedish translation

pehlm:
I've also a new translation in Swedish for DynamicWeb report, which I think is 100%. It required some changes in dynamicweb.py-file to be complete. So I create a pull request also for this file. I often use DynamicWeb it's a great tool.


prculley:
Rebuilt and listed, thank you for your translation.

PR #175, opened 2018-11-24
PhotoTaggingGramplet fix right-click menu

vantu5z:
fixes #10866

### DIFF
--- a/PhotoTaggingGramplet/PhotoTaggingGramplet.py
+++ b/PhotoTaggingGramplet/PhotoTaggingGramplet.py
@@ -674,7 +674,7 @@ class PhotoTaggingGramplet(Gramplet):
         """
         self.prepare_context_menu()
         self.context_menu.show_all()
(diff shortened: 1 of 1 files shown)


prculley:
This appears to correct the right-click doesn't pop up menu but does something weirdness, but an extremely minor oddity is that unlike most right-click menus, it doesn't pop up at once, but only after letting go the right click.

Tested on the Windows AIO env with Gtk 3.22.30 and with 3.18.9.

Did you figure out what it was doing with the 'button-press-event' that made it work that way?  Perhaps multiple events arriving at the same time?  If you are comfortable with this the way it is, I can merge/rebuild/list it, just say the word.

vantu5z:
Figure out what it was doing after showing menu, "cursor_changed" was called.
Stop signal emission should be enough to have common behavior of right click menu.

@prculley can you test it?
Strange but for me it is not differ (with or without stopping signal emission), menu is shown and doesn't hide.

prculley:
I like the way it works now.  Thanks for the update.  Rebuilt and listed.

PR #176, opened 2018-12-07
[Plugin manager] add filter

vantu5z:
Looks like this:


prculley:
Thanks; since I already dealt with this, we don't have to merge.

prculley:
I also want to thank you for your work on translating all of the Gramps Addons and Gramps itself.

PR #177, opened 2018-12-09
GenealogyTree: Add new report for sandclock of family[gramps51]

hahnjo:
This is modelled after DescendTree which also uses a single class to
reuse as much code as possible. Tested with real data on Gramps 5.0
and basic tests on Gramps 5.1 / master.

### DIFF
--- a/GenealogyTree/gt_sandclock.py
+++ b/GenealogyTree/gt_sandclock.py
@@ -45,7 +45,7 @@ LOG = logging.getLogger(".Tree")
 from gramps.gen.errors import ReportError
 from gramps.gen.plug.report import Report
 from gramps.gen.plug.report import MenuReportOptions
-from gramps.gen.plug.menu import PersonOption, NumberOption, BooleanOption
+from gramps.gen.plug.menu import PersonOption, FamilyOption, NumberOption, BooleanOption
 from gramps.gen.const import GRAMPS_LOCALE as glocale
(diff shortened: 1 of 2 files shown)


Nick-Hall:
Thanks for your contribution.  It is nice to see that people are finding this new report type useful.

It is acceptable to add new third-party addons to the _gramps50_ branch, if you want to make them available to users immediately.

Nick-Hall:
In core Gramps, the correct procedure would be to add new features to the _master_ branch.  However, we allow greater flexibility in the third-party addon reporsitory.

hahnjo:
I've also opened #178, please let me know how to proceed.

Nick-Hall:
There is nothing more you need to do for now.  A member of the _Addon Publishers_ group will review the PR and publish the report for you.

hahnjo:
Is it ok to just close this and you pull in the changes from `gramps50`? There more anyway after #184 was merged.

prculley:
We have been intermittently merging changes from the gramps50 branch into master on both the main Gramps repo and addons-source.  So you don't have to duplicate work for both branches.
We are NOT doing this for gramps42, so if you make changes in that branch that also apply to gramps50, cherry-pick to another branch/PR for both.

hahnjo:
Okay, I'm not planning to cherry-pick these for `gramps42`.

PR #178, opened 2018-12-09
GenealogyTree: Add new report for sandclock of family[gramps50]

hahnjo:
This is modelled after DescendTree which also uses a single class to
reuse as much code as possible. Tested with real data on Gramps 5.0
and basic tests on Gramps 5.1 / master.

### DIFF
--- a/GenealogyTree/gt_sandclock.py
+++ b/GenealogyTree/gt_sandclock.py
@@ -45,7 +45,7 @@ LOG = logging.getLogger(".Tree")
 from gramps.gen.errors import ReportError
 from gramps.gen.plug.report import Report
 from gramps.gen.plug.report import MenuReportOptions
-from gramps.gen.plug.menu import PersonOption, NumberOption, BooleanOption
+from gramps.gen.plug.menu import PersonOption, FamilyOption, NumberOption, BooleanOption
 from gramps.gen.const import GRAMPS_LOCALE as glocale
 _ = glocale.translation.gettext
 
@@ -54,6 +54,10 @@ _ = glocale.translation.gettext
 # SandclockTree
 #
 #------------------------------------------------------------------------
+
+_PERSON_RPT_NAME = 'gt_sandclock'
+_FAMILY_RPT_NAME = 'gt_sandclock_family'
+
 class SandclockTree(Report):
     """ Sandclock Tree report """
 
@@ -67,6 +71,7 @@ class SandclockTree(Report):
         get_option_by_name = menu.get_option_by_name
         get_value = lambda name: get_option_by_name(name).get_value()
 
+        self._person_report = options.name.split(",")[0] == _PERSON_RPT_NAME
         self._db = self.database
 
         self._pid = get_value('pid')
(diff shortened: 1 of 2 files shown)


sam-m888:
Thanks for the new report.

Do I have to use any special settings to have the report display correctly when output to pdf as if prints one blank page and you can partiality see the report at the bottom of page 2 (see screenshots below) ?

sandclock-of-family-report-example-output-defaults-50
sandclock-of-family-report-options-default-50
menu-sandclock-of-family-50

hahnjo:
> Do I have to use any special settings to have the report display correctly when output to pdf as if prints one blank page and you can partiality see the report at the bottom of page 2 (see screenshots below) ? Edit: don't worry this looks to be related to https://gramps-project.org/bugs/view.php?id=10512 )

Right, the page is probably just too small and LaTeX does its layout-thing. You could try larger dimensions like `a3paper`...

SNoiraud:
I have the same problem : I only see my tree partially. The page size is inoperative for all page format.
I tried to solves this adding the template directive after the genealogytree:
\genealogytree[
template=database pole reduced,
...

I can now see all my tree. The problem is we can't resize more than 400% 

To test it, i have the following patch:
diff --git a/gramps/gen/plug/docgen/treedoc.py b/gramps/gen/plug/docgen/treedoc.py
index 4918090ae..074edc52f 100644
--- a/gramps/gen/plug/docgen/treedoc.py
+++ b/gramps/gen/plug/docgen/treedoc.py
@@ -359,6 +359,7 @@ class TreeDocBase(BaseDoc, TreeDoc):
 
     def start_tree(self, option_list):
         self.write(0, '\\genealogytree[\n')
+        self.write(0, 'template=database pole reduced,\n')
         self.write(0, 'processing=database,\n')
         if self.marriage:
             info = self.detail + ' ' + self.marriage

Nick-Hall:
> \genealogytree[
> template=database pole reduced,
> ...


Alternatively, if the option is only required for this report, it can be added to the [options](https://github.com/gramps-project/addons-source/blob/fca95f5798492fa06f236d2a74f3394270a07482/GenealogyTree/gt_sandclock.py#L98) list.

hahnjo:
I'm not really see how these problems are related to this PR: The new report might generate larger trees (because it considers roughly twice as many nodes going upwards), but I guess you could also get cropped graphs with the existing sandclock or even other genealogytree reports...

Nick-Hall:
@hahnjo Yes, I agree.  A separate PR should be created for the page size issues.

sam-m888:
Thank you

hahnjo:
> @hahnjo Yes, I agree. A separate PR should be created for the page size issues.

Another route (that I'm taking) is to only use the graph database from the sandclock report and have your own LaTeX file around that, customized to your needs. I'll probably try to add a new option to only export the .graph file, so that users don't need to manually copy the database :-)

sam-m888:
Addon has been rebuilt with your new report.  Please update your addons.[1]

Thank you

[1]
https://gramps-project.org/wiki/index.php/5.0_Addons#Installing_Addons_in_Gramps

>

hahnjo:
> Addon has been rebuilt with your new report.

Thanks! I have some more fixes and enhancements that I'll submit after the holidays.

PR #179, opened 2018-12-21
[PrerequistesCheckerGramplet]Bump Gramps version[gramps50]

sam-m888:
Update check for Gramps 5.0.1 release.

### DIFF
--- a/PrerequisitesCheckerGramplet/PrerequisitesCheckerGramplet.py
+++ b/PrerequisitesCheckerGramplet/PrerequisitesCheckerGramplet.py
@@ -242,8 +242,8 @@ class PrerequisitesCheckerGramplet(Gramplet):
         '''
         self.append_text("\n")
         # Start check
-        LATEST_GRAMPS_VERSION = (5, 0, 0)
-        LATEST_GRAMPS_DATE = "2018-07-24"
+        LATEST_GRAMPS_VERSION = (5, 0, 1)
+        LATEST_GRAMPS_DATE = "2018-12-20"
         latest_release_message = ("Gramps " + verstr(LATEST_GRAMPS_VERSION) +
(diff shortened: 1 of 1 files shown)


sam-m888:
@prculley Thanks for the new Gramps 5.0.1 release. Regarding the wonderful new make.py command **as-needed** (Thank you by the way 👍 🥇 ) and running _git status_ afterwards gives the following result below. My question should the **po/template.pot** files be deleted when I haven't touched those addons at all, aren't they used for translations? Instead of me continuing with the addon rebuild can I have you confirm that I should be deleting those file please?

> 	deleted:    AssociationsTool/po/template.pot
> 	deleted:    FamilyTree/po/template.pot
> 	deleted:    Form/po/template.pot
> 	modified:   GenealogyTree/treeplugins.gpr.py
> 	deleted:    LastChange/po/template.pot
> 	deleted:    MediaVerify/po/template.pot
> 	deleted:    NoteCleanup/po/template.pot
> 	deleted:    PlaceCompletion/po/template.pot
> 	modified:   PrerequisitesCheckerGramplet/PrerequisitesCheckerGramplet.gpr.py
> 	deleted:    RebuildTypes/po/template.pot
> 	deleted:    RelID/po/template.pot
> 	deleted:    SurnameMappingGramplet/po/template.pot
>

prculley:
I added the make 'clean' equivalent code to the 'as-needed' which results in the .pot files getting deleted.  I have to admit that I was not sure about that, so I have been doing a "git checkout -- *.pot" after running the script to save the pot files.

I theory, anytime someone updates a po, he should be making a new pot file anyway, just in case, but who knows...
If you think we should always preserve pot files, I'll fix the script to not delete them.

PR #180, opened 2018-12-23
Set correct font resolution for Mac when drawing text directly to cairo.

jralls:
Fixes [#0010924](https://gramps-project.org/bugs/view.php?id=10924)

Will cherry-pick to master if accepted.


sam-m888:
Thank you

PR #183, opened 2019-01-02
[DynamicWeb]Update OpenStreemaps[gramps50]

sam-m888:
Fixes [#10954](https://gramps-project.org/bugs/view.php?id=10954)


sam-m888:
Addon rebuilt as
https://github.com/gramps-project/addons/commit/bace31366e57ef0763c6c458ce4361e5dee18810

PR #184, opened 2019-01-03
Minor enhancements and fixes for GenealogyTree[gramps50]

hahnjo:
* Fix for `max_up` if there is more data than the user export.
 * Add option `include_siblings` which can be disabled.
 * Extend range of generations to allow 0 generations in sandclock.


sam-m888:
Works for me.

New Option for "[ ] Include siblings" on Report tab.

pr184-new-option-include-siblings-gt-sandclock
pr184-new-option-include-siblings-gt-sandclock-family

sam-m888:
Thank you

PR #185, opened 2019-01-09
[Make.py]Lets keep the *.pot files[gramps50]

prculley:
After some comments from translators that not having pot files makes their work more difficult, I think it may make sense to keep the pot files with the sources.

I've update the make.py to be able to do a "./make.py gramps50 init all" which will remake all the pot files.  We may want to do this every once in a while to make sure that the pot files are up to date with the source code.

I'm also considering modifying the make.py as-needed option to also make the pot file for any updated source files.  Let me know if you think that makes sense.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,6 @@ tags
 
 # Translations
 *.mo
-*.pot
 
 # Distutils
 AllNamesQuickview/locale/
--- a/AllNamesQuickview/po/template.pot
+++ b/AllNamesQuickview/po/template.pot
@@ -1,0 +1,43 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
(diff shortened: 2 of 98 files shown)


sam-m888:
All for anything that helps improve the ease of Translation.

Makes sense as I've often thought that the **init** option also need an "all" parameter 👍 

> I'm also considering modifying the make.py as-needed option to also make the pot file for any updated source files. Let me know if you think that makes sense.

Yes it does 👍 

Can you also update the README.md to mention [subcomand: all] for init  and add an example etc..

(Tested by copying your make.py to a rewound git repo for addon-sources and no obvious errors in my quick check, will check again in a few days.)

sam-m888:
Tested make.py with the default and then went to try with the **all** command and didn't work, checked this pr and looks like the updated make.py wasn't added?

prculley:
Oops, looks like I mistakenly added it to a prior commit push, already in repo.

sam-m888:
Same oops with me looked at my command history I typed *init* incorrectly
the second time!

I suggest after updating the read.me this be merged and the other change to
the make be submitted as a separate PR.

PR #187, opened 2019-01-22
Update make.py as-needed to keep .pot files up to date

prculley:
This updates make.py as-needed function to do a recreation of the *.pot file for any change in a source file found during the as-needed scan.  The recreated .pot file does keep any previous manual header changes.


sam-m888:
Thank you

PR #192, opened 2019-03-08
Croatian translation for addons[gramps51]

milotype:
Some updates, but mostly new croatian translations for all addons.

### DIFF
--- a/FRWebConnectPack/po/hr-local.po
+++ b/FRWebConnectPack/po/hr-local.po
@@ -1,0 +1,55 @@
+# Croatian translation for Gramps
+# This file is distributed under the same license as the Gramps package.
+# Copyright (C) Gramps project
+# Milo Ivir <mail@milotype.de>, 2019.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Gramps 5.x\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2019-01-22 09:28-0600\n"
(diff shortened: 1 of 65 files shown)


sam-m888:
@milotype Are you sure you want this on the Gramps master branch and not the current release gramps50?

milotype:
Well, ideally I guess on both, but whatever you find more usefull. As I do my commits one-by-one over GitHub directly, it's a lot of work for both branches. Can you guys duplicate it for gramps50 (if at all necessary)?
Should I in general rather update the highest "maintenance" branch?

sam-m888:
> Should I in general rather update the highest "maintenance" branch?

Yes the **maintenance/gramps50**  branch is periodically merged to master (sorry not the other way around :( )

milotype:
Can you push/pull (or whatever needed) this "commit-package" to the **maintenance/gramps50 branch** instead of to master?

prculley:
I have manually shifted these commits to gramps50, rebuilt etc.  And then merged up into gramps51 and rebuilt there as well.  So your translations should appear in all the addons for each (5.0.x, 5.1.x) version of Gramps.

milotype:
@prculley Thank you! ... I found one small error in addon "NoteCleanup" (true for both 5.0.x and 5.1.x branches). The file `hr-local.po` is not placed in the po-subfolder.
Current location:
`addons-source/NoteCleanup/hr-local.po`
Should be:
`addons-source/NoteCleanup/po/hr-local.po`

I've checked all other addon folders and they are fine.

PR #193, opened 2019-03-19
DynamicWeb: openlayers use "values_" instead of "p"

SNoiraud:
Fixes [#11057](https://gramps-project.org/bugs/view.php?id=11057)

Some changes in the openlayers library


sam-m888:
Thank you

PR #195, opened 2019-03-25
[GraphView] Add search capability[gramps50]

vantu5z:
Снимок экрана от 2019-03-25 11-32-40

### DIFF
--- a/GraphView/graphview.py
+++ b/GraphView/graphview.py
@@ -679,13 +679,15 @@ class GraphWidget(object):
         self.goto_other_btn.set_tooltip_text(
             _('Center view on selected bookmark'))
         hbox.pack_start(self.goto_other_btn, False, False, 1)
-        self.bkmark_popover, self.bkmark_box = self.build_bkmark_popover()
+        self.bkmark_popover, self.bkmark_box, \
+            self.bkmark_box_other = self.build_bkmark_popover()
         self.goto_other_btn.connect("clicked", self.show_bkmark_popup)
         self.show_images_option = self.view._config.get(
             'interface.graphview-search-show-images')
 
         # add search widget
-        self.search_box = SearchWidget(self, self.activate_search, self.dbstate)
+        self.search_box = SearchWidget(self, self.activate_search,
+                                       self.dbstate)
         hbox.pack_start(self.search_box, True, True, 1)
         self.search_box.set_options(
             search_all_db=self.view._config.get(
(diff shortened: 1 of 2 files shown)


vantu5z:
Снимок экрана от 2019-03-27 10-10-18

vantu5z:
Снимок экрана от 2019-03-28 11-25-59

vantu5z:
Have some problem, I use methods to set max size of scroll windows that came from Gtk 3.22:
```
ScrolledWindow.set_max_content_height(200)
ScrolledWindow.set_propagate_natural_height(True)
```
And for now can't find some solution or workaround.

vantu5z:
@SNoiraud can you test it? Maybe you have some other ideas to include?

SNoiraud:
Tested. it's OK for me. I don't see something new to add.

SNoiraud:
Perhaps. Do you mean to search someone in the database then bookmark him?

vantu5z:
No. Now we show bookmarks only for graph and other bookmarks are hidden. I suggest add this bookmarks to second scroll window like in the search popup.

SNoiraud:
OK.

SNoiraud:
It could be interesting to bookmark a person from the menu (right button: edit, edit tags...)

vantu5z:
Yes, I think about this too.
But think it should be separate PR like: right click popup enhancement
Related issue about tags https://gramps-project.org/bugs/view.php?id=10870

SNoiraud:
It could be interesting to bookmark a person from the menu (right button: edit, edit tags...)

PQYPLZXHGF:
> [Have some problem, I use methods to set max size of scroll windows that came from Gtk 3.22:](https://github.com/gramps-project/addons-source/pull/195#issuecomment-477688428)

@vantu5z You probably want to take the advice [here](https://github.com/gramps-project/gramps/pull/783#issuecomment-481321833) by @Nick-Hall that only properties that work on Gtk v3.12 be used for Gramps v5.1 and for the current Gramps 5.0 or older only use Gtk v3.10 properties.

sam-m888:
Please ensure that it works with gtk 3.10 as mentioned. Thank you

vantu5z:
Workaround was added for gtk 3.10, but I can't test it on gtk 3.10. If some one make test it will be nice.

adamvazquez82:
I really like the search box, maybe somehow it can be integrated into the toolbar so you can use it all over Gramps, what do you think? Or should I raise a feature request?

55165883-ef8efa80-514c-11e9-91eb-05dfe7070b81

vantu5z:
Should results popup be closed on click some item?

vantu5z:
rebased

vantu5z:
rebased after PR #199 revert

vantu5z:
Seems that it requires Gtk 3.12 because of Gtk.Popover.
So it should go to Gramps 5.1 ([link](https://github.com/gramps-project/gramps/pull/783#issuecomment-481321833))

PR #196, opened 2019-04-03
[GraphView]does not show all descendants[gramps50]

SNoiraud:
Fixes [#11088](https://gramps-project.org/bugs/view.php?id=11088)


sam-m888:
Thank you

PR #197, opened 2019-04-20
[Interactive Family Tree]New Addon (Topola)[gramps51]

PeWu:
### Instructions
Open the addon from the menu: Tools -> Analysis and Exploration -> Interactive Family Tree
image

A browser tab will be opened with an interactive tree view. See [Topola Genealogy Viewer](https://pewu.github.io/topola-viewer) for examples how the visualization works.

### How it works
This Gramps addon starts a HTTP server that serves the GEDCOM file that is loaded by the [Topola Genealogy Viewer](https://pewu.github.io/topola-viewer) interface. The user data is not sent to any remote location and only loaded into the browser memory.

### DIFF
--- a/Topola/Topola.gpr.py
+++ b/Topola/Topola.gpr.py
@@ -1,0 +1,15 @@
+register(TOOL,
+         id = 'Topola',
+         name =_('Interactive Family Tree'),
+         description = _('Opens an interactive tree in the browser'),
+         status = STABLE,
+         version = '1.0.0',
+         fname = 'topola.py',
+         gramps_target_version = '5.1',
+         authors = ['Przemek Więch'],
+         authors_email = ['pwiech@gmail.com'],
+         category = TOOL_ANAL,
+         toolclass = 'Topola',
+         optionclass = 'TopolaOptions',
+         tool_modes = [TOOL_MODE_GUI]
+)
--- a/Topola/po/pl-local.po
+++ b/Topola/po/pl-local.po
@@ -1,0 +1,27 @@
+# Polish translations for PACKAGE package.
+# Copyright (C) 2019 THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
+# Przemek Wiech <pwiech@gmail.com>, 2019.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: gramps51\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2019-04-16 20:32+0200\n"
+"PO-Revision-Date: 2019-04-16 20:32+0200\n"
+"Last-Translator: Przemek Więch <pwiech@gmail.com>\n"
+"Language-Team: Polish\n"
+"Language: pl\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=3; plural=(n==1 ? 0 : n%10>=2 && n%10<=4 && (n%100<10 "
+"|| n%100>=20) ? 1 : 2);\n"
+
+#: Topola/Topola.gpr.py:3
+msgid "Interactive Family Tree"
+msgstr "Interaktywne drzewo genealogiczne"
+
+#: Topola/Topola.gpr.py:4
+msgid "Opens an interactive tree in the browser"
+msgstr "Otwiera interaktywne drzewo genealogiczne w przeglądarce"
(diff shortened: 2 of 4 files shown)


sam-m888:
Hi Przemek Więch,

Thank you for the new addon

Can you please add a description and usage instructions here please if
possible? And maybe some screenshots?


Kind regards
Sam

PeWu:
Hi Sam,

Thanks for the response. I updated the PR description.

wroldwiedbwe:
@PeWu  Only get the following with my family tree , do I have to do something else?

Error Topola Genealogy Viewer

> The user data is not sent to any remote location and only loaded into the browser memory.

Browser url shows:

> https://pewu.github.io/topola-viewer/#/view?utm_source=gramps&handleCors=false&standalone=false&url=http://127.0.0.1:8156/

Which seems to indicate that https://pewu.github.io/topola-viewer/#/view  is going to receive the GEDCOM but ?utm_source=gramps&handleCors=false&standalone=false&url=http://127.0.0.1:8156/ is the source ( see [utm_source](https://en.wikipedia.org/wiki/UTM_parameters) )of the GEDCOM , so this actually if it works will send your GEDCOM to the Topola website, the opposite of what you say, can you make this work truly in offline mode!

PeWu:
@wroldwiedbwe The addon **does not** send your GEDCOM anywhere. This was one of the core ideas behind making this application that is different than most online family tree building sites.
You don't have to trust my word. You can:
1. Look at the source code of the Topola Viewer site here: https://github.com/PeWu/topola-viewer
2. Look at the network traffic in the browser developer tools to see what happens – the GEDCOM is retrieved from http://127.0.0.1:8156/ (Gramps) and is never sent anywhere.

The reason it is done like this is to always run the most recent version of the viewer just like any other web application. The application is loaded from the web but is run on your local computer.
utm_souce=gramps is only there for statistical purposes and doesn't mean any of your data is sent anywhere.

I don't believe you really need offline mode. However, if you really care, after you open https://pewu.github.io/topola-viewer/ once in your browser, it should be stored in the browser cache and be available even when you are offline.

Going back to the "Failed to load file" message. I don't know what is happening in your case. My guess is either that the database is empty (the app doesn't handle this case gracefully), or that the GEDCOM file generated by Gramps is not being read correctly. You can look at the GEDCOM file by starting the addon and then going to http://127.0.0.1:8156/

sam-m888:
Great addon 👍 

Good looking charts.

<hr>
Some notes (screenshots) from testing using Example.gramps https://gramps-project.org/wiki/index.php?title=Example.gramps

* Opens a browser window and shows the family of the first person in this case person id **I0000 The First Person**
Topola Genealogy Viewer - example gramps the first person

* You have an option of two views (1) Hourglass Chart & (2) All relatives
Topola Genealogy Viewer -chart types

* To see other people use the search box at top.
Topola Genealogy Viewer - searching

<hr>
Possible issue:

* Attempting to show my personal family tree gave the following screen like @wroldwiedbwe 
2019-06-10 17_47_14-Topola Genealogy Viewer failed to load file

Believe it has to do with the size of the family tree in my case 14 thousand people , so it may have been a timing issue while the GEDCOM was being output to http://127.0.0.1:8156/  

@wroldwiedbwe Was your family tree large ? If so how many people?
<hr>
Suggestions (for later, as I'll be accepting this PR in a few days):

(1) Is it possible to filter the GEDCOM output (eg call the GEDCOM export GUI and set the filer etc then export to Topola) as Topola seems to work great for smaller family trees?

(2) Have an option on which person id to show initially?

<hr>
@wroldwiedbwe  @PeWu  The offline method worked for me and I believe nothing is sent to an external source ( @Nick-Hall @prculley Can you verify, I've looked at the code and nothing seems to indicate that.)

PeWu:
Thanks @sam-m888 for trying the addon. I'm not convinced it is the data size that is the problem. I tested topola with my family tree with ~4k entries and it runs smoothly. Do you know if there is a large data set somewhere I could test against? Or a data generator?
Can you check if the GEDCOM file output to http://127.0.0.1:8156/ correctly?

sam-m888:
> I'm not convinced it is the data size that is the problem. I tested topola with my family tree with ~4k entries and it runs smoothly. 

Ok, keep in mind mine was family tree 14k entries.

> Do you know if there is a large data set somewhere I could test against? Or a data generator?

You could possibly try [GEDFAN](https://www.tamurajones.net/DownloadGedFan.xhtml) ? Or one of the large GEDCOM files mentioned [here](https://genealogy.stackexchange.com/questions/8004/big-and-public-family-tree-as-gedcom) or [here](https://www.tamurajones.net/TwoHugeGEDCOMFiles.xhtml)

> Can you check if the GEDCOM file output to http://127.0.0.1:8156/ correctly?

Took a while but yes.

sam-m888:
Thank you.

wroldwiedbwe:
> Was your family tree large ? If so how many people?

@sam-m888 23K people

nicorikken:
@PeWu Thanks for creating Topola and integrating it in Gramps. I can see myself using this to showcase the family tree to relatives. It is a nice way of exploring the tree.

PR #207, opened 2019-06-04
[AvatarGenerator]Add new addon[gramps50]

Mattkmmr:
The new addon **AvatarGenerator** allows users to add and remove one or serveral images to a group of filtered people. Users can choose the images from their database. The selected avatar images are only added to people without an image. More information [here](https://gramps-project.org/wiki/index.php/Addon:AvatarGenerator).

alt text

Before & After:
Avatar_Generator


sam-m888:
- [x]  Updated support page with before & after screenshot

sam-m888:
Thank you

PR #208, opened 2019-06-11
[GraphView]Add options for Vertical/horizontal spacing between nodes[gramps51]

vantu5z:
fixes [11155](https://gramps-project.org/bugs/view.php?id=11155)
Снимок экрана от 2019-06-11 11-18-41
Снимок экрана от 2019-06-11 11-18-26

### DIFF
--- a/GraphView/graphview.py
+++ b/GraphView/graphview.py
@@ -820,83 +820,33 @@ class GraphWidget(object):
 
         # add spinners for quick generations change
         gen_box = Gtk.Box(orientation=Gtk.Orientation.VERTICAL)
-        box = Gtk.Box(orientation=Gtk.Orientation.HORIZONTAL)
-        img = Gtk.Image.new_from_icon_name('go-up-symbolic',
-                                           Gtk.IconSize.MENU)
-        box.pack_start(img, False, False, 1)
-        self.ancestors_spinner = Gtk.SpinButton.new_with_range(0, 50, 1)
-        self.ancestors_spinner.set_tooltip_text(_('Ancestor generations'))
-        self.ancestors_spinner.set_value(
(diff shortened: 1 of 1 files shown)


fxtmtrsine:
Wouldn't that option be better as a button on the tool bar, instead of hidden?

vantu5z:
Like generation spinners? Yes, put it on the toolbar will be better.

Mattkmmr:
Vertical spacing works great, but maybe the family nodes should scale too, so the lines of children rich families aren't packed too tight together. ⬇️

fxtmtrsine:
> Like generation spinners? Yes, put it on the toolbar will be better.

@vantu5z Great 👍

vantu5z:
rebased for Gramps51

vantu5z:
While testing, found that with graphviz 2.42.1 we have problem with scale, it generates cropped SVG.
Have to downgrade to 2.40.1 version.

sam-m888:
Looks good 👍 

Spelling correction
* beetwen  > between

Please change the tool tips to respectively say:

*  "Horizontal spacing between generations"
and 
* "Vertical spacing between generations"

2019-09-19 10_03_56-Window

sam-m888:
The popovers are much more compact 👍 

generations to show

chart spacings

GraphView/graphview.py
```diff
@@ -899,6 +924,16 @@ def __init__(self, view, dbstate, uistate):
         # used for popup menu, prevent destroy menu as local variable
         self.menu = None
 
+    def spinners_popup(self, widget, popover):
+        """
+        Popover for generations and spacing params.
+        Different popup depending on gtk version.
```
sam-m888: 👍

sam-m888:
Thank you

PR #213, opened 2019-07-04
[BirthdaysGramplet] Add 'Only show birthdays with tag' functionality[gramps51]

Mattkmmr:
As requested by @wroldwiedbwe an opposite option that only shows birthdays of persons tagged with a certain tag:
`All birthdays are shown when neither the 'ignore tag' nor the 'only tag' is selected. The 'ignore tag' hides all birthdays of people tagged with it and the 'only tag' hides all birthdays of person NOT tagged with it. If a person is tagged with both tags (ignore and only) the birthday will be ignored. The tag names don't have to be 'ignore' or 'only'. All tag names are valid.`


wroldwiedbwe:
@Mattkmmr That was quick, thanks 💯 👍 

only show living birthsdays with tag

sam-m888:
Thank you

PR #214, opened 2019-07-09
GEPS045 update addons for Enhanced Places

prculley:
This PR is intended to go along with https://github.com/gramps-project/gramps/pull/809 and MUST NOT be commited until that PR is accepted and commited.

I've provided it at the moment to allow testing of some of these addons with the main PR, in particular GetGOV and PlaceCleanup, which have been modified to support the primary PR and fully utilize the main PRs features.


romjerome:
> MUST NOT be commited until that PR is accepted and commited

Sorry I did some commits on lxml gramplet related to GEPS 045 but they are minor and could be easily reverted. Only one **real** change related to master branch (an exception with empty source title).

Also have a [strange issue around dummyDB on etree gramplet](https://github.com/gramps-project/addons-source/commit/3e94b7036793e5b7e406e5772953c11396c93715), getting an error by calling '_self.dbstate.db.get_number_of_citations()_' which seems specific to master branch as this worked before.

prculley:
@romjerome It appears you made your changes in master branch.  Perhaps you can rebase your work on top of mine and push to this branch?  Probably not too important, since I suspect you are the only one using these Gramplets...

Regarding the dummyDB; since this db is only active when the main dbs are closed, and dummyDB doesn't implement any of the 'get_number_of_???' methods, I'm surprised you only saw the issue with citations.  I would have expected 'NotImplementedError' for all the calls.  You should probably put all the print statements inside of a "if self.dbstate.db.db_is_open:" like you did for the tags.

romjerome:
> I would have expected 'NotImplementedError' for all the calls.

Yes, it is logical to get the 'NotImplementedError' according to current code,
but that should not be specific to citation objects on this gramplet.

My first impression was that something has been improved around environment for gedcom support, which might (could?) explain that the method for citation has this custom behaviour?
e.g., most objects could use a new way and citation object has not be completely updated?
I need to look at piece of code using this counting method. Also need to look at a proper use 
of '_db.surname_list_' which is also broken on my experimental etree gramplet.

>  I suspect you are the only one using these Gramplets...

Yes, maybe !
They could be hidden or removed.

As they provided some informations without advanced coding or large investigations for people discovering gramps, maybe someone could be happy to look at them for its own use.

Thanks.

romjerome:
> # FutureWarning:
> # The behavior of this method will change in future versions.
> # Use specific 'len(elem)' or 'elem is not None' test instead.

oh, yes I also need to look at my sample (gramps xml file) as 
this old comment has been added some years ago ...

Often using len() as the simplest method for counting
but also some ugly part of code for returing collected data
on this etree gramplet (lazy way). Maybe a re-factoring
could be easier that looking at some specific issues
with this current code!

romjerome:
Note, one (feature) on etree gramplet lets help me to check structure and content of my data after an upgrade of gramps version : counting references (links or xlinks) not primary object or tags, often secondary object but also some hidden data like attributes or specific data.

It was a quick and dirty check based on counters.
```
others = len(entries) - (len(tags) + len(events) + len(people) + len(families) + len(sources) + 
len(citations) + len(places) + len(objects) + len(repositories) + len(notes)
```
So, it is not a proper and advanced DB check, only an other test before a migration.

romjerome:
@prculley 
> 
> Regarding the dummyDB; since this db is only active when the main dbs are closed, and dummyDB doesn't implement any of the 'get_number_of_???' methods, I'm surprised you only saw the issue with citations. 

I tried to generated a [pull request](https://github.com/gramps-project/gramps/pull/857) via github (online) but it was a little bit complicated for me (gramps50, gramps51, master branches), so it has been [closed](https://github.com/gramps-project/gramps/pull/857)!

Anyway, as far as I can understand ... get_number_of_citations method is just missing on dummydb module and it returns an error with my etree gramplet related to logging environment. So, it is not visible
and maybe not very important, but I can propose a possible patch:
```diff
index 89c93ab..3f39fbc 100644
--- a/gramps/gen/db/dummydb.py
+++ b/gramps/gen/db/dummydb.py
@@ -757,6 +757,14 @@ class DummyDb(M_A_M_B("NewBaseClass", (DbReadBase, Callback, object,), {})):
             LOG.warning("database is closed")
         return 0
 
+    def get_number_of_citations(self):
+        """
+        Return the number of citations currently in the database.
+        """
+        if not self.db_is_open:
+            LOG.warning("database is closed")
+        return 0
+
     def get_number_of_tags(self):
         """
         Return the number of tags currently in the database.
```
```diff
index 731fcd7..c6f3139 100644
--- a/gramps/plugins/db/bsddb/test/db_test.py
+++ b/gramps/plugins/db/bsddb/test/db_test.py
@@ -81,6 +81,7 @@ class DbTest(unittest.TestCase):
         "get_number_of_places",
         "get_number_of_repositories",
         "get_number_of_sources",
+        "get_number_of_citations",
         "get_number_of_tags",
         "get_media_from_gramps_id",
         "get_media_from_handle",
```

I can guess why this 'citation' method is not useful with gedcom (import, export) and only saw an alternate way (_diff between 'source' and 'citation' objects on merge, where source dos not call get_number_of_sources and citation did._). So, currently it seems that the method is just missing on dummydb and that's why I got the cryptic issue.

PR #215, opened 2019-07-11
[Dynamicweb]Fix crash when running in cli mode[gramps51]

SNoiraud:
Fixes [#11213](https://gramps-project.org/bugs/view.php?id=11213)


prculley:
I don't think this is the right way to fix this; the issue affects all reports which allow filters.  So use https://github.com/gramps-project/gramps/pull/894 instead.

PR #223, opened 2019-09-15
Hebrew translation for GetGov & Phototagging Gramplets

avma:


sam-m888:
Do you really want this in Gramps master and not the current gramps51 branch?

avma:
Isn't master the source for the rest, (gramps50, gramps51..)?

prculley:
For the Gramps main code, during early development, new code is put into master.  So if the translator is keeping up to date, all changes can happen there.  If the translator doesn't get his translations in before release, it is best to make the changes into the most current released maintenance branch.

For addons it is a bit different.  If the addon was released directly into a maintenance branch (very common) then translation changes should be made there.

For both Gramps main code and addons, we periodically merge all changes from the most recent maintenance branch into master, and sometimes from a lower branch into an upper one.

So if you are updating an addon and you think that you want your updates to appear for a particular version of Gramps, make the changes to that associated branch.  Either merge your changes to higher branches yourself (extra PRs), or make a note that you hope that we will do this for you, particularly if you make your changes to an older maintenance branch than the most recent.

avma:
Thank you prculley for clarifying this for me.
So I guess this should go into maintenance/gramps51...
Should I create new pr's and close this one, or can you move them across?

prculley:
I rebased onto gramps51, then rebuilt the addons and committed/pushed everything so the new translations should be available.

avma:
Thanks Paul....

PR #229, opened 2019-09-18
[Graph View]Display invalid dates when show_full_dates==True[gramps51]

ailintom:
Currently Graph View only displays dates which have at least a valid year. This behavior differs from that of other charts (Pedigree, H-Tree). The correct behavior is displaying invalid dates (`MOD_TEXTONLY`) as well when the date is not empty and `show_full_dates` is set to `True`. 

Also the addon currently erroneously displays the birth/death place when `self.show_places==False` and date is empty. Correct behavior: only display place when  `self.show_places==True`

### DIFF
--- a/GraphView/graphview.py
+++ b/GraphView/graphview.py
@@ -2487,7 +2487,7 @@ class DotSvgGenerator(object):
                         rtrn += ' - %s' % place_title
                 return rtrn
             else:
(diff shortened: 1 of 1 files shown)


sam-m888:
Thank you

@vantu5z  Thank you for reviewing this 😊

PR #201, opened 2019-05-29
[GraphView] right click menu updates [Gramps50] recreation

vantu5z:
Recreation of #199.

PR was reverted since [issue 11161](https://gramps-project.org/bugs/view.php?id=11161#bugnotes).

Now it should work. [fix](https://github.com/gramps-project/addons-source/commit/dfab772359626fdf70ea3910cbef9774e34bdbba)
I have forgot about this [link](https://github.com/gramps-project/addons-source/pull/120#issuecomment-370108363).

Anyway we should test it before merge.

PR #232, opened 2019-09-23
[GraphView] Add drag and drop function

vantu5z:
It works, but have some issue with drag icon.


Mattkmmr:
@vantu5z:  Drag&drop does not work for me. Dragging works, symbol is shown, but the dropping to the clipboard fails.

vantu5z:
Matt, can you provide more info?
OS, program versions, exception with it fails.

Mattkmmr:
Windows 10
Gramps 5.1.1
GTK+ 3.18.9

There aren't any exception raised. What do you mean with the text fields?

vantu5z:
You can drop to text entry fields (where you can type) and other apps like notepad.

vantu5z:
Matt, can you test this:
In function "cb_drag_data_get" change "id(self)" to "id(self.get_widget())"

sam-m888:
Needs to be tested on each of the OS's

- [ ] Linux
- [ ] Apple MacOS Package
- [ ] Microsoft Windows AIO Installer

sam-m888:
> In function "cb_drag_data_get" change "id(self)" to "id(self.get_widget())" or "id(widget)"

Neither work on Windows AIO 64bit or 32 bit Gramps 5.1.1 on Win 10 (1903) 64bit

PQYPLZXHGF:
Works but with a big catch, the drag works as per normal but the for the drop you need to press the "Enter" Key (, which nobody would normally do with drag & drop) or else the drag icon stays on the screen unless you press Escape to cancel

test


On Windows 10 Home 64bit /  Gramps 5.1.1

vantu5z:
@PQYPLZXHGF can you test this on Windows:
in function `motion_notify_event` in call `drag_widget.drag_begin_with_coordinates` change parameter `event` to `None`

vantu5z:
Now is should work in Windows.
I setup VirtualBox for testing. But it should be tested on real machines.
Mouse button param in `drag_begin_with_coordinates` should be 1 (as left mouse)

sam-m888:
> Now it should work in Windows.

Confirming new changes work on Windows.

@jralls Do you have time to test on MacOS please?

jralls:
It sort of works on MacOS. If I control-touch I get a page icon for a drag symbol. Releasing the control key adds a + in a green circle and I can then drag the page icon over to the clipboard window and when I lift my finger from the touchpad it adds the chart's focus person, not the person I dragged. The click-and-drag gesture doesn't work.

Attempting to drop on the name field of the filter gramplet does nothing.

If I don't release the control key the page icon does a fly-back when I lift my finger from the touchpad.

I used the Gramps-5.1.1-1 app bundle with  Gtk 3.14.15, copying graphview.py over the latest released one. Unfortunately more recent versions of Gtk are unstable (bugs [9718](https://gramps-project.org/bugs/view.php?id=9718) and [10434](https://gramps-project.org/bugs/view.php?id=10434)) with Gramps on MacOS though they work fine with C applications.

That said, drag and drop work as expected on the regular chart view: click and drag to the clipboard window drops the dragged person with no extra calisthenics required.

vantu5z:
Seems this is touchpad issue.
Do you make tests with mouse?

sam-m888:
>>  [my old laptop that has ubuntu 14.04 and gtk 3.10.8,](https://github.com/gramps-project/addons-source/pull/201#issuecomment-500655231)

@wroldwiedbwe If you still have this laptop, can you test this PR works for you please?

vantu5z:
rebased

wroldwiedbwe:
@sam-m888 I only have my old laptop that has ubuntu 14.04 and gtk **3.10.8.**  , but yes it seems to work well 👍

prculley:
Been a while, is this still a work in progress?

vantu5z:
We have some problems with MacOS and I don't know how to fix this. On Linux and Windows should work fine.

vantu5z:
@jralls, I change drag-n-drop activation. Please, can you test if it works for you on Mac now?

vantu5z:
Example, drop person from clipboard to family node:
изображение

jralls:
@vantu5z, tested on 5.1.4.

Well, click-and-drag now works, and the first drag after startup works as expected except that it changes the focus person to the dragged person. Dropping on another person in the chart brings up an "add relationship" dialog and dropping on the clipboard window adds the person to the clipboard. So far so good. Dropping on the name field in the filter gramplet changes the chart as if I'd double-clicked on the person, which isn't what I'd expect.

With the selected person changed by the first drag subsequent drag attempts don't change the selected person and drop the person from the first drag unless I first explicitly change the selected person first by clicking on someone else.

To illustrate, suppose I have a chart with Alice, Barbara, Charles, David, and Edward. Charles is the selected person when I switch to Graph view. I click and drag Alice to the clipboard window; Alice becomes the selected person. I click and drag David to the clipboard and get Alice again. I double click on Barbara to make her the selected person, then try again to drag David to the clipboard. That works, and now David is the selected person. I try to drag Charles and get another instance of David unless I click on someone (could be David) first.

vantu5z:
@jralls, it seems you wait alot after click and before move mouse to enable drag.
So when you click (left mouse down) at person node - this activates timer (200 ms) to separate single/double clicks, and if drag is not activated before timer is triggerred then handle single click on person node and make it active person (rebuild graph).
Try to fix this behavior, now drag will be disabled if active person changes (by single click).
Try if this works for you?

jralls:
@vantu5z That works better. I see that you came to your senses after your comment and changed click recognition to button-up instead of a (very short in human time) timeout on the button down.

vantu5z:
Add drag support from search widget. Now we can find some one and drag it to graph or clipboard.

PR #233, opened 2019-09-24
[Graphview]Fix tag colors on boxes[gramps51]

romjerome:


romjerome:
fix [10721](https://gramps-project.org/bugs/view.php?id=10721)

romjerome:
#86

GraphView/graphview.py
```diff
@@ -2377,6 +2377,12 @@ def get_tags_and_table(self, obj):
             tag_table = ('<TABLE BORDER="0" CELLBORDER="0" '
                          'CELLPADDING="5"><TR>')
             for tag in tags:
+                rgba = Gdk.RGBA()
+                rgba.parse(tag.get_color())
+                value = '#%02x%02x%02x' % (int(rgba.red * 255),
+                                   int(rgba.green * 255),
+                                   int(rgba.blue * 255))
+                tag.set_color(value)
```
vantu5z: Should we really need to set color to tag?
`tag.set_color(value)`
Maybe just use generated value?
`tag_table += '<TD BGCOLOR="%s"></TD>' % value`

romjerome: I do not know ... Does it work too?
In my mind, it was for re-assigning
a value after the conversion, 
like done by the hand via GUI.

Yes, feel free. Thanks.

romjerome: Note, as there is no DB commit,
I only supposed that [setting color](https://www.gramps-project.org/docs/gen/gen_lib.html?highlight=tag#gramps.gen.lib.tag.Tag.set_co) 
was in memory.

vantu5z: Ok for me as we don't commiting to db.

sam-m888:
Mentioned on bug reported as already having been mentioned here

https://github.com/gramps-project/addons-source/pull/86#issuecomment-327999074

sam-m888:
@romjerome Can you please add a screenshot of before and after , so I can work out what I should be looking ?

romjerome:
@sam-m888 I can make it better ...
I will use your screenshot! 
img_20170908_130447727

See your [comment](https://github.com/gramps-project/addons-source/pull/86#issuecomment-327987719) on pull request #86

sam-m888:
Thank you that helps, so the issue is that the Chart was displaying the wrong colors compared to the tooltip & "Organise tags" dialog 👍   Tested and confirmed.

Before:
2019-09-29 11_57_28-Window

After: 

2019-09-29 11_59_07-Window

2019-09-29 11_58_53-Window

sam-m888:
Thank you

PR #237, opened 2019-09-29
[Graphview] Label position and font [v51]

vantu5z:
Next attempt to fix [8822](https://gramps-project.org/bugs/view.php?id=8822)

Don't use `font_family_map` and just use `font-family` from SVG.

Screenshots from Windows 
Befor:
befor
After:
after

### DIFF
--- a/GraphView/graphview.py
+++ b/GraphView/graphview.py
@@ -1501,14 +1501,6 @@ class GraphvizSvgParser(object):
         # list of persons items, used for animation class
         self.items_list = []
 
-        # This dictionary maps various specific fonts to their generic font
-        # types. Will need to include any truetype fonts here.
-        self.font_family_map = {"Times New Roman,serif": "Times",
-                                "Times Roman,serif":     "Times",
-                                "Times-Roman":           "Times",
-                                "Times,serif":           "Times",
-                                "Arial":                 "Helvetica",}
-
         self.transform_scale = 1
         self.font_changed()
 
@@ -1769,14 +1761,10 @@ class GraphvizSvgParser(object):
         # does the following always work with symbols?
         if style:
             p_style = self.parse_style(style)
-            try:
-                font_family = self.font_family_map[p_style['font-family']]
-            except KeyError:
-                font_family = p_style['font-family']
+            font_family = p_style['font-family']
             text_font = font_family + " " + p_style['font-size'] + 'px'
         else:
-            font_family = self.font_family_map[
-                self.text_attrs.get('font-family')]
+            font_family = self.text_attrs.get('font-family')
             font_size = self.text_attrs.get('font-size')
             text_font = font_family + " " + font_size + 'px'
 


vantu5z:
@jralls Can you test on MacOS please?

jralls:
I don't see any difference:
<figure><figcaption>Current gramps51</figcaption>Before_Patch</figure>

<figure><figcaption>With #237 applied</figcaption>After_Patch</figure>

It looks OK both ways, but `font-family` is obviously the correct CSS tag.

vantu5z:
This is good!

sandysson:
Thanks to the dev wizards!

sam-m888:
Thank you

prculley:
image
I note that on my system, for some reason, the font selected is a Sans.  And that the box size/vs font size is not right.  Does anyone know why/where the font was set for the display?

prculley:
And here is the Chrome browser display of the SVG using the 'Edit/Print' command...
image

jralls:
@prculley How did you get it to export an SVG? Or did you export the  `.gv` and then run dot on it to make an SVG?

Anyway, the font map hard-coded into `graphview.py` is
```
        # This dictionary maps various specific fonts to their generic font
        # types. Will need to include any truetype fonts here.
        self.font_family_map = {"Times New Roman,serif": "Times",
                                "Times Roman,serif":     "Times",
                                "Times-Roman":           "Times",
                                "Times,serif":           "Times",
                                "Arial":                 "Helvetica",}
```
So I'd guess that your Gramps fontconfig can't find a Times font and is falling back to Arial. AFAICT there's no way to change the font selections short of editing `graphview.py`.

vantu5z:
Strange, now we don't set font for graphviz (use default) and parse it from generated svg (without using font map).
Maybe issue with dpi?
You can try to set font and dpi in "init_dot" method and check if this helps.

vantu5z:
I want to add option to choose font and size, when have some time.

prculley:
@jralls If you use "Edit/Print..." while looking at the Graphview, it will spit out both the .gv and the .svg files.
I looked into this a bit.  It appears that dot is supposed to be using fontconfig (if it is enabled during dot's build) to find the font and do the substitutions necessary to deal with missing characters etc.  Presumably, the supposed length of the text strings are available by summing up the size of characters' glyphs.  And that would be used to determine the size of the boxes used.

Our svg to widget process uses the various GooCanvas/Pango/fontconfig?? elements to take the utf8 strings from the svg output and draw them.

The problem occurs if the two mechanisms end up with different ideas on the font to use, or the substitution characters etc.

It is quite clear that on my system, with the default font when nothing is specified is getting resolved to one of the sans type fonts, the calculated lengths of the string are significantly different by the GooCanvas/Pango and what dot found.  And despite the fact that the svg clearly has "Times-Roman" for the font-family.

For fun I asked dot to generate an svg with the "cairo" option, this results in an svg file where each individual glyph is drawn out in detail.  It is not usable by the Graphview svg parser, but when I looked at it with the Chrome browser, all the text fitted the boxes perfectly.  So dot can get it correct internally.

I suspect that if dot and Gtk are built with different font backends, (as in your OSx) this might not be an issue, at least as long as they match.

I'm thinking about putting some sort of adaptive code into place, something that runs a test string in a box through the dot/svg/GooCanvas/Pango process, checks for the fit of the text, and uses that as a scale factor for other text.  We might still have to adjust on the fly for text strings that got a lot of character font substitutions...

vantu5z:
Add font choose option #240

SNoiraud:
Graphview should use the font selected in the preferences (symbols).

prculley:
I now understand why the text doesn't always fit.  When generating svg, GraphViz uses a text length **estimator**, that doesn't take into account the actual font, nor Bold (which makes the characters a bit larger).  See https://gitlab.com/graphviz/graphviz/blob/master/lib/common/textspan.c to see what I mean.  I think it uses the actual rendered size when using cairo or another renderer to make other graphical file types.

So I don't think we will ever get a perfect fit when using SVG.  And it is likely to be worse if we use fonts chosen by the user.

The best we could do would be to accept the box size, and try to fit the font to the box (when it exceeded the size at least) by scaling the GooCanvasText width to fit in the box.  And limit the font to either Times or Arial/Helvetica (dont' think we want Courier) so the match is a good as possible.

Any thoughts before I try to generate a patch?

jralls:
I'm pretty sure that neither Times nor Arial have 100% Unicode coverage, so while forcing those two will work great for us users of western European languages it will cause trouble for everyone else.

Summing the size of the glyphs gives the correct layout only for fixed-width faces, which is why Pango exists. The right way to get the size of the box is to give Pango the font info and the whole string and ask it for for the layout size and position, then draw the box around that. That's how Gtk works internally. 

If I wanted to generate SVG from inside a Gtk program I'd use a Cairo SVG surface, draw to it like any other Cairo surface, and save the result. I'm not familiar with the details of using Graphviz but it sure looks to me that dot doesn't provide enough control for us to be able to do that.

prculley:
I used my developer status to add my own code to do a better fitting of the text to the dialog boxes.  Took a bunch of work to figure out a way that seems pretty reliable.
It is now live on Gramps 5.1.x versions after addon check/upgrade.  It seems to work fairly well with the fonts on my system, let me know if you see any issues.

PR #134, opened 2018-04-24
Geps/gep 039 genealogical symbols[gramps51]

SNoiraud:
Trying to use symbols for views and report.

Issue [#9098](https://gramps-project.org/bugs/view.php?id=9098)

- Symbols for QuiltView and TimelinePedigreeView

PR #240, opened 2019-10-09
[GraphView] choose font option [v51]

vantu5z:
Now we can choose font. Only font_name and size (not style).

In Windows7 it looks gook for me with `Arial 16`:
win

### DIFF
--- a/GraphView/graphview.py
+++ b/GraphView/graphview.py
@@ -640,7 +640,7 @@ class GraphView(NavigationView):
         font_lbl = Gtk.Label(_('Font:'), xalign=0)
         grid.attach(font_lbl, 1, row, 1, 1)
         font = self._config.get('interface.graphview-font')
-        font_str = '%s %d' % (font[0], font[1])
+        font_str = '%s, %d' % (font[0], font[1])
         font_btn = Gtk.FontButton.new_with_font(font_str)
         font_btn.set_show_style(False)
         grid.attach(font_btn, 2, row, 1, 1)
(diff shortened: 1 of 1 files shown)


SNoiraud:
I don't understand that.
When I added the Symbols, my PR #134  added the correct unicode font with all symbols.
With or without this patch, this doesn't work anymore.
Now when we change the font in the preferences, the font is not change as it should in graphview.

vantu5z:
@SNoiraud, I don't use symbols preferences, because i can't install Python-fontconfig on my Arch Linux. So I can't choose font this way.

GraphView/graphview.py
```diff
@@ -707,6 +708,15 @@ def search_config_panel(self, configdialog):
...
         return _('Search'), grid
+
+    def font_filter_func(self, family, face):
+        style = face.get_face_name()
+        if style in ('Обычный', 'Regular', 'Normal', 'obyčejné', 'Standard',
+                     'Κανονικά', 'Normaali', 'Normál', 'Normale', 'Standaard',
+                      'Normalny', 'Normálne', 'Navadno', 'thường', 'Arrunta'):
+            return True
+        return False
+
```
prculley: I'm not sure you should do it this way; we support a lot more languages than this I think.  I suggest:
<pre>
        desc = face.describe()
        stretch = desc.get_stretch()
        if stretch != Pango.Stretch.NORMAL:
            return False  # avoid Condensed or Expanded
        sty = desc.get_style()
        if sty != Pango.Style.NORMAL:
            return False  # avoid italic etc.
        weight = desc.get_weight()
        if weight != Pango.Weight.NORMAL:
            return False  # avoid Bold
        return True
</pre>

vantu5z: Yes, your solution is better. Thanks!

prculley:
I merged this and added my own code to do a better fit of the font to the boxes.  It is now live on Gramps 5.1.x versions after addon check/upgrade.  It seems to work fairly well with the fonts on my system, let me know if you see any issues.

sam-m888:
- [x] Support page updated

PR #251, opened 2019-10-23
[assosiationtool] hebrew translation [gramps51]

avma:


prculley:
It would be a bit easier for us if you would try to keep your commits in the appropriate branch.  This is the third time I have seen the PlaceCleanup translation in a PR, and I need to check each time to make sure that it is not something new, but one that is already in the repository.

I suspect that when you created the branch for assosiationstools that you were referencing the PlaceCleanup branch, or maybe had that commit in your base gramps51 branch.

For future work, I think you need to
git checkout maintenance/gramps51
git reset --hard HEAD~1       # removes the PlaceCleanup commit
git pull           # update to latest, assuming you have it set up to pull from upstream gramps51

then you can create a new clean branch.

If you use "gitk --all" you can see a semi-graphical view of how the branches and commits are connected.

If you get into something like this in the future, I often use:
git rebase -i commithash
where commithash is the correct base commit or branch name.  This allows you to remove inappropriate commits, merge some together, rename etc.

We do appreciate that git is not easy to get right, and we also appreciate the translation work.  Please keep it up.

I will accept this as it is this time.

prculley:
Added and rebuilt...

PR #254, opened 2019-10-25
[MediaReport] Add image size options [gramps51]

Mattkmmr:
Feature request [#11348](https://gramps-project.org/bugs/view.php?id=11348)

### DIFF
--- a/MediaReport/media_report.gpr.py
+++ b/MediaReport/media_report.gpr.py
@@ -23,7 +23,7 @@ register(REPORT,
          name = _("Media Report"),
          description = _("Generates report including images, image data"
                          " and notes."),
-         version = '1.0.2',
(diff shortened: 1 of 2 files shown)


sam-m888:
mediareport-updated 11348

- [ ] Update support [page](https://gramps-project.org/wiki/index.php/Addon:MediaReport)

lordemannd:
doesn't allowing height and width to be adjusted independently add additional distortion to the image? would be a single "Scale image" percentage be possible, forcing width and height to be scaled by the same amount?

Mattkmmr:
@lordemannd No, distortion to the image isn't the goal of those two seperate values. The width/height options are only the limit for the max. image size. The image is already scaled by the same amount until it reaches the first limit.

PR #256, opened 2019-11-01
[NLWebConnectPack]New plugin[v52]

coret:
New plugin NLWebConnectPack based on similar ConnectionPack addons forr other languages, but now with Dutch resources.


sam-m888:
> gramps_target_version = "5.1",

Did you mean to commit this to the gramps51 branch instead of master (future gramp 5.2.x) which requires that this line be:

> gramps_target_version = "5.2",

PR #257, opened 2019-11-10
Add new addon [EventDescriptionEditor] [gramps51]

Mattkmmr:
The Event Description Editor tool allows users to select several events, search their event descriptions for a substring and change the description or the substring to a new one. This feature was requested in [#4757](https://gramps-project.org/bugs/view.php?id=4757) and in [#2656](https://gramps-project.org/bugs/view.php?id=2656).

### DIFF
--- a/EventDescriptionEditor/EventDescriptionEditor.gpr.py
+++ b/EventDescriptionEditor/EventDescriptionEditor.gpr.py
@@ -1,0 +1,36 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2019  Matthias Kemmer
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


prculley:
Looked this over and it appears to work.

But I think you need to add the "parent" parameter for the OkDialog; on my multi-monitor system it did not appear over Gramps, but centered on a different monitor.

Some stylistic comments;
You may also want to look into the Python string replace function, it would considerably simplify some of your code and make it a bit more readable.  If you used that, you could replace the 'count' function with "if sub_str in description:" type tests.

Also, I think pylint would complain about the use of single character variables.

Mattkmmr:
[Wiki documentation](https://gramps-project.org/wiki/index.php/Addon:EventDescriptionEditor)

prculley:
I see this still has a Work in progress flag???  Planning more fixes?

Mattkmmr:
I updated some code, but I do not plan more features/fixes.

prculley:
Then you should turn off 'work in progress' so we know it may be eligible for merging.

Mattkmmr:
@prculley I don't think that I can add or remove labels. @sam-m888 added them originally.

prculley:
I didn't realize that labels were limited to the privileged few; @sam-m888 can we change this?

Nick-Hall:
I have sent an invitation.

PR #261, opened 2019-12-01
[Topola] Added support for exposing files referenced from the GEDCOM file [gramps51]

PeWu:
This allows displaying photos next to people in the Topola Viewer.


prculley:
I will preface my comment with the fact that I know very little about Python programming and threads.  So I don't really know if your program will have an issue.  However, the way Gramps is written it is NOT designed for any kind of simultaneous access to the database.  So if the fact that you are using threads means that you might be trying to create the GEDCOM file in one thread, and other parts of Gramps are actively running, allowing the user to modify the db, in the main thread, you are likely to get corrupted GEDCOM data or a crash in the GEDCOM creation.

If you were to create the GEDCOM file as part of the tool startup, but merely serve that file without touching Gramps db or other data in a thread, I think you would be all right.

I do note that the set_database code may be handling this, but I don't know enough to tell if this is the case.

Otherwise, if it works, cool...

PeWu:
Thanks Paul for the response.

You noticed correctly that I added a background thread that reads the database. If Gramps doesn't handle concurrent database operations then this may be a problem.

I haven't found any sources saying anything about thread-safety of the code apart from this: https://www.gramps-project.org/wiki/index.php/GEPS_042:_Multi-user_record_locking_and_editing

Since it is not certain whether this threading code is safe, I'll remove threading from this PR.

PeWu:
I removed the code that generates the GEDCOM file in a background thread. What I wanted to do with this code is to avoid freezing Gramps for the duration of this task. In another PR I'll try adding a modal window with a progress bar to avoid freezing.

PeWu:
@prculley @sam-m888 I'd like to merge this PR. Can you help me?

prculley:
Done and rebuilt/listed.

Paul C.

PeWu:
Thanks!

PR #271, opened 2020-01-03
Fix NetworkChart to work with v2.x of networkx library

prculley:
Fixes #11471

Recent Linux systems have started including networkx version 2.x, rather than the older v1.11.  v2.x introduced some incompatibilities, but one in particular was causing the bug.  If/when networkx reaches v3.x, we can expect more such issues.

This Pr makes a minor change to allow either version of networkx to work.

### DIFF
--- a/NetworkChart/NetworkChart.py
+++ b/NetworkChart/NetworkChart.py
@@ -474,6 +474,10 @@ class NetworkChartReport(Report):
         G = nx.DiGraph()
         G.clear()
 
+        # deal with incompatibiliy between v1 and v2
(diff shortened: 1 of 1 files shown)


prculley:
In a private email Jan says:

> Great job! Your modifications work like a charm, I have tested it by creating svg's and pdf's of large and small network chart reports.
> No error occurred, even with creating a svg-report of 17,5MB...
> 
> Your solution is ready for production, I think!
> Thank you very much for your help!!!
> 
> Best regards,
> Jan Sparreboom

PR #272, opened 2020-01-04
New filter rule 'personshavingeventfiltermatch' [gramps51]

Mattkmmr:
Matches people who are matched by an event filter.


Mattkmmr:
Not as new for persons as I thought initially. (Persons matching an event filter does already exist)

PR #274, opened 2020-01-04
New filter rule 'familieswitheventfiltermatch'

Mattkmmr:
Matches families that are matched by an event filter. Requested in [#11499](https://gramps-project.org/bugs/view.php?id=11499)

### DIFF
--- a/FilterRules/familieswitheventfiltermatch.gpr.py
+++ b/FilterRules/familieswitheventfiltermatch.gpr.py
@@ -1,0 +1,34 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2020       Matthias Kemmer
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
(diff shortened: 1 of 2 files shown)


prculley:
I looked this over and it seems a bit more complex than necessary.  I don't understand why you scan through person looking for families; why not just scan families?  Furthermore, I think you could have just inherited from MatchesEventFilterBase and updated the name etc. without any prepare/apply code.

Mattkmmr:
The feature request was: 
> I'm trying to come up with a way of filtering out *all* people that have a connection to a place *with contained places*

Therefore this filter rule scans through all persons and their events (e.g. births, deaths, etc) and checks if they match with the event filter. The families containing persons with their events matching the event filter are returned by this filter rule. (I don't think I can do that only with MatchesEventFilterBase. I guess it's only checking the family events.)

Any ideas for improvements/simplifications? I'm going to rewrite the filter rule description to make it less confusing.

prculley:
I think I understand what you want, but I am a bit concerned that making complex filters that are difficult to describe in a one-liner will be hard to understand and use.  I wonder if we should try for two or more simpler filters to be used in the custom editor.  Maybe something like
0) events with places matching place filter
1) persons with events matching event filter (0)
2) custom filter with either
  a) families with events matching event filter (0)
  b) families containing persons (with options for parent and child) matching person filter (1)

I think that both a and b would be new filters...

The idea is that simpler filters can be combined in many other ways.  It is clear that we don't have all of the simpler filters we need for a truly general filter set as of yet.

Pleas take these as suggestions, and keep up the good work.

Mattkmmr:
I agree, simpler filters would be better. I'll split them up.

I think currently the main problem for a general filter set or better filtering in Gramps is that we need a way to select the first filter result and then use the following filters on that without having to create a new new filter each time. Right now the creation of several complex filters in Gramps always ends up in a mess.

Mattkmmr:
I changed this filter to "Families matching <event filter>" (family events) instead of "Families of persons matching an <event filter>" (person events).

PR #275, opened 2020-01-04
New filter rule 'personshavingfamilyfiltermatch'

Mattkmmr:
Matches people who are matched by a family filter. Requested in [#11499](https://gramps-project.org/bugs/view.php?id=11499)


prculley:
Looks ok to me; I assume you tested to make sure it works?

prculley:
If we are going to add a filter like this, why don't we make it a bit more general in that we have option checkbox for children and parents; which would allow it to serve the function of "persons who are children of family filter match" and/or "persons who are parents in family filter match"

FilterRules/personshavingfamilyfiltermatch.py
```diff
@@ -0,0 +1,83 @@
...
+                    father = family.get_father_handle()
+                    mother = family.get_mother_handle()
+                    children = family.get_child_ref_list()
+                    self.persons.add(father)
+                    self.persons.add(mother)
+                    for child in children:
+                        child_ref = child.get_referenced_handles()
+                        if child_ref[0] == 'Person':
+                            self.persons.add(child_ref[1])
+
```
prculley: 'child' is actually a ChildRef object, so it would be a bit easier to use:
`self.persons.add(child.ref)`
rather than looking through referenced handles to find this.

Mattkmmr:
Hmm how should no checkbox selection being handled?  Is there a way to tell users that they have to select at least one checkbox for the filter to work?

Alternatively we could also just split it into three filters.
a) all family members of family filter match
b) parents of family filter match
c) children of family filter match

PR #277, opened 2020-01-04
PhotoTagging addon: added Italian localization

giansalvo:


prculley:
Could you rebase this to maintenance/gramps51?  At the moment we are doing all the addon work there and periodically merge up to master.  This is because addons-source master is not used at the moment; and won't be until we update Gramps master to gramps52 status.

giansalvo:
I opened https://github.com/gramps-project/addons-source/pull/281 rebased to gramps51 and closing this one.

PR #279, opened 2020-01-04
Fix Forms Gramplet to allow forms to be retrieved when language changes

prculley:
Fixes #11504

The Forms Gramplet identifies forms by looking at the custom 'Form" attribute on the Source.  Since this is a custom attribute, it never gets translated when storing in the db/xml.  So when you open Gramps in a different language, the comparison to the localized "Form" fails.

The right way to fix this is to add "Form" as a standard AttributeType, so it would be localized in all supported languages.  But that won't happen quickly.  In the meantime, I have made the comparison to all the current languages that Forms supports.  It looks like it works for me.

### DIFF
--- a/Form/form.py
+++ b/Form/form.py
@@ -61,6 +61,8 @@ ORDER_ATTR = _('Order')
 
 # The key of the data item in a source to define it as a form source.
 DEFINITION_KEY = _('Form')
+# the following should be all the translations of "Form" in our po files
+INTL_FORM = {"Form", "Formular", "Formulaire", "Obrazac", "Форма"}
 
 # Prefixes for family attributes.
 GROOM = _('Groom')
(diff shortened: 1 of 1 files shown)


stevenyoungs:
@prculley, I believe this will fix one part of #11504, namely that the forms are not visible at all when changing gramps UI language.
#11504 also illustrates the problem of form content itself "vanishing" when switching language. The underlying problem is the same - the attribute name being translated.

If we consider the attributes created by the form addon as internal data of the addon (i.e. not intended for direct user editing), perhaps it is acceptable to always store the attribute in the database using the source name (English), but still translate for display in the GUI. For backwards compatibility, the addon would need to search under the attributes English name, and then translated name. TBD if the translated name is deleted when saving the form.

I tend to think that the 'Form' attribute itself, by virtue of having to be directly created by end-users, should remain in translated form. Your PR helps make this.

Form/form.py
```diff
@@ -61,6 +61,8 @@
 
 # The key of the data item in a source to define it as a form source.
 DEFINITION_KEY = _('Form')
+# the following should be all the translations of "Form" in our po files
+INTL_FORM = {"Form", "Formular", "Formulaire", "Obrazac", "Форма"}
```
stevenyoungs: In case a new langauge is created, can a comment be added to template.pot to remind the translator to also get this list updated?

PR #280, opened 2020-01-04
DescendantsLines addon: several improvements

giansalvo:
- added left and right margins
- renamed options to be more user friendly
- added same width, same height options
- added header and footer and font size coeff
- added some comments to the code
- rearranged options to reduce size on screen

### DIFF
--- a/DescendantsLines/DescendantsLines.gpr.py
+++ b/DescendantsLines/DescendantsLines.gpr.py
@@ -5,6 +5,7 @@
 # Copyright (C) 2010 Jerome Rapinat <romjerome@yahoo.fr>
 # Copyright (C) 2010, 2012 lcc <lcc.mailaddress@gmail.com>
 # Copyright (C) 2015 Don Piercy
+# Copyright (C) 2020 Giansalvo Gusinu <giansalvo.gusinu@gmail.com>
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


prculley:
Looks ok to me...

giansalvo:
@prculley Thanks for your feedback

> We try to make sure that dialogs/windows fit onto an 800x600 screen, I think this is getting too tall for that. Maybe we need another tab or two. Consider putting the three 'display format' boxes in their own tab, header/footer text near privacy text, and possibly in a tab of its own. And rearrange the report options and display to hold items more evenly in height.
> 
ok, I can do this

> I see you tried to add better tooltips and descriptive names to some of the options in the last two tabs, but could you try for better names for the tabs themselves? What is S&F or O&C? And you missed a couple MIN_C_WIDTH and OL_PAD.
> 
I didn't miss those parameters, I just can't understand what they do.
Please let me know what you want to put instead of S&F and O&C and I'll be happy to substitute them.

SNoiraud:
You have many trailing spaces:
DescendantsLines/DescendantsLines.py:255:        MIN_C_WIDTH - 
DescendantsLines/DescendantsLines.py:695:        
DescendantsLines/DescendantsLines.py:1608:        
DescendantsLines/DescendantsLines.py:1774:        
DescendantsLines/DescendantsLines.py:1792:        
ERROR - Trailing whitespace found in source file(s)

giansalvo:
@SNoiraud 
I removed the trailing spaces. I had a problem with my editor's settings. It should be fixed now.

prculley:
Because of minor file merge conflicts, I had to do this manually.  It is accepted and uploaded.

PR #282, opened 2020-01-04
DescendantBooks: 'Report appearances' section option

giansalvo:
- added Report appearances section option

### DIFF
--- a/DescendantBooks/DetailedDescendantBookReport.py
+++ b/DescendantBooks/DetailedDescendantBookReport.py
@@ -121,6 +121,7 @@ class DetailedDescendantBookReport(Report):
         incindexnames - Whether to include the index of names at the end of the report.
         incindexplaces- Whether to include the index of places at the end of the report.
         incindexdates - Whether to include the index of dates at the end of the report.
+        incappearances- Whether to include Include the 'Report appearances' section with each person.
         incssign      - Whether to include a sign ('+') before the descendant number in the child-list to indicate a child has succession.
         pid           - The Gramps ID of the center person for the report.
         name_format   - Preferred format to display names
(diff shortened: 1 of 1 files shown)


prculley:
I think the option name "Include only once" and help description doesn't provide a good description of what the option does.  Maybe change popup to something like:
"Don't include the 'Report appearances' section with each person (which shows other places in the report where the person is mentioned)."

Or even better, "Include other person appearances" with popup: "Include the 'Report appearances' section with each person (which shows other places in the report where the person is mentioned)."  and invert your logic and have it checked by default.

DescendantBooks/DetailedDescendantBookReport.py
```diff
@@ -442,11 +444,11 @@ def write_report(self):
             self.report_count += 1
             if len(self.ascendants) > 1:
                 self.title = \
-                    self._("%(report_count)s. Descendant Report for " \
+                    _("%(report_count)s. Descendant Report for " \
```
Nick-Hall: This will break the deferred translation functionality.

DescendantBooks/DetailedDescendantBookReport.py
```diff
@@ -531,7 +533,7 @@ def write_index_of_dates(self):
         This function prints the index of dates.
         """
         self.doc.start_paragraph("DDR-Title")
-        self.doc.write_text_citation("Index of Dates")
+        self.doc.write_text_citation(_("Index of Dates"))
```
Nick-Hall: This should use deferred translation and belongs in a separate commit.

DescendantBooks/DetailedDescendantBookReport.py
giansalvo: Do you mean that I must revert all "\_(" to "self.\_(" or only here?

Nick-Hall: Everywhere.  It's needed to run reports in different languages.

prculley:
Made the aforementioned changes, built and uploaded.

PR #283, opened 2020-01-16
Add new filter rule:  DegreesOfSeperation

Mattkmmr:
The filter rule allows to filter for relatives by their degrees of seperation to the selected person. The code uses a modified version of the IsLessThanNthGenerationAncestorOf filter rule for ancestor and descandants calculation. It is also possible to include the partners of the descandants.

### DIFF
--- a/FilterRules/degreesofseparation.py
+++ b/FilterRules/degreesofseparation.py
@@ -17,7 +17,7 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
-"""Filter rule that matches relatives by degrees of seperation."""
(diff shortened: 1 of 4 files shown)


FilterRules/degreesofseparation.py
```diff
@@ -0,0 +1,203 @@
...
+
+        get_partners = bool(int(self.list[2]))
+        if get_partners:
+            self.__get_partners()
+
+    def __get_root_handle(self):
+        """Get the handle of the starting person."""
+        pid = self.list[0]
+        person = self.db.get_person_from_gramps_id(pid)
+        root_handle = person.get_handle()
```
prculley, line 126: You need to make a test for no person here (None), in case the filter is in use with a different tree and root handle is not defined.  See https://gramps-project.org/bugs/view.php?id=12426

PR #284, opened 2020-01-19
PlaceTool gramplet

kkujansuu:
This is a gramplet that helps to manage multiple places at the same time. See README.


wroldwiedbwe:
Travis ci found a bunch of trailing white spaces see : https://travis-ci.org/gramps-project/addons-source/builds/639056699#L2423

wroldwiedbwe:
Found the https://github.com/kkujansuu/addons-source/blob/maintenance/gramps51/PlaceTool/README.md !!

Gramplet to change properties of multiple places at the same time. The properties that can be changed are:

*  place type
*  tag
*  enclosing place

PlaceTool UI

Nick-Hall:
I think that this will be useful.  A few suggestions:

* Change the name to something like PlaceUpdate.  It is a gramplet not a tool.
* Use bold for headings only, not for field labels.
* Use a standard place selector for the enclosing place to make it consistent with other parts of Gramps.

wroldwiedbwe:
Thanks, Travis ci found one more trailing white space see : https://travis-ci.org/gramps-project/addons-source/builds/639216629#L2403

PR #285, opened 2020-01-25
MediaReport: new image DPI option

Mattkmmr:
A possible solution to fix issue [#11492](https://gramps-project.org/bugs/view.php?id=11492). It requires [#1007](https://github.com/gramps-project/gramps/pull/1007). A more detailed description can be found there.

### DIFF
--- a/MediaReport/media_report.gpr.py
+++ b/MediaReport/media_report.gpr.py
@@ -23,7 +23,7 @@ register(REPORT,
          name = _("Media Report"),
          description = _("Generates report including images, image data"
                          " and notes."),
-         version = '1.1.1',
+         version = '1.2.0',
          gramps_target_version = "5.1",
          status = STABLE,
          fname = "media_report.py",
--- a/MediaReport/media_report.py
+++ b/MediaReport/media_report.py
@@ -1,7 +1,7 @@
 #
 # Gramps - a GTK+/GNOME based genealogy program
 #
-# Copyright (C) 2019 Matthias Kemmer
+# Copyright (C)    2019 Matthias Kemmer
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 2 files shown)


lordemannd:
I tried to test this from CLI on Windows and got the following, what am I doing wrong?
Failed to write report. 
  File "C:\Program Files\GrampsAIO64-5.1.4-prelim\gramps\cli\plug\__init__.py", line 699, in cl_report
    my_report = report_class(database, clr.option_class, User())
  File "C:\Users\dave\AppData\Roaming\gramps\gramps51\plugins\MediaReport\media_report.py", line 64, in __init__
    self.window = self.user.uistate.window
AttributeError: 'NoneType' object has no attribute 'window'
Cleaning up.

Mattkmmr:
@lordemannd The report does not work for CLI in the current state, because it's using dialog windows which require a GUI.

prculley:
The report is coded as if can run in CLI mode (in the .gpr.py file), but the original author did not take that into account when he created the dialogs; he assumes that they will be run in GUI mode.  The dialogs should be using some user type agnostic code, that would work equally well in GUI or CLI mode.

lordemannd:
A little more testing:
CLI: MediaReport does respond to the show options command, so the CLI user can see how to enter options.
(not thoroughly tested)

Books: I added the Book report mode to media_report.gpr.py
         report_modes = [REPORT_MODE_CLI, REPORT_MODE_GUI, REPORT_MODE_BKI],

By this 1 line change, the book report seems working and was able to produce a pdf with MediaReport for 3 separate images.

prculley:
The dialogs in question occur after the initial options dialog; they occur during actual processing.  So yes, it will appear to work unless you actually run the report.

Mattkmmr:
@lordemannd I updated the report and it should no longer rely on the Gramps GUI window when in CLI. Please test if it is working now.

I have to fix more of my report and tool addons also using dialogs, since I only test in GUI mode and didn't thought enough about CLI related issues. I also didn't knew I forgot to add the `REPORT_MODE_BKI` to all of them, so people can also use them in books. :sweat_smile:

lordemannd:
@Mattkmmr, oops I guess I didn't see your change in Sep 2021, I will take a look now at your recent changes.

PR #288, opened 2020-02-29
[Topola] Fix addon under Windows

PeWu:
- remove unused ThreadPoolExecutor import
- force usage of UTF-8 instead of Windows default encoding


PeWu:
@prculley Hi, do you think you could merge this PR?
This change fixes the addon when running on Windows.

PeWu:
Thank you!

PR #292, opened 2020-03-07
GeoAncestor: trying to remove the dancing map.

SNoiraud:
+ add a progress meter popup window to display the progress of the map creation
+ create markers only at the end and not each person
+ solves the icon problem

Fixes #011594


lordemannd:
i see it working better/faster on Win10 and on Linux. 
(but I guess I wasn't seeing the original issue, So I can say is that it is better).

lordemannd:
i can see the new icon. on windows I never see the progress meter.

SNoiraud:
> on windows I never see the progress meter.
If you have a small tree, I think that the popup doesn't have time to appear.

lordemannd:
ok, just saw progress meter on windows. is the progress meter only for "Ancestor Map" or does it also apply to "Every residence or move for a person and ant descendants"? (I dont see it there)

SNoiraud:
For the moment, I only added it to the ancestor view.

lordemannd:
just a note when testing with pr1022 log file utf-8 encoding:
for person names with international characters, before this pr292 there was a line like "show one person : Родословная, Родословная" written into the log file, along with a warning like:
"C:\Program Files\GrampsAIO64-5.1.2-working\lib\python36.zip\gi\overrides\GObject.py:553: Warning: ../../glib-2.54.3/gobject/gsignal.c:2641: instance '000000000947c880' has no handler with id '52694'"
after using pr292 the "show one person" messages are gone, but the warning message is still logged.

PR #295, opened 2020-03-29
pt_PT Translations update

pmraps:
Translation updates


pmraps:
OK, will do. Thank you.

PR #306, opened 2020-04-09
New ActivePersonRule

prculley:
This rule allows the user to create reports with a filter that includes the active person.  The filter will generate a warning (when run with no active person or from CLI).


prculley:
In order for this filter to work correctly with the Person tree view, the https://github.com/gramps-project/gramps/pull/1039 must be accepted and in place.  So I don't think we should release this for general use until at least the next Gramps bug release.

prculley:
Works with 5.1.4 and newer...

PR #311, opened 2020-04-15
[GraphView] fix bug 0011658

vantu5z:
Escape illegal symbols in image path.


vantu5z:
Filenames should be in the GLib file name encoding.
[see in docs](https://lazka.github.io/pgi-docs/index.html#GdkPixbuf-2.0/classes/Pixbuf.html#GdkPixbuf.Pixbuf.new_from_file)

vantu5z:
Seems the problem is a Graphviz bug.

PR #312, opened 2020-04-15
Update WWII draft registration fields

Jmuccigr:
added missing fields & corrected a few others.

### DIFF
--- a/Form/form_us.xml
+++ b/Form/form_us.xml
@@ -7841,13 +7841,22 @@ Birth, Death, Marriage
                 <_attribute>Place of Birth</_attribute>
             </column>
             <column>
-                <_attribute>Who will always know your address (Name &amp; Address)</_attribute>
+                <_attribute>Country of Citizenship</_attribute>
             </column>
             <column>
-                <_attribute>Employer&apos;s Name &amp; address</_attribute>
+                <_attribute>Name of Person who will always know your address</_attribute>
             </column>
             <column>
-                <_attribute>Place of employment or business</_attribute>
+                <_attribute>Relationship of that Person</_attribute>
+            </column>
+            <column>
+                <_attribute>Address of that Person</_attribute>
+            </column>
+            <column>
+                <_attribute>Employer&apos;s Name</_attribute>
+            </column>
+            <column>
+                <_attribute>Place of employment or business (Address)</_attribute>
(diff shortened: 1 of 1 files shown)


Caulley:
This Form need a complete re-wright. When I created this form I was under the impression that there was only 1 card. There were 6 cards. The form need to be rewritten in a nondestructive manure.

Caulley:
A revised version of the form is at [](https://github.com/Caulley/Gramps-form_us/blob/main/World%20War%20II%20Draft%20Registration%20Card)/World%20War%20II%20Draft%20Registration%20Card. If someone/anyone/everyone could take a look at it and make suggestions to improve it would be helpful.

hgohel:
> A revised version of the form is at /World%20War%20II%20Draft%20Registration%20Card. If someone/anyone/everyone could take a look at it and make suggestions to improve it would be helpful.

@Caulley Could you provide a full link to the form? Thanks.

GaryGriffin:
I am confused by this change.

First, it is against master branch. It should be against gramps52.

Second, the differences in the PR includes one field _**Place of employment or business (Address)**_, so breaks existing forms. BUT this change is not consistent with gramps52 branch. gramps52 has the (Address) portion of the field name.

Third, when I use the form provided in the previous Comment, it appears to not break existing form entries. So that specific form change seems okay. 

What I see is some additional fields (citizenship, occupation, ...) . I don't have any real examples using these new fields, so I cannot confirm that the update matches all of the WW2 DRC forms. All of my examples use the simple form that is in the original Form layout.

I think this PR needs to be closed and reopened against gramps52 and use the form_us.xml which would include the revised WW2DRC portion. Then we can test it for merging.

Caulley:
Examples of the draft cards can be found here - https://drive.google.com/drive/folders/1h2h23yhgT6V1Do6vqcGMqr2FvcoA0KA7?usp=drive_link

GaryGriffin:
The intention is to CLOSE https://github.com/gramps-project/addons-source/pull/312, https://github.com/gramps-project/addons-source/pull/542, and https://github.com/gramps-project/addons-source/pull/546. [This](https://github.com/gramps-project/addons-source/pull/717) PR covers the Draft changes. The 1950 US Census changes will be discarded as they break current users.

PR #329, opened 2020-04-30
[CheckPlaceTitles]: Dutch translation

janpython:
Please, replace Dutch translation of addon CheckPlaceTitles.
It is NOT tested in Gramps 5.1.2 as this addon isn't installed and I can't find it...

Maybe this addon is not needed anymore, see: https://gramps-project.org/wiki/index.php/GEPS_045:_Place_Model_Enhancements


prculley:
The addon is not listed as it doesn't work properly.  Not sure if it really does anything useful anyway...
I've added your translation, just in case someone ever resurrects this.

PR #334, opened 2020-05-03
Place coordinate gramplet

CWSchulze:
New gramplet to add geographic location to places.

There are two parts.

1. part: Gramplet Dialog - This is used to add location information to existing places.
grafik
    - select a place
    - if necessary change the "search for" field
    - click "go"
       grafik
    - if the place was found you can apply the location by pressing "apply geo location to database"
    - To see what was found, you can show the location in a browser window by clicking "Show found place externally in Google Maps"

2. part: Geography View - This is used to add new places graphically.
grafik
    - In this view you can add new places in the context menu
      grafik
    - Either with building number (add address as place), or only the city name (add city as place)
    - The place name, coordinates and postal code are preset in the dialog
      grafik


adamvazquez82:
Testing this and wondering if the warning is significant?

When pressing  "Take last click from  Geography Map"  without going to Geography map first, so no coordinates or last click.

> ** (gramps.exe:9224): WARNING **: locale 'English_United Kingdom.1252' isn't valid

Open the Gramplet up and then press "Search Location from DB" button and get:


> 248118: ERROR: grampsapp.py: line 157: Unhandled exception
>   File "C:\Users\adamvazquez82\Documents\gramps\gramps51\plugins\PlaceCoordinateGramplet\PlaceCoordinateGramplet.py", line 223, in on_fromDBButton_clicked
>     latitude = float(latitude)
> ValueError: could not convert string to float: 
>

CWSchulze:
Yes, maybe my simple code to convert coordinates was implemented a bit too optimistically...
Is the format of your location in the database is using degrees/minutes/seconds? So something like "53°57'15.54"N"? I might have to improve that.

Regarding the warning: I get basically the same.
```** (Gramps.py:54068): WARNING **: 09:59:38.799: locale 'German_Germany.1252' isn't valid```
It is caused by:
```result = GeocodeGlib.Reverse.resolve(obj)```
Some GTK/Gramps related problem I am not familiar with. Any Expert listening?

**Comment by prculley** [comment]
> Yes, maybe my simple code to convert coordinates was implemented a bit too optimistically...
> Is the format of your location in the database is using degrees/minutes/seconds? So something like "53°57'15.54"N"? I might have to improve that.

You should look at the code in gramps.gen.utils.place it has some useful functions to convert all the possible lat/long formats. 

> Regarding the warning: I get basically the same.
> `** (Gramps.py:54068): WARNING **: 09:59:38.799: locale 'German_Germany.1252' isn't valid`
> It is caused by:
> `result = GeocodeGlib.Reverse.resolve(obj)`
> Some GTK/Gramps related problem I am not familiar with. Any Expert listening?
I would look into whatever is getting passed in the 'obj'.  I would guess that it might include a buffer with a locale attached to allow decoding from binary to ...  And the locale/encoding was not supported.  I'd guess that somewhere you are writing a buffer without specifying a specific 'utf-8' encoding, so it is inherited from the OS settings.  I suspect it would probably fail worse if you looked for a place name with characters outside of the ASCII subset.

**Comment by CWSchulze** [comment]
The gen.utils.place functions work well.

I could not remove the warning. Basically the code is:
```
loc = GeocodeGlib.Location.new(latitude, longitude, 0)
obj = GeocodeGlib.Reverse.new_for_location(loc)
result = GeocodeGlib.Reverse.resolve(obj)
```
I don't know which buffer/codec could be set. There are no additional arguments for those functions.

I copied it from gramp/plugins/lib/maps. I get the same warning when that plugin is used on my computer.

prculley:
Note: I found the width of the Gramplet too large for my screen, so reworked it to use less space.  I also found a couple of minor bugs and fixed them as well as did a bit of pylint cleanup.  Hope you don't mind.

PR #342, opened 2020-05-07
RemoveTagTool rewrite

Mattkmmr:
- allow adding tags too => Add/Remove Tag Tool
- fixes [#11648](https://gramps-project.org/bugs/view.php?id=11648)


lordemannd:
looks like some new translations needed.

PR #356, opened 2020-07-14
[Lines of descendency] Localisation update and new translations

janpython:
Hi team,

Hereby an update for the LinesOfDescendency addon.
Unfortunately, the translation appeared to work only partially, namely only the startup screen. The generated report was not translated, so I made two changes to the source code.
Also made a new template.pot and updated three translations (DE, FR, NL).
All this has been tested in Gramps 5.1.2 without any problems.

Hopefully you can handle these adjustments?
Thanks in advance!

### DIFF
--- a/LinesOfDescendency/lines-of-descendency.py
+++ b/LinesOfDescendency/lines-of-descendency.py
@@ -108,7 +108,7 @@ class LinesOfDescendency(Report):
         next_person = self.database.get_person_from_handle(path[0])
 
         self.doc.start_paragraph('LOD-Line')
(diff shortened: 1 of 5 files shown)


SNoiraud:
One remarks:
In french, we have a space before each ":"
so 
msgstr "%(line)s. ligne:"
should be
msgstr "%(line)s. ligne :"

janpython:
> One remarks:
> In french, we have a space before each ":"
> so
> msgstr "%(line)s. ligne:"
> should be
> msgstr "%(line)s. ligne :"

Ok, thank you, I have changed it!!

PR #375, opened 2020-07-26
[PlaceCoordinateGramplet] Bugfixes

CWSchulze:
The actions in the context menu seemed to be dependent on the click order. After a few clickes only "cities" could be added, without street name or building number. On the one hand a (statically allocated) default argument had been modified by removing items, on the other hand the actual list to be displayed had not been passed to the function.

Seems to be fixed now.

Maybe its worth waiting for further feedback before merging.


CWSchulze:
It seems that I'm not able to make this work in the windows AIO installer.

- Using (not importing!) GeocodeGlib I get the error:
"GLib.GError: g-io-error-quark: TLS/SSL support not available; install glib-networking (0)"
So it seems to work, but cannot properly connect to the server. No nothing can be found:
grafik
I get the same error messages in the gramps/plugins/lib/maps/placeselection.py:213 in the core. However, that exception is catched and ignored.

- Importing geopy I get the error:
"ModuleNotFoundError: No module named 'asyncio'"

bugtracker:
https://gramps-project.org/bugs/view.php?id=11872

prculley:
Rebased, Built and merged; please update the bugs in the bug tracker that you believe to be fixed.
Thanks for your contributions.

PR #446, opened 2020-08-28
New filter rule "Match associations of <person filter>"

Mattkmmr:
Requested in [#11735](https://gramps-project.org/bugs/view.php?id=11735)

### DIFF
--- a/FilterRules/associationsofpersonmatch.py
+++ b/FilterRules/associationsofpersonmatch.py
@@ -27,7 +27,11 @@
 from gramps.gen.filters.rules import Rule
 from gramps.gen.filters.rules.person._matchesfilter import MatchesFilter
 from gramps.gen.const import GRAMPS_LOCALE as glocale
-_ = glocale.translation.gettext
+try:
+    _trans = glocale.get_addon_translator(__file__)
+except ValueError:
+    _trans = glocale.translation
+_ = _trans.gettext
 
 
 # -------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


FilterRules/associationsofpersonmatch.py
```diff
@@ -0,0 +1,63 @@
...
+#
+# -------------------------------------------------------------------------
+from gramps.gen.filters.rules import Rule
+from gramps.gen.filters.rules.person._matchesfilter import MatchesFilter
+from gramps.gen.const import GRAMPS_LOCALE as glocale
+try:
+    _trans = glocale.get_addon_translator(__file__)
+except ValueError:
+    _trans = glocale.translation
+    _ = _trans.gettext
```
prculley, line 34: This is failing Travis test; I think this line (34) should NOT be indented; as this makes the '_' function unavailable unless the exception is taken.

Mattkmmr: Ah right, wrong indention. Thank you Paul!
BTW I removed the whole exception, since I do know when it's needed and I don't get errors without it.

prculley: I think that you need the exception if you run Gramps in a language for which there is no translation in the addon.

Mattkmmr: I also thought that, but I do run Gramps in German and there is no German translation for the addon yet. It's just giving a log warning, that English is used for the addon instead, but no ValueError.

prculley: I guess I should have been more specific; if you have LANGUAGE set to 'de' in your environment (indicating that you will not accept a fallback language), and there is no translation for German, you will get the ValueError.

Mattkmmr: Okay I understand, thank you for the explaination. I added the exception back. 😄

PR #410, opened 2020-08-12
Update Sync Associations with message box on actions

GaryGriffin:
Add an OK box on what actions were taken. Choices are either A) N reciprocal associations created; B) no new associations created as they already exist; C) no associations exist so no reciprocal created.

PR #455, opened 2020-09-03
[SyncAssociations] Localization fix and Dutch translation

janpython:
Dutch translation for addon SyncAssociations.
Although this tool has no dialogue with the user (which is bad and sometimes harmful), I did fix the localization anyway.

Tested in Gramps 5.1.3 on Linux Mint 20.
Please, add it to this branch.
Thanks in advance!


GaryGriffin:
I submitted a PR 3 weeks ago which adds dialogue. It is awaiting review. 
https://github.com/gramps-project/addons-source/pull/410 <https://github.com/gramps-project/addons-source/pull/410>

PR #457, opened 2020-09-03
[LifeLineChartView] Translation, navigation, map

CWSchulze:
* https://gramps-project.org/bugs/view.php?id=11905 
    Added new controls for zooming and locking ctrl for navigation in the chart. Scrolling is also possible using arrow keys (with or without shift). Additionally a small map of the chart is shown on the bottom right.
    grafik

* NL translation by Jan Sparreboom

* Translation can now be done in po format on gramps side


CWSchulze:
additional changes:
- bugfix: no death label for living persons
- bugfix: marriage location visible in label
- bugfix: mouse wheel zoom focus location is mouse location
- changed behaviour of newly introduced move lock button
- improved drawing performance by using a text rendering buffer

prculley:
Merged and built.

PR #458, opened 2020-09-04
 [PlaceCoordinateGramplet] Translation, simplify loading

CWSchulze:
I removed the connectivity check in the gpr. It slows down the startup, and also shows error messages if no internet connection is available.
Connectivity errors will now be displayed in the lineedit in the gramplet.

Additionally minor changes to the translation.


prculley:
Conflicts fixed, built and published.

PR #459, opened 2020-09-07
[GraphView] add custom avatars

vantu5z:
Inspirit by Brian ideas (send me email with avatar included).

Add presets for avatars and allow user to set it from own files.
Styles:
- Custom
- Dark (default)
- Light
- Cartoon

Avatar support can be extended to show different avatars depending on person attributes (not only gender).


emyoulation:
Several people have asked about illustrating graphs in the way described:
"Avatar support can be extended to show different avatars depending on person attributes (not only gender). "
Currently, they can do this by showing Tags. But the meaning is not intuitive... there has to be a color key to lookup what 'purple' or 'orange' means.

An example might be that they want to use a 'boat' avatar for anyone with an "immigration" event. Or a specific Attribute value.

How could the Avatar option be implemented?

(If the Avatar modification is complex, I can see a viable alternative of writing a SuperScript that inserts a particular media object at the beginning of their Gallery.)

vantu5z:
@emyoulation Do you mean something like this #477. It's not finished for now and I have no time to make it done.

emyoulation:
That looks cool!

Do you mind if I point it out to Chris Horn? It looks like something that fit nicely into his CardView.

vantu5z:
Why not, if this help to make Grams better.

PR #467, opened 2020-10-19
[Graphview] Show all connected persons

vantu5z:
Снимок экрана от 2020-10-19 13-12-12

Resolve issue [0011846](https://gramps-project.org/bugs/view.php?id=11846).

Need some tests, if it works as expected.

### DIFF
--- a/GraphView/graphview.py
+++ b/GraphView/graphview.py
@@ -74,7 +74,8 @@ from gramps.gui.dialog import OptionDialog, ErrorDialog, QuestionDialog2
 from gramps.gui.display import display_url
 from gramps.gui.editors import EditPerson, EditFamily, EditTagList
 from gramps.gui.utils import (color_graph_box, color_graph_family,
-                              rgb_to_hex, hex_to_rgb_float)
+                              rgb_to_hex, hex_to_rgb_float,
+                              process_pending_events)
 from gramps.gui.views.navigationview import NavigationView
 from gramps.gui.views.bookmarks import PersonBookmarks
 from gramps.gui.views.tags import OrganizeTagsDialog
@@ -1326,6 +1327,8 @@ class GraphWidget(object):
         """
         # set the busy cursor, so the user knows that we are working
         self.uistate.set_busy_cursor(True)
+        if self.uistate.window.get_window().is_visible():
+            process_pending_events()
 
         self.clear()
         self.active_person_handle = active_person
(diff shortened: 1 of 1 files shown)


Mattkmmr:
@vantu5z I think you have somewhere an error by getting all relationships. In my family tree I get for settings one generation up/down: The active person A and the son B, but also the partner of the son C, her sister D, the partner of her sister E, his parents F and G and their parents H and I. I think all persons after C are not indented to show up, right?

If you need help feel free to use my code for the filter rule [Degrees of seperation](https://github.com/gramps-project/addons-source/pull/463) which does search recursivly for the parents, siblings, children and partners of a person.

vantu5z:
If I right understand you, graph is:
Снимок экрана от 2020-10-20 09-15-02
But siblings C and D should have some parent X.
With active person A all persons are shown.

Correct me if I am not right.

vantu5z:
Found that busy cursor is not set on Populate (building graph), and fix this.

Mattkmmr:
> But siblings C and D should have some parent X.

Yes you are right, they have parent X and Y. When person A is selected all persons show up. Generation setting up/down is just 1.
By the way this is just one path I followed and it's hard to follow the lines, since the family structures are more difficult with 6-14 children for many families and sometimes multiple partners. I had to put the vertical spacing up to 50.

I also see one more strange thing:
C and D have a brother J. The siblings C and J are at the same generation level as B (which is the correct one), but their sibling D and her partner E are placed one level lower, which also affects the parents and grandparents of E. I don't know what is causing that behaviour. It is reproduceable and occurs again after closing Gramps.

vantu5z:
> By the way this is just one path I followed and it's hard to follow the lines, since the family structures are more difficult with 6-14 children for many families and sometimes multiple partners. I had to put the vertical spacing up to 50.

Yes, this is usable only on small parts of graph or for beginers who have added just a few persons and wants to see done work.
But users ask for this feature, so I try to add it (may it will be useful for someone).

> but their sibling D and her partner E are placed one level lower

I didn't see this behavior. If you find how to reproduce it let me know please.

vantu5z:
> but their sibling D and her partner E are placed one level lower

I can reproduce this:
Снимок экрана от 2020-10-20 14-12-20

If we have same parent P for H and Y and set active F.
I think reason is in order of generated subgraphs for families. For now I don't know if we can detect and handle this.

Mattkmmr:
> If we have same parent P for H and Y

Yes it looks the same, but there is no common parent P for H and Y in my tree. D and E are not related to each other.

PR #470, opened 2020-10-27
New web report [Heatmap]

Mattkmmr:
[General discussion about heatmaps](https://gramps.discourse.group/t/hotspot-heatmap-for-event-clusters/776)


SNoiraud:
Seems good to me.
Some comments:
1 - Could we have the map provider choice: openstreetmap, googlemap,stamenmap(terrain, watercolor, toner)
1 - The file path and file name are not clear. I needed to read the source to see we cannot put a dot in the file name. perhaps a tooltip would help.

Mattkmmr:
Thank you for your feedback @SNoiraud!

> Could we have the map provider choice: openstreetmap, googlemap,stamenmap(terrain, watercolor, toner)

I'll try to add that, but I think it's difficult to add without the [folium ](https://github.com/python-visualization/folium) package, since the different map provider use different generated code.

> The file path and file name are not clear. I needed to read the source to see we cannot put a dot in the file name. perhaps a tooltip would help.

Yes I see that's a problem. I'll improve the file path and file name options.

SNoiraud:
It's possible. In the template, I tested with:
    var tile_layer_d9991b8fadab4083808fd78ba166d6c9 = L.tileLayer(
        //"https://stamen-tiles-{s}.a.ssl.fastly.net/terrain/{z}/{x}/{y}.jpg",
        //"https://stamen-tiles-{s}.a.ssl.fastly.net/watercolor/{z}/{x}/{y}.jpg",
        //"https://stamen-tiles-{s}.a.ssl.fastly.net/toner/{z}/{x}/{y}.png",
        //"https://tile.openstreetmap.org/{z}/{x}/{y}.png",
        "http://mt1.google.com/vt/lyrs=y&hl=en&x={x}&s=&y={y}&z={z}",
        {"

remains to manage the license.

Mattkmmr:
@SNoiraud I added more map tiles providers and a tooltip for the filename menu option. I haven't added google maps, since it seems to violate the [Google Maps ToS](https://stackoverflow.com/a/59003524).

Mattkmmr:
I added more menu options which allow more control of map settings.
heatmap1
heatmap2

Mattkmmr:
More menu options updates:
* `Options` tab was renamed to `General` tab
* **Radius** option was renamed to **Point size** and added to the `General` tab
*  `Advanced options` tab was renamed to `Advanced`
* `Advanced` tab has now only options for setting a custom starting position and zoom
* **Max. zoom** and **Min. zoom** were moved from `Advanced` tab to new tab `Limits`
* `Limits` tab has options to set custom map limits (upper left and lower right lat and lon)
* `Limits` tab also has an option to render the custom map border (helps setting up the position and zoom limits)

PR #471, opened 2020-10-29
[GraphView] Improve search

vantu5z:
To improve search:
 - Use `multiprocessing.Process` to search persons;
 - Use `threading.Thread` to interact with GUI (live update popover);
 - Queues are used to connect Processes and Threads;
 - Simultaneous search and GUI update for current graph and all db persons;

Illustrating diagram:
Без имени 1

Processes and Threads runing by main Gtk thread (application).
Processes generates 2 queues with found persons (for current graph and for others).
Threads wait for some data in this queues.
As main Gtk thread is locked for other threads we use `GLib.idle_add` to update results in GUI.
Results updating by adding persons one by one (get pending  person from queue, set task by `GLib.idle_add` and wait until it's done);

Warnings:
 - ~~May not work on Mac OS X as we use [qsize()](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Queue) to calculate search progress. If it so - we should make some workaround for this.~~
 - May have some collisions with db reading as we use Processes. I test it with sqlite db and don't get any.

### DIFF
--- a/GraphView/graphview.py
+++ b/GraphView/graphview.py
@@ -126,7 +126,7 @@ gtk_version = float("%s.%s" % (Gtk.MAJOR_VERSION, Gtk.MINOR_VERSION))
 #-------------------------------------------------------------------------
 import sys
 sys.path.append(os.path.abspath(os.path.dirname(__file__)))
-from search_widget import SearchWidget, Popover, ListBoxRow, get_person_tooltip
+from search_widget import SearchWidget, Popover, ListBoxRow
 
 
 #-------------------------------------------------------------------------
@@ -1143,13 +1143,9 @@ class GraphWidget(object):
                     person_image = self.get_person_image(person, 32, 32)
                     if person_image:
                         hbox.pack_start(person_image, False, True, 2)
-                row = ListBoxRow(description=bkmark, label=name)
+                row = ListBoxRow(person_handle=bkmark, label=name,
+                                 db=self.dbstate.db)
                 row.add(hbox)
-
-                # add tooltip
-                tooltip = get_person_tooltip(person, self.dbstate.db)
-                if tooltip:
-                    row.set_tooltip_text(tooltip)
 
                 if present is not None:
                     found = True
--- a/GraphView/graphview.py
+++ b/GraphView/graphview.py
@@ -201,8 +201,8 @@ class GraphView(NavigationView):
         Method called on shutdown.
         See PageView class (../gramps/gui/views/pageview.py).
         """
-        # join threads to allow close app properly
-        self.graph_widget.search_widget.join_threads(None)
+        # stop search to allow close app properly
+        self.graph_widget.search_widget.stop_search()
 
     def font_changed(self):
         self.graph_widget.font_changed(self.get_active())
--- a/GraphView/graphview.py
+++ b/GraphView/graphview.py
@@ -196,6 +196,14 @@ class GraphView(NavigationView):
         self.define_print_actions()
         self.uistate.connect('font-changed', self.font_changed)
 
+    def on_delete(self):
+        """
+        Method called on shutdown.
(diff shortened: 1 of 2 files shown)


jralls:
> May not work on Mac OS X as we use qsize() to calculate search progress. If it so - we should make some workaround for this.

I guess that's based on the [note in the Python docs](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Queue.qsize), but that seems misinformed. `sem_getvalue` is [defined in macOS](https://opensource.apple.com/source/xnu/xnu-201/bsd/sys/semaphore.h.auto.html). Perhaps there's another reason that `multiprocessing.qsize` would raise a `NotImplemented` exception, but regardless since the docs says that it can raise an exception you *must* wrap the call in a `try-except` block and handle the exception.

vantu5z:
@jralls thanks for your feedback. I will add try-except block for this exception.

vantu5z:
Handle `NotImplementedError` exception.
Try to play with [Pipe()](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Pipe) but it has freezing effect.
So if we catch exception - progress bar will set to '0', but found person count label is still works.

vantu5z:
Sometimes get this warnings:
```
../search_widget.py:286: Warning: g_source_is_destroyed: assertion 'g_atomic_int_get (&source->ref_count) > 0' failed
  if not task or task.is_destroyed():
../search_widget.py:264: Warning: g_source_unref: assertion 'g_atomic_int_get (&source->ref_count) > 0' failed
  task = context.find_source_by_id(task_id)
```
Should I somehow fix it or this is normal behavior?

jralls:
That's because the idle has already run--it's a different thread, remember--and, quoting the ['g_main_context_find_source_by_id` documentation](https://developer.gnome.org/glib/stable/glib-The-Main-Event-Loop.html#g-main-context-find-source-by-id): "It is a programmer error to attempt to look up a non-existent source." It goes on to use *exactly this use case* as an example of the error.

vantu5z:
Hm, it's not clear why idle return id which we can't use. And why it works all time but some time send this warnings.
Do you know proper way to detect when idle is finished?

jralls:
It's intermittent because it's a race condition: Which one goes first depends on the OS's scheduler. It's a basic principal of concurrent programming, leading one to suspect that you need to spend some time studying.

You can detect that the idle is finalized, either by its function returning `FALSE` or by having `g_idle_remove...` called, with a callback passed to `g_idle_add_full`. Instead of wasting cycles busy-looping, make that callback read the next item off of the queue and set the next idle event. Note that you have a problem: If `make_search` crashes `apply_search` will block forever. Consider setting a timeout on `queue.get()`.

vantu5z:
I don't understand how to pass callback function to **idle**.
Can you give some example please?

I found [this](https://valadoc.org/glib-2.0/GLib.Idle.add_full.html). But how to use it in python?

jralls:
Well, `Glib.idle_add_full` says "AttributeError: 'gi.repository.GLib' object has no attribute 'idle_add_full'" while Glib-2.0.gir says that `idle_add` is shadowed by `idle_add_full` so I suppose that's not an option in Python. Apparently pygobject replaces the C g_idle_add with its [own version](https://developer.gnome.org/pygobject/stable/glib-functions.html#function-glib--idle-add) that takes any number of parameters to pass to the handler instead.

[This](https://python.hotexamples.com/examples/gi.repository/GLib/idle_add/python-glib-idle_add-method-examples.html#0xfb4f7f4ed2a1676f544bb924d98aec2aa399cca8c7213627e4947098b87fe261-553,,582,) shows a recursive alternative that you could adapt instead.

vantu5z:
@jralls, thanks for your help!
Make recursive call of updating GUI function by GLib.idle_add().

PR #472, opened 2020-11-10
Update themes.py

DaveSch-gramps:
Brings Themes _panel tabs into agreement with Pull

https://github.com/gramps-project/gramps/pull/1149


lordemannd:
I was looking at something else and happened to use a clean version of 5.1.3 (without PR #1149) BUT this change to themes.py  was included, and get the following crash (to Error Reporting Wizard). Question is if themes.py needs to protect itself from failing on older versions once #1149 is merged?

20206: ERROR: grampsapp.py: line 157: Unhandled exception
  File "C:\Program Files\GrampsAIO64-5.1.3-V2-clean\gramps\gui\viewmanager.py", line 672, in preferences_activate
    GrampsPreferences(self.uistate, self.dbstate)
  File "C:\Users\dave\AppData\Roaming\gramps\gramps51\plugins\Themes\themes.py", line 98, in __init__
    self.add_display_panel,
AttributeError: 'GrampsPreferences' object has no attribute 'add_display_panel'

DaveSch-gramps:
The main thing is that the _panel structure of \gramps\gui\confif.py and the _panel structure of themes.py have to agree. With Themes installed, its _panel configuration structure overrides the order of the _panels in Config.py

The only change I offered up in the pull to Themes was making the _panel structure match that of Config.py with its Themes_panel.

You sent me a note (private?) when you first tested my changes to Config.py  and Themes was not agreeing. If you made those changes to Themes, it no longer agrees with your new install that overrode config.

lordemannd:
I didn't mean to sound negative. I like the change in #1149. 
I am just worried that when this PR #472 is released, the themes.gpr.py has the line "gramps_target_version = '5.1',"
which I think means that it can be applied to 5.1.0, 5.1.1, 5.1.2, 5.1.3 and the possible 5.1.4.

the earliest version that can include PR #1149 would be 5.1.4 (if it happens), users of 5.1.0 to 5.1.3 will see the above crash.

(I am not any authority) but one solution could be to make sure that PR 1149 and 492 are included together when 5.2 is released.

Nick-Hall:
Yes, the addon should work with all 5.1.x versions.  Not everyone upgrades when a new version of Gramps is released.

Including checks for different versions is possible, but it may be better to target the changes for v5.2.

DaveSch-gramps:
I thought I put the pull onto the master and not to 5.1 and the pull should not be merged unless and until 1149 is merged.

The problem is that Themes is accessing Preferences in a one off fashion. No other addon would be able to replicate the way it is coded. Instead of adding a tab in Prefererences as an additional independent tab, its code for the _panel list supersedes the _panel list in config.py. 

https://gramps.discourse.group/t/brainstorming-preferences-interface/893/10

DaveSch-gramps:
This PR functionality was committed here.

https://github.com/gramps-project/addons-source/commit/4eba3ce26399cfebedb88c9b1bf97f0fb60bbbb1

PR #474, opened 2020-12-25
SetPrivacyTool: add filters, don't touch if unchanged

DavidMStraub:
Hi,

@Mattkmmr's set privacy tool is great, however I was missing one feature and noticed one problem. This PR fixes both:

1. It implements optionally applying custom filters to people, events, and handles. These filters are applied on top of the already existing options. Thanks to the very flexible filter system, they allow for very complex patterns.
2. I noticed the tool commits all objects it iterates over, even if it doesn't change them. This is bad (IMO) since it changes the last changed timestamp. In an extreme case, if it iterates over the entire DB and doesn't change any of the objects (e.g. because it wants to set them all public but they are already public), it still resets all the timestamps, so one looses all the information when an object was last modified.

### DIFF
--- a/SetPrivacyTool/SetPrivacyTool.py
+++ b/SetPrivacyTool/SetPrivacyTool.py
@@ -1,6 +1,7 @@
 #
 # Gramps - a GTK+/GNOME based genealogy program
 #
+# Copyright (C) 2020  David Straub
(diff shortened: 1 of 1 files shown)


Mattkmmr:
@DavidMStraub Good ideas!

PR #475, opened 2021-01-02
Addresses bug 0012134 in Addon:GetGOV

GaryGriffin:
Addresses bug 0012134 in Addon:GetGOV


GaryGriffin:
For context, see https://gramps.discourse.group/t/importing-a-place-with-getgov-fails/1077

PR #476, opened 2021-02-08
New Report for Web showing Pedigrees of the Central Person with time

mawi2021:
This new AddOn (Reports >> Web Pages >> Timeline Pedigree...) creates a website, which shows pedigrees of the central person. The box of each person is put on the website according to a timeline and its year of birth. There are several options, which can be used to change appearance of the output. However, it is planned to still extend this in the future.

pylint without option --include-ids=y is rated 9.67 for the main py file and 10 for the gpr.py file. the missing option - I didn't get working in my windows environment

Since this is my first participation in this awesome project I hope for constructive feedback.


prculley:
I did not actually try this.  A few comments based on quick scan of code.

Very nice and clean/readable code.  Thanks for trying your best to achieve a good pylint.

Why ask for center person birth year?  Why not just look it up?

Recursion by itself can be a problem if a user has an error in his tree where an ancestor is also his descendant (a loop).  We usually try to put a test in place to catch this so Gramps doesn't crash. 

We use the "register" function in our gpr.py files for addons rather than the older "newplugin" technique.  This is because the scripts that process addons to generate the downloadable and listings don't understand the "newplugin".  See any of the gpr.py files in the addons-source repo for an example.  The "register" function works fine as either an addon or one of Gramps standard plugins.

I'm not sure I understand the "missing option - I didn't get working in my windows environment"  Is this a comment about pylint?

mawi2021:
Thanks. 

> Why ask for center person birth year? Why not just look it up?
Just think of a tree without this year of birth of the first person, then, the algorithm would start with year zero. Then someone is born in let's say 1860 - this would cause a very long drawing range from year 0 to connect to 1860. If the central person has a concrete year of birth, then of course this year would be taken and all other missing years of birth of descendants are selected from database or can be estimated. The problem is with the first person only.

> I'm not sure I understand the "missing option - I didn't get working in my windows environment" Is this a comment about pylint?
Usage of pylint is described in wiki with option "--include-ids=y --reports=y", second option works, whereas --include-ids=y did not work for me.

I will take your comments of course for serious and try to get a reworked version soon. Is the usual way now to remove the current commit from git or is someone who finally (hopefully) accepts a pull request waiting for the next commit? Sorry for this kind of questions...

SNoiraud:
Tried it and have some request:
1 - The date should be conform to the date format in the Preferences
2 - You should use the symbols (death, ...)
Timeline 1
Timeline 2

mawi2021:
Hi Serge,

 

thanks for your comments.

 

Internationalization indeed is a todo left (had a remark in the code, too).

 

Which symbols are you thinking of when saying I should use them. Do you have an exmple (which report/gramplet uses this)?

 

Your first image looks like having zero in option "Offset X Coordinate", on the other hand might be the same reason with an algorithm error as you showed in example 2.

Breaking the left frame range - could you please send an example ged file (I don't care about the data, might be anonymous except for year of birth). I haven't had such an example with error in algorithm, otherwise I wouldn't have checked in of course.

 

Thanks and Best Regards,

   Manuela.

 
 

Gesendet: Mittwoch, 10. Februar 2021 um 00:04 Uhr
Von: "Serge Noiraud" <notifications@github.com>
An: "gramps-project/addons-source" <addons-source@noreply.github.com>
Cc: "mawi2021" <gramps@ur-ahn.de>, "Author" <author@noreply.github.com>
Betreff: Re: [gramps-project/addons-source] New Report for Web showing Pedigrees of the Central Person with time (#476)


 

Tried it and have some request:
1 - The date should be conform to the date format in the Preferences
2 - You should use the symbols (death, ...)

SNoiraud:
> Which symbols are you thinking of when saying I should use them.

addons: 
TimeNets/timenets.py
...

gramps plugins:
gramps/gui/widgets/fanchart.py
gramps/plugins/view/relview.py
gramps/plugins/view/pedigreeview.py
gramps/gen/utils/libformatting.py 

you can read he reload_symbols method.

example: for the death symbol:
from gramps.gen.utils.symbols import Symbols
...
        symbols = Symbols()
        death_idx = config.get('utf8.death-symbol')
        death_symbol = symbols.get_death_symbol_for_char(death_idx)

SNoiraud:
> Breaking the left frame range - could you please send an example ged file
1 - between gramps users, only gramps xml files must be shared. (gedcom could have missings info)
2 - my database contains more than 11000 persons. So I tried to reproduce the problem as I know the missing person.

timeline 3

The size of the window is the following: width=20242 and height=3498
I have X position -3363 !

In the html file, I get the following:
<br><b>X</b>
<br>* xx.yy.1961 place
<br>†  
<div style='position:absolute;top:-12px;right:0px;'>I06187</div></div>
<div class='box personbox_m' id='I07143' style='left:-3283px;top:3099px;'>
<b>Y</b>
<br><b>X</b>
<br>* xx.yy.1961
<br>†  
<hr><b>A B</b>
<br>* xx.yy.1982 
<br>†  
<div style='position:absolute;top:-12px;right:0px;'>I07143</div></div>
<div class='box personbox_m' id='I07145' style='left:-3363px;top:3315px;'>
<b>N</b>
<br><b>X</b>
<br>* xx.yy.1980
<br>†  
<div style='position:absolute;top:-12px;right:0px;'>I07145</div></div>
<div class='box personbox_m' id='I07146' style='left:-3203px;top:3339px;'>
<b>M</b>
<br><b>X</b>
<br>* xx.yy.2013
<br>†

SNoiraud:
OK. I reproduce the problem with a small example


[timeline.gramps.gz](https://github.com/gramps-project/addons-source/files/5957775/timeline.gramps.gz)

SNoiraud:
timeline 4

mawi2021:
issue left range = 0 => fixed

issue left range < 0 => fixed, too, but in my opinion you had let's say "interesting" data - two families with same mother and same child. Is this realistic? Or probably I miss any philosophy behind it. Sorry for asking.

gramps

I will upload this version later after some more changes, probably at the weekend

SNoiraud:
> "interesting" data - two families with same mother and same child. Is this realistic?

Yes, I have a dozen of these cases. The child was adopted by the new husband.
That is why the line between the new family and the old one (mother daughter) is drawn with dotted lines

prculley:
> I will take your comments of course for serious and try to get a reworked version soon. Is the usual way now to remove the current commit from git or is someone who finally (hopefully) accepts a pull request waiting for the next commit? Sorry for this kind of questions...
For a new addon like this, it doesn't matter too much how you work on it.  The easiest is to create new commits as you fix things up to the point it seems ready to release.  Then push your branch to Github.  It appears here with that history.  Use the 'Work in progress" label on this page to mark it not ready.

When it is really ready, remove that label, and one of us will do a "Squash and merge" thereby collapsing all your commits into a single one.  And then publish the final addon.  

If you don't want/need us to see the history of your PR, you can do a merge on your branch yourself, then do a 'force push' to update Github with only the commits you want us to see.

Once it is 'out there', the process should be a bit more careful, we like to say one commit for one bug or enhancement.  If several such are related, they can be in a single PR, otherwise it should be separate PRs.  If a PR depends on another, then either wait for the dependent PR to be accepted, or put a comment in the second one telling us that.

mawi2021:
Hi folks,

I changed all the points, which were mentioned in our discussion above. Please check and release.

Best Regards,
   Manuela.

PR #479, opened 2021-02-24
[GraphView] Fix all-connected feature and search

vantu5z:
Fixes for Graphview issues:
- [12220](https://gramps-project.org/bugs/view.php?id=12220) (search)
  Use `Glib.idle_add` instead of `multiprocessing`.
- [12222](https://gramps-project.org/bugs/view.php?id=12222) (instances)
  Fixed by 12220.
- [12225](https://gramps-project.org/bugs/view.php?id=12225) (all-connected recursion exception)
  Rewrite building all-connected graph without recursion.
  Add restriction to person count (<1000 persons in the graph).

### DIFF
--- a/GraphView/graphview.py
+++ b/GraphView/graphview.py
@@ -2548,6 +2548,8 @@ class DotSvgGenerator(object):
                 if num_desc > 0:  # generation restriction
                     # add every child recursively
                     for child_ref in family.get_child_ref_list():
+                        if child_ref.ref in person_handles:
+                            continue
                         self.add_connected(
                             self.database.get_person_from_handle(child_ref.ref),
                             num_desc-1, num_anc+1, person_handles)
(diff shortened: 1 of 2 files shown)


prculley:
FYI; We have a few other recursive operations in Gramps, however we usually protect them against loops by maintaining a set or dict that contains a "been there" group of object handles or IDs.  If the recursive routine starts with a check against this set, and sees that it has already processed that object, it just returns.  This protects against loops in the data, and further helps if the data has more than one way to enter a group of objects; you only have to process them once.

This may be better than just waiting for the recursive exception to occur.

vantu5z:
@prculley if I understand you right, we already have checks for processed objects. Issue is occured due to the large number of objects on displayed graph but not loops. We try build graph with more than ~1000 persons and this depth of relations can't be calculated before recursion is runing.

```
def f(i):
    print(i)
    f(i+1)
f(0)
```
This simple example excepts on 995 value with:
RecursionError: maximum recursion depth exceeded while calling a Python object

prculley:
I'm beginning to get a hint of what is probably happening.  "All connected" does imply that you will eventually touch every person in the tree that has a connection (which for most, will be everybody).  And the way your code is built that could mean a recursion for every person.  You may want to include a test for families in progress; don't recurse if the family for a person is already in progress (after all, it will get attended to when the recursion unwinds).  I think that would significantly reduce the recursion depth, although for large trees you still may get failures.

The other solution is to not use recursion.  The gen.filters.rules.person._deeprelationshippathbetween filter does something like this, however it doesn't use recursion.  It instead processes in a breadth first fashion, rather than depth first.  Maybe you can use this type of processing.

vantu5z:
Should we merge this PR to fix addon for users?
When I have time will rewrite building graph without recursion or use "Glib.idle_add" like in search module.

vantu5z:
rebased

PR #481, opened 2021-02-26
[D3Charts] D3 Chart update of d3.js and jquery libraries to new version (#11800)

mattmaniflaf:
d3 changed dramatically at v4 which is incompatible with this plugin, as such updating to the most recent v3 version.

Updating jquery to 3.5.1 and removing old unused v1 and v2 libraries.


prculley:
It looks like you might have just copied your files from master branch to gramps51; you seem to have missed some updates that are in the gramps51 branch.  This effectively removes the gramps51 bug fixes that were missed.

It would be better if you cherry-picked your commits from master branch back to gramps51.  If you do this right, the following bug fixes should end up in your commit list.
https://github.com/gramps-project/addons-source/commit/756f4af5576fc1f33bca9824d51f37003fc66daa#diff-5e47311b92e9b4202b4722d3f7a21c8d716f53e58f4f59c9a827144bc3981897
https://github.com/gramps-project/addons-source/commit/756f4af5576fc1f33bca9824d51f37003fc66daa#diff-5e47311b92e9b4202b4722d3f7a21c8d716f53e58f4f59c9a827144bc3981897
as well as many translation updates.

We have not been merging bug fixes etc. from gramps51 to master or gramps52 branches much lately, which is why these are not in master branch.  Probably won't happen until we get more serious about gramps52 release...

mattmaniflaf:
I plan to re-do all pull requests associated with my merges.
Before proceeding Can I solicit some advice first with regard to what branches I should be targeting.
My plan was to attempt to merge to:
   origin/gramps52
   origin/maintenance/gramps51
   origin/maintenance/gramps50
   
Should I be targeting origin/master as well ?

As I've not performed cherry-picks before would the sequence of git comments be something like :

# Perform intiial edits to either master or gramps52 branch:
$ git checkout gramps52
... perform my edits...
$ git commit 'My changes'
$ git push origin gramps52

# cherry-pick gramps52 back to gramps51
$ git checkout maintenance/gramps51
$ git cherry-pick <gramsp52 branch SHA from above commit>
$ git push origin maintenance/gramsp51

# Perform similar dance with gramps50 as completed above for gramsp51

Generate a pull request for each branch I want to merge so at least 3 (gramps52, maintenance/gramps51, maintenance/gramps50) and possibly a fourth for master.

Any further advice much appreciated.

prculley:
The addons repos are handled a bit differently than Gramps itself.  What I would do is to make all changes to maintenance/gramps51.  The developers (mostly me) have been merging up from that to master and gramps52 branches periodically.  I'll be doing this more regularly as Gramps master gets closer to 5.2 status and release.  This is mostly to allow testing of Gramps in the newer branches.

Currently there are no API differences that require changes to the addons for master or gramps52, but this is expected to change as time goes by.  When this happens, we may want to have the authors do work on both gramsp51 and the other branches, doing so now just makes more work for the author.  And if the changes you do to other branches create conflicts with merge up, then more work for me.

Note: we have NOT been supporting gramps50 or prior branches at the developer level, so if you want to back port your work to that branch for older version users, then you should do that as well in additional PRs.

mattmaniflaf:
Thanks Paul, appreciate the feedback. I'll close this PR now, apply my changes to maintenance/gramps51 and submit a new PR for that.

PR #484, opened 2021-03-03
[Translation] Addon Placecompletion Hebrew v5.2

avma:
Add Hebrew .po


Nick-Hall:
@avma Is there anything in this pull request that still needs to be merged?

avma:
I think they can all go. (I do have local backups so worst case ...; )) Can/should i remove them?

Nick-Hall:
It looks like these translations have already been merged.

PR #485, opened 2021-04-01
multiple parents filter

DaveSch-gramps:
modified multiple marriage filter (_muliiplemarriage.py) to find multiple parents


DaveSch-gramps:
https://gramps.discourse.group/t/identifying-individuals-with-multiple-sets-of-parents/1336

prculley:
Since you based this on 'master' branch, I had to do this manually, should have been on maintenance/gramps51 branch.  But it is now accepted, built and published.

PR #491, opened 2021-05-01
[GraphView] Add configuration for the box size around each person node

gburto01:
The thickness of the box surrounding each person node was previously fixed and set at a very small value and therefore it was difficult to see the colour that has been configured. This PR allows the user to configure the border size so that it is possible now to set the box colour as white and set a thick border so that the border colour can be used to show the gender of the person.


gburto01:
An example of using the border around the person nodes to illustrate gender.

image

gburto01:
The border is configured on the Themes tab. Two configuration settings are provided; one for the active person and another for all the rest of the person nodes.

image

prculley:
Manually merged.

PR #506, opened 2021-05-09
[GraphView] Implement configuration for changing the graph direction

gburto01:
Until now GraphView has drawn the tree with time flowing vertically from top to bottom. This PR implements configuration for drawing the tree in four different directions:

Vertically, top to bottom (the default)
Vertically, bottom to top
Horizontally, left to right
Horizontally, right to left


prculley:
manually merged due to conflict with other PR.

PR #511, opened 2021-07-06
SetPrivacyTool enhancements

Mattkmmr:
Requested on [discourse](https://gramps.discourse.group/t/changing-event-privacy-in-bulk/333/5):
- set addresses private
- set internet objects private

SetPrivacyTool


emyoulation:
[Wiki updated to add 1.0.9 docs](https://gramps-project.org/wiki/index.php/Addon:SetPrivacyTool#Version_1.0.9).  (1.0.8 section will remain until this version is released to the Plug-in Manager.)

PR #515, opened 2021-08-29
New filter rule [peopleeventscount]

Mattkmmr:
Filter for persons which have multiple events of given (event) type.
Requested in [#5494](https://gramps-project.org/bugs/view.php?id=5494)

### DIFF
--- a/FilterRules/peopleeventscount.gpr.py
+++ b/FilterRules/peopleeventscount.gpr.py
@@ -20,15 +20,15 @@
 """Matches persons which have multiple events of given type."""
 
 register(RULE,
-  id = 'personswithmultipleevents',
-  name = _("Persons with multiple events of <type>"),
-  description = _("Matches persons which have multiple events of given type."),
+  id = 'peopleeventscount',
+  name = _("People with <count> of <event>"),
+  description = _("Matches persons which have events of given type and number."),
   version = '1.0.0',
   authors = ["Matthias Kemmer"],
   authors_email = ["matt.familienforschung@gmail.com"],
   gramps_target_version = '5.1',
   status = STABLE,
-  fname = "personswithmultipleevents.py",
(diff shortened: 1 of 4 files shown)


emyoulation:
Could you change this to search for a `<count>` instead of multiple? Then it could be used for zero counts, single counts & more than 2 counts too.

Please consider renaming to "`People with <count> of <event>`"

Mattkmmr:
@emyoulation You can already search for people with 0 occurences of an event of a certain type with existing filters. Person filter **persons with event** e.g. birth events, then inverse results.

emyoulation:
That's true. But... if this rule had that capability plus single count, plus >zero, plus >n; then we could justify a request to Eliminate the single purpose rules ... which simplifies selection of rules for confused users.

Mattkmmr:
Maybe it'd be more universal to add the counter to the built-in filter rule `People with the personal <event>` instead.

emyoulation:
That makes sense but would that work? The rule name would have to be changed to include `<count>` either way. If the rule name changes, is it the same Rule?

On the flip side, since this would be an enhancement & in the core, which means it cannot go public until the 5.2 release. And any non-bug tweaks would be delayed until the NEXT enhancement release. Worse, the prototype only get tested by users with Development level tech skills. Not the average users who do unexpected & sometimes illogical things.

That throttles evolution too severely. The Core enhancement cycle is just too rigid to get fast, comprehensive evolution.  

So an add-on that can evolve quickly then supplant the core function when stable would seem to be the best trade-off in choices.

Mattkmmr:
@emyoulation I rewrote the filter rule and renamed it to `People with <count> of <event>`. It has now a number field and "greater than", "equal", "smaller than" selector. It also got a checkbox for the primary event roles. 😄

prculley:
Built and in addons repo.  Thanks.

PR #518, opened 2021-09-28
[GraphView] Show direct ancestors of two related people

gburto01:
A feature to show the direct ancestors of the "home" person and any other selected person, back to their common ancestors. For example, in this tree, Ann Abram is the home person.

image

Selecting "Show path to home person" for Isabel Blundell redraws the tree as follows, showing Isabel & Ann's direct ancestors back to their common ancestors, John & Ellen.

image

### DIFF
--- a/GraphView/graphview.py
+++ b/GraphView/graphview.py
@@ -88,6 +88,8 @@ from gramps.gui.pluginmanager import GuiPluginManager
 from gramps.gen.plug import CATEGORY_QR_PERSON, CATEGORY_QR_FAMILY
 from gramps.gui.plug.quick import run_report
 
+from gramps.gen.filters import GenericFilterFactory, rules
+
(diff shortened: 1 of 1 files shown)


SNoiraud:
I have a problem with this. You lose the history. It is impossible to go to the previous page.

emyoulation:
It is not impossible but the method IS invisible. (Basically, changing any setting or focus goes back to the broader view of data.)

And the original access to the feature is rather well-hidden too. A chart illustrating a pair of distant cousin's lineage to a common ancestor is a frequent request. 

It is wonderful that diagramming this relationship doesn't require changing the Active Person... with all the refreshes that accompany that change of focus drives.

I'm wondering if the other 'relationship path' methods might be added as options? Like the one that shows ALL paths between 2 people.

Or an option that include spouses of all the people on the path?

PR #523, opened 2021-12-26
PostgreSQL: don't re-create collation if it already exists

DavidMStraub:
Currently, the PostgreSQL plugin drops the existing collation and creates it again on each database connection. That shouldn't be necessary and creates [issues with Web API](https://github.com/gramps-project/gramps-webapi/issues/262). So this PR changes the behaviour to only create the collation for a given locale if it doesn't exist yet.


Rahel-A:
I've tested this patch with both gramps and gramps-web-api, although I only used a single instance of gramps, and multiple instances of gramps.js (frontend).
Seems OK to me! :+1:

DavidMStraub:
Change base branch to `maintenance/gramps51`

PR #524, opened 2021-12-29
Add SharedPostgreSQL add-on

DavidMStraub:
This database add-on is a variation of the existing PostgreSQL add-on which allows to store an arbitrary number of family trees, rather than just a single one, in a Postgres database.

The original PostgreSQL add-on works as follows: it assumes the existince of a database with the name being the name of the family tree in Gramps. It then creates the database tables exactly like in the SQLite core module.

The new shared PostgreSQL add-on works on a single database (assumed to be named `gramps` by default). It has the same database tables as the original one, but adds a new column `treeid` to all tables and a new table `trees` with two columns: `treeid` and `uuid`. When a new family tree of type `sharedpostgresql` is created in Gramps, it generates a random UUIDv4, stores it in `settings.ini` in the family tree's directory (where also the host and port are stored just like in the original add-on) and adds a row to the `trees` table, linking the UUID to an integer. This integer `treeid` is set for all objects in the other tables. In the end, for instance, the `person` table can contain people from arbitrarily many family trees, but the data doesn't interefere because all the SQL queries have `WHERE treeid = ?` in them to filter out only the data from a single tree. The indices are adapted to be on the `(treeid, handle)` tuple rather than just the handle. Only this tuple is required to be unique, not the handle itself.

The motivation for creating this add-on is that it allows to build a generalized version of Gramps Web API where a single server hosts arbitrarily many family trees (each user only having access to a single tree).

### DIFF
--- a/SharedPostgreSQL/po/de-local.po
+++ b/SharedPostgreSQL/po/de-local.po
@@ -1,0 +1,82 @@
+# German translation for Gramps
+# This file is distributed under the same license as the Gramps package.
+# translation of de.po to Deutsch
+#
+#
+# Anton Huber <anton_huber@gmx.de>, 2005,2006.
+# Sebastian Vöcking <sebastian_voecking@gmx.de>, 2005.
+# Sebastian Vöcking <voeck@web.de>, 2005.
+# Martin Hawlisch <martin@hawlisch.de>, 2005, 2006.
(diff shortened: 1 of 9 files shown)


DavidMStraub:
Changed base branch to `maintenance/gramps51`

PR #528, opened 2022-01-10
Dynamic web: Gender fill colour hex incomplete

SNoiraud:
Fixes #012513

Add an option to select a light or a dark theme
when the background gender option is selected.


prculley:
built and listed

PR #529, opened 2022-02-15
Fixes #0012103 Validate person handle before using

hgohel:
If either mother or father person handles could not be obtained, don't try to get details about the person because it raises an exception and leads to the crashed described in the bug report.

Fixes [#12103](https://gramps-project.org/bugs/view.php?id=12103)

### DIFF
--- a/FamilyTree/FamilyTree.py
+++ b/FamilyTree/FamilyTree.py
@@ -660,21 +660,19 @@ class FamilyTree(gramps.gen.plug.report.Report):
 
     def __family_get_display_name(self, family):
 
+        father_name = _("Unknown")
         father_handle = family.get_father_handle()
+        if father_handle:
+            father = self.database.get_person_from_handle(father_handle)
+            if father:
+                father_name = self.__person_get_display_name(father)
+
+        mother_name = _("Unknown")
         mother_handle = family.get_mother_handle()
-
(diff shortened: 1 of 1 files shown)


lordemannd:
tested working on windows with 5.1.5, sqlite and bsddb

Nick-Hall:
This fix looks good to me.   Is it intended for the _gramps51_ branch?

lordemannd:
user hidp123 reports it also solves #0012568

hgohel:
> Is it intended for the _gramps51_ branch?

@Nick-Hall Since 5.1 is the current release, it makes sense to submit it there. There are two branches:
1. gramps51
2. maintenance/gramps51
Which one should I target the PR to?

PQYPLZXHGF:
According to the [readme](https://github.com/gramps-project/addons-source#readme)  use  **maintenance/gramps51**

> Most of the time addons should be developed to work with the current released version of Gramps (**maintenance/gramps51** for the Gramps 5.1.x versions for example).

hgohel:
Changing the base didn't work as easily as expected. I will close this PR and open a new one.

hgohel:
Please review new PR #530 targeted to [gramps-project:maintenance/gramps51](https://github.com/gramps-project/addons-source/tree/maintenance/gramps51).

PR #530, opened 2022-02-16
[FamilyTree]: Fixes #0012103 Validate person handle before using.

hgohel:
If either mother or father person handles could not be obtained, don't try to get details about the person because it raises an exception and leads to the crashes described in the bug report.

Fixes [#12103](https://gramps-project.org/bugs/view.php?id=12103)


reinhard-mueller:
As I was asked explicitly (by email) to review this (as I'm the original author of the report), I had a look and the change is very plausible and clean. From my side it's fine to be merged. Thank you @hgohel for the fix!

PR #533, opened 2022-03-26
[CombinedView] Fixes #0011814 Validate father handle before using.

hgohel:
If father handle can not be obtained, don't try to get details about the father because it raises an exception and leads to the crash described in the bug report.

Fixes [#11814](https://gramps-project.org/bugs/view.php?id=11814)


hgohel:
@Nick-Hall Requesting review and approval of this PR. The code has been tested with Gramps 5.1.5 and the crash is eliminated. Thanks.

Nick-Hall:
@hgohel Thanks for reminding me.  I'll try to get this done tomorrow.

prculley:
Built and listed, thanks.

PR #534, opened 2022-04-02
Add 1950 Census

pixelasticity:
The 1950 Census data has been made public and the forms add-on doesn't have it yet

### DIFF
--- a/Form/form_us.xml
+++ b/Form/form_us.xml
@@ -2768,7 +2768,8 @@
             <_attribute>E. D. Number</_attribute>
         </heading>
         <heading>
-            <_attribute>Hotel, large rooming house, institution, military installation, etc.</_attribute>
+            <_attribute>Institution</_attribute>
+            <_longname>Hotel, large rooming house, institution, military installation, etc.</_longname>
         </heading>
         <heading>
             <_attribute>Page No.</_attribute>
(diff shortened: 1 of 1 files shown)


GaryGriffin:
The proposed 1950 US Census form is not consistent with previous (1900 - 1940) forms. Previous year forms do not include the street/number/index as fields in the form. And all the other forms have the name of the person in the first field. I suggest the first 3 fields be removed from the form. And that the first field in the form should be the persons name.

pixelasticity:
I like having the form match the census as closely as possible, but I understand wanting it to match previous forms. I updated my merge request using your notes.

prculley:
I merged this into gramps51 and rebuilt from there as this will allow other people to see it.

kpaffy:
I think the following fields are missing:

- Name of street, avenue, or road
- House (and apartment) number
- Serial number of dwelling unit
- Agricultural questionaire number

Given these are usually shared among a group, I think they belong in the headers. There is precidence from prior censuses. Also I think the farm and acre items should be with them.

fxtmtrsine:
Fyi 

> The 1950 Census in form_us.xml introduced in Forms gramplet version 2.0.33 contains two instances of a column with _attribute key "Class of worker". While this accurately mirrors the headings of columns 20c and 35c on the original census form, it leads to data loss since the value of the second column will overwrite that of the first column when the attributes are saved. Everything looks OK during data entry, but upon closing and reopening the Form Editor window, both "Class of worker" columns will contain the value entered for the second instance, which is usually blank since only one person of the up to 30 per sheet has a column 35c. Viewing the attributes in the Event Reference Editor or a backup file confirms that the data entered for the first instance of the "Class of worker" column is truly lost.

Posted on the gramps bugtracker and includes a patch:

* [0012637: Forms gramplet: data loss due to duplicate key in US 1950 Census](https://gramps-project.org/bugs/view.php?id=12637)

Caulley:
An updated vision of the 1950 Census can be found at [.](https://github.com/Caulley/Gramps-form_us/blob/main/1950%20Census) This is ONLY the 1950 Census not the complete form_us file. Comments are requested.

PR #536, opened 2022-05-11
Update Forms definitions to be ISO 8601 dates

Wdavery:
This updates the two non-ISO compliant files, the rest are already in ISO 8601.


prculley:
Are you sure you want this in master branch?  It may be a long time (years) before another major release of gramps from master branch occurs.  Usually it makes more sense to apply addons changes to the current maintenance branch, as these eventually get merged into master branch anyway.

Wdavery:
Not sure at all. Whatever makes the most sense. The branches in this repo are not intuitive to the uninitiated. 
Should it be 'maintenance/gramp51'?

Nick-Hall:
Rebased onto _gramps52_ branch.

PR #539, opened 2022-06-02
Adding events to person and family pop menus.

SNoiraud:
Add the possibility to show and edit person or family events.


Nick-Hall:
It looks like this was already merged into _gramps52_ as part of PR #583.

See commit a2bd5a19cda1d2996c282364e59b57a4e9e07158.

PR #540, opened 2022-06-04
fix: recast list to tuple on load, pylint, tweaks

cdhorn:
After load recast history items that are lists back to tuples, renames some variables to better describe intent, reformatted with black, numerous changes to make pylint happier.


GaryGriffin:
Should this be rebased to gramps60?

PR #548, opened 2022-07-23
Fixes #11596 GetGOV language setting is case sensitive

hgohel:
Gramps Preferences > Display > Place Format's language setting allows input in mixed case, but GetGOV addon interprets it correctly only when the language code is in lower case otherwise it defaults to German ('de').

This change converts the value of the language field to lower case when used by the addon so that GetGOV accepts it and returns results in that language (as far as possible).

Fixes #11596

@GaryGriffin A review of this change would be appreciated. Thanks.

### DIFF
--- a/GetGOV/getgov.py
+++ b/GetGOV/getgov.py
@@ -332,7 +332,7 @@ class GetGOV(Gramplet):
         except AttributeError:
             fmt = config.get('preferences.place-format')
             pf = _pd.get_formats()[fmt]
-            preferred_lang = pf.language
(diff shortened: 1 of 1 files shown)


hgohel:
It's a bit odd that the code path being executed is in the exception handler. Does Line 331 support how older versions of Gramps stored the language setting?

GaryGriffin:
Code change looks good.

Yes, the way the place is stored is different in older versions, so an exception handler was needed to not break compatibility.

PR #552, opened 2022-12-02
Update default known Associations

GaryGriffin:
Removed cM Association. It is deprecated.
Changed Grandfather/Grandmother to Grandparent.


emyoulation:
Won't this update overwrite customized lists? 

The next update needs to have the ability edit it list in a survivable form and NOT just be a changed ;ost/

See 
https://gramps.discourse.group/t/share-your-patches-for-the-syncassociations-add-on/2861

PR #554, opened 2022-12-14
New version of PrerequisiteCheckerGramplet and some troubleshooting.

SNoiraud:
1 -  New version imported from Bug report 012769
    
    The text reformatting includes:
        swapping in leading spaces & bullets for asterisks.
        italicizing descriptive text
        composing backup text and correcting menu labels
        making the Required-Recommended-optional section labels more consistent in tense
    
    Updated Help URLs
       Changed & tested the Help URLs to the new standard "Addon:" prefix and added a
       "in document" scrolling target. (Including the help_url in this repository's Gramps
       Plugin's Registration file.)
    
2 -  PrerequisitesChecker gramplet: Gexiv2 and ICU
    
    Detection of these libraries does not work or partially.
    
3 - If the Prerequisites Checker Gramplet is active when Gramps start, the Gramplet runs the report 3 times

Fixes #012767, #012769, #012770

### DIFF
--- a/PrerequisitesCheckerGramplet/PrerequisitesCheckerGramplet.gpr.py
+++ b/PrerequisitesCheckerGramplet/PrerequisitesCheckerGramplet.gpr.py
@@ -17,22 +17,22 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
-#------------------------------------------------------------------------
+# ------------------------------------------------------------------------
 #
 # Diagnostic check that prerequisites have been met for the current install.
 #
-#------------------------------------------------------------------------
+# ------------------------------------------------------------------------
 
 register(GRAMPLET,
          id="Prerequisites Checker Gramplet",
          name=_("Prerequisites Checker"),
-         description = _("Prerequisites Checker Gramplet"),
-         version = '0.8.37',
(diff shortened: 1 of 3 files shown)


emyoulation:
After copying over the source, the Gramplet only shows "Diagnostic Gramplet to help evaluate if Gramps has all prerequisites installed." and stops.

emyoulation:
I had not tried it as a Dashboard. (My dashboard is Full. So it is an awkward place for this Gramplet that will normally be used once and then be unloaded. Since it needs maximum vertical space, it fits the sidebar best.)

In the sidebar and the ”self.count = 3”, the report only shows Twice when loading a different tree database. That's a rare situation - so it is livable. 

It no longer shows Multiple reports if the Hramplet is loaded when you restart Gramps. (Restarts are likely if the user is trying to fix one missing perequisites at a time.)

But WITHOUT the "self.count = 3”, the sidebar only shows the 1st line of the report. That is not usable.

SNoiraud:
> But WITHOUT the "self.count = 3”, the sidebar only shows the 1st line of the report. That is not usable.

NO. as we are in a list view, the database must be opened. It works like that. I’m not changing that.
I think this plugin is heresy. You have gramps -V to have the same results.

emyoulation:
Sorry, I had Gramps set to autoload the last used Tree but turned that feature off for the previous bug test.

I've removed the override on my installation. 

Thank you for including the updated help_url s

PR #556, opened 2023-02-18
Add Spanish translations for some plugins

jcsaaver:
Add Spanish translations for some plugins:

- Descendant count
- DNA
- Media Verify
- Date calculator


romjerome:
Thank you !

PR #564, opened 2023-07-18
Fix missing space leading to syntax error

DavidMStraub:
Dear maintainers,

I found this bug that breaks Gramps Web instances using the shared PostgreSQL addon. I would appreciate if you could merge this as soon as possible. Thanks a lot!


Nick-Hall:
@DavidMStraub Thanks.  I'll do it right now.

DavidMStraub:
Thanks, that was record timing :rocket: 

Will update the docker image now.

PR #565, opened 2023-07-20
Fix another SQL syntax error

DavidMStraub:
@Nick-Hall I'm really sorry but it turned out my fix yesterday #564 was not sufficient because there was an additional error in this statement. Can you please merge & build also this fix? Thanks!


DavidMStraub:
Thanks!

PR #570, opened 2023-09-06
Polish translations for Themes

stojex:


GaryGriffin:
Should this be branch gramps52 (not gramps51) since 52 is current.

GaryGriffin:
And gpr does not need to increment version. This is done automatically on publish. Should remove the gpr change from this PR.

PR #571, opened 2023-09-06
Polish language for DownloadMedia

stojex:


GaryGriffin:
Should this be branch gramps52 (not gramps51) since 52 is current.

GaryGriffin:
And no need to increment version in gpr file. That is down automatically on publish. Plz remove gpr change from PR

PR #578, opened 2023-10-19
New Gramplet: DNA matches (Gramps 5.1)

RampantSpirit:
Hey, I made this Gramplet for personal use but thought others may find it useful as well, so here it is. 

DNAMatches

DNA Matches will list the active person's DNA matches including information about their relationship, cMs, number of segments, largest segment and sources. It is designed to complement the [DNA Segment Map](https://gramps-project.org/wiki/index.php/Addon:DNASegmentMapGramplet) and also works with the [Sync Associations](https://gramps-project.org/wiki/index.php/Addon:SyncAssociation) gramplet, but can be used without them. 

Information is gathered from Notes in Associations of type "DNA" between persons, same as with the DNA Segment Map (exact same format). However, there is also an alternative format for websites that don't have chromosome browsers such as Ancestry. This takes the form of a note with a single line specifying shared DNA, shared segments and the largest segment, in that order, separated by tabs or commas, like so:

`168,5,68`

You can see one Ancestry match in the attached screenshot above to give you an idea.

This initial version also comes with a Slovak translation by myself as I am Slovak. 😄

Btw, I did spin this off from the DNA Segment Map initially so I left the original developers' credits in the files. Hope that's ok.

Let me know if there are any issues!

### DIFF
--- a/DNAMatches/dnamatches.gpr.py
+++ b/DNAMatches/dnamatches.gpr.py
@@ -35,7 +35,8 @@ register(GRAMPLET,
          gramplet_title=_("DNA Matches"),
          detached_width = 600,
          detached_height = 450,
-         version = '1.0.1',
+         version = '1.1.0',
          gramps_target_version='5.1',
+         help_url="Addon:DNAMatches",
          include_in_listing = True
          )
--- a/DNAMatches/dnamatches.gpr.py
+++ b/DNAMatches/dnamatches.gpr.py
@@ -1,0 +1,41 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+# http://gramps-project.org
+# Gramplet registration - plug-in/add-on to extend Gramps
+#
+# Copyright (C) 2020        Nick Hall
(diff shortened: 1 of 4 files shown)


GaryGriffin:
The help_url specified does not exist.

Should this be a report instead of a gramplet? The output is not actionable (clickable), so feels more like a report.

Should this be moved to 5.2 (from gramps51)? Or in both branches?

RampantSpirit:
Apologies for the delayed response!

> The help_url specified does not exist.

This has now been created: https://gramps-project.org/wiki/index.php/Addon:DNAMatches

> Should this be a report instead of a gramplet? The output is not actionable (clickable), so feels more like a report.

It is possible to interact with the table. Left-clicking will make the person you click on in the table the active person. Right-clicking will open the relevant DNA association. 

Personally, I'm not particularly interested in making this into a report, but someone else is welcome to do so. 😄

> Should this be moved to 5.2 (from gramps51)? Or in both branches?

I'm happy for it to be available on both!

PR #579, opened 2023-10-23
Update GrampsWebSync to version 1.1.0

DavidMStraub:
Dear @prculley,

this is another update of the Gramps Web Sync plugin.

It fixes a bug (related: https://github.com/DavidMStraub/gramps-web-sync/issues/22) and adds a new feature, the option to reset one of the two sides to the other side.

### DIFF
--- a/GrampsWebSync/const.py
+++ b/GrampsWebSync/const.py
@@ -12,7 +12,6 @@ class Type(metaclass=TypeMeta):
     """Workaround for missing typing module in Gramps AIO."""
 
 
-
 try:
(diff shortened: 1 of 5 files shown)


DavidMStraub:
Hi, I'm demoing Gramps Web at a genealogy convention today and am hoping for a number of new Gramps & Gramps Web users. It would be awesome if the sync plugin could be updated soon so the new users don't run into this bug.

Nick-Hall:
@DavidMStraub I'll do it now.

DavidMStraub:
Thanks!

PR #580, opened 2023-10-30
Fix bug 13075 - SyncAssociation BatchTool

GaryGriffin:
Added BatchTool to SyncAssociation


emyoulation:
Could you rev the version in the .gpr.py too?

Nick-Hall:
No need to update the version number.  It is updated automatically during the build.

PR #581, opened 2023-11-10
Update DescendantBooks.gpr.py

emyoulation:
Register as Books capable Reports list.


GaryGriffin:
Should this be in master or maintenance51?

emyoulation:
Probably in both 5.1 and 5.2 maintenance branches.

Nick has suggested that the addons-source "master" branch serves no function and might be eliminated.

GaryGriffin:
please change to appropriate branch (gramps51 or gramps52).

Nick-Hall:
Rebased onto _gramps52_.

PR #582, opened 2023-12-03
Improve the robustness of finding form events

stevenyoungs:
Previously, for a given form event, the citation was stored. If later, the form was edited, an attempt was made to locate the correct event, given just the citation. If the citation was shared by multiple events, this was prone to return a different event than we started with.

This change tracks the original form event and passes it into EditForm directly, removing the need for the reverse lookup.


Nick-Hall:
Yes.  This looks good.

I'm not sure if we need to pass both the event and citation to the form editor, but it may be more convenient to do so.

Something we have to think about for the future is if we want to generate secondary events from the form.  For example, creating birth, residence and occupation events for a census.

Nick-Hall:
Rebased onto _gramps52_ branch.

PR #583, opened 2024-01-18
Graphview search+id+events

SNoiraud:
Adding 4 fonctionalities

1 - The possibility to add the gramps ID (option).
2 - The possibility to edit the person or family eventd if they exists.
3 - The search is always using the regexp to have a better selection.
4 - Improve the information in the menu tooltip.

In this case, when the mouse is over a searched person, we can see:
- the birth and/or death date with their respective place,
- the  partners list with birth/death date
- the children list  with birth/death date
- the parents with birth/death date

graphview


elcarpentero19:
Thanks!

PR #593, opened 2024-03-11
Update PrerequisitesCheckerGramplet.gpr.py

emyoulation:
The Prerequisites Checker will only give results in the Dashboard.  Add 'navtypes' plugin registration restriction for that category


emyoulation (commented):
has an extra comma at the end.

emyoulation:
see   "[13246](https://gramps-project.org/bugs/view.php?id=13246): PrerequisitesCheckerGramplet registration should be restricted to Dashboard"

PR #595, opened 2024-04-12
Update BiographyQuickview.py

emyoulation:
use bullets instead of asterisks


GaryGriffin:
Please change this from master to gramps52

Nick-Hall:
Rebased onto _gramps52_.

PR #605, opened 2024-09-16
Fix edge case on common ancestor calc and added visual phasing

GaryGriffin:
Fix the edge case where the relationship between 2 people can be determined but there is no common ancestor in the tree. Specifically the relationship leverages siblings where the parent does not exist in the tree.

### DIFF
--- a/DNA/dnasegmentmap.py
+++ b/DNA/dnasegmentmap.py
@@ -46,6 +46,7 @@ from gramps.gen.const import GRAMPS_LOCALE as glocale
 from gramps.gen.config import config
 import random
 import re
+import csv
 from gramps.gen.relationship import get_relationship_calculator
 _ = glocale.translation.gettext
 
@@ -66,6 +67,14 @@ CONFIG.register('map.legend-char-height',12)
 CONFIG.register('map.include-citation-notes',0)
 CONFIG.register('map.chromosome-x-scale',1.4)
 CONFIG.register('map.chromosome-y-scale',1)
+#
+# disabled config options
+#
+#CONFIG.register('map.association-string',"DNA")
+#CONFIG.register('map.output-visual-segments',0)
+#
+# Visual Phasing only to grandparent, not including great-grandparent
+#
 
 CONFIG.init()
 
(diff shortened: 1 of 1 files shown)


emyoulation:
That worked.  Thanks!

emyoulation:
I was unfamiliar with the term "visual phasing".  Could you introduce it in the Addon's wiki page?

Asked Perplexity ai to give a history of the term and introduce the concept. Some of it might be useful:

> Visual phasing is a powerful technique in genetic genealogy that allows researchers to assign DNA segments of three full siblings to their four grandparents by comparing crossover positions on specific chromosomes[1][2]. This method was developed by Kathy Johnston and Randy Whited as a way to reconstruct ancestral DNA profiles without having direct access to parental or grandparental genetic data[1].

## History and Development

The concept of visual phasing emerged as genetic genealogists sought ways to leverage autosomal DNA data from multiple siblings to infer ancestral genetic information. While the exact date of its inception is not specified in the available sources, the technique gained prominence in the mid-2010s[2].

## Key Components

Visual phasing requires:

1. Autosomal DNA raw data files from three full siblings (or more) uploaded to a platform like GEDmatch
2. Multiple 2nd or 3rd cousins on several ancestral lines for comparison
3. Specialized tools, such as Steven Fox's automated Excel spreadsheet, available through dedicated online communities[1]

## Process and Applications

The process involves analyzing the crossover points where genetic recombination has occurred in each sibling's chromosomes. By comparing these points across all three siblings, researchers can deduce which segments came from which grandparent[1][2].

Visual phasing has become an important tool in genetic genealogy, allowing researchers to:

- Reconstruct ancestral DNA profiles
- Assign DNA segments to specific grandparents
- Enhance understanding of genetic inheritance patterns

While the technique is most effective with three or more siblings, variations of the method can be applied to scenarios with fewer siblings or half-siblings, though these applications are more complex and may yield less comprehensive results[1].

As the field of genetic genealogy continues to evolve, visual phasing remains a significant methodology for those seeking to unravel their genetic ancestry without direct access to older generations' DNA data.

Citations:
[1] https://geneticgenealogygirl.com/en/the-basics-of-genetic-genealogy/visual-phasing/
[2] https://isogg.org/wiki/Visual_phasing
[3] https://www.ncbi.nlm.nih.gov/pmc/articles/PMC9407302/
[4] https://www.fsigenetics.com/article/S1872-4973(23)00045-5/fulltext
[5] https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8774801/
[6] https://digitalcommons.liberty.edu/cgi/viewcontent.cgi?article=2306&context=honors
[7] https://www.sciencedirect.com/science/article/pii/S1872497323000455
[8] https://dune.une.edu/cgi/viewcontent.cgi?article=1004&context=edu_diss

emyoulation:
The Visual Phasing term is used (but not introduced) at:

https://gramps-project.org/wiki/index.php/Addon:DNASegmentMapGramplet#Grandparent_.28Phasing.29_View

GaryGriffin:
Updated documentation to define visual phasing and provide link for additional info. Also discusses great grandparent level phasing potential, which is not currently available (due to UI concerns). 

If there are future user requests for this capability, it can be added - at the price of a more confusing display.

PR #612, opened 2024-10-13
Fix  bug #0013468 

CameronD73:
Bugs where shown when caches were not cleared on DB reload and some cache content was not cleared on (birth) event change/addition


CameronD73:
1. add event-update callback to accommodate simple edits to existing events
2. add _birth_date cache clearance when personal data is added or edited (previous code only reset text cache.
3. clear both caches on loading a different tree.

PR #614, opened 2024-10-20
created GraphView italian translation

SebastianoPistore:
Created GraphView italian translation


GaryGriffin:
@tosky Please review. If approved I will publish.

PR #620, opened 2024-12-08
Update 1921 England Census form: Separate age years and age months

stevenyoungs:
The UK 1921 census form has separate columns for a persons age in  years and months.
Store these are separate columns in the gramps form definition.
The existing gramps UK1921 form Age column name has not been renamed to maintain backwards compatibility.

We have no way to know how someone may have encoded the years and months columns from the printed form into the single column in the gramps form. This makes it challenging to migrate existing data to the new structure.

### DIFF
--- a/Form/form_gb.xml
+++ b/Form/form_gb.xml
@@ -740,6 +740,11 @@
                 <_longname>AGE</_longname>
             </column>
             <column>
+                <_attribute>Age Months</_attribute>
+                <size>5</size>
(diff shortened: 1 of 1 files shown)


GaryGriffin:
@Nick-Hall 
Nick, are you the reviewer for UK forms? I wanted to add a reviewer to this PR but not sure who.

Is there a wiki page that lists the reviewers by country?

PR #621, opened 2024-12-10
Date Calculator Enhancements

stevenyoungs:
Changes to the date calculator gramplet as discussed in discourse [here](https://gramps.discourse.group/t/update-to-uk1921-form/6581/2?u=stevey) and bug [#12345](https://gramps-project.org/bugs/view.php?id=12345)

In summary:
- add tooltips to the entry widgets
- use Gtk.Entry instead of Gtk.TextView for a more compact display
- use displayer to format the result to match users configuration
- set the qualifier on the result to calculated
- add a copy button for easier copying to the clipboard

credit to the commenters for their ideas.

image
.

### DIFF
--- a/DateCalculator/DateCalculator.py
+++ b/DateCalculator/DateCalculator.py
@@ -79,14 +79,18 @@ class DateCalculator(Gramplet):
         self.result = self.__add_entry(_("Result"))
         self.result.set_editable(False)
 
-        bbox = Gtk.ButtonBox()
+        button_box = Gtk.ButtonBox()
+        button_box.set_layout(Gtk.ButtonBoxStyle.START)
+        button_box.set_spacing(6)
+        button_box.set_border_width(12)
+
         apply_button = Gtk.Button(label=_("Calculate"))
         apply_button.connect('clicked', self.apply_clicked)
-        bbox.pack_start(apply_button, False, False, 6)
+        button_box.add(apply_button)
         clear_button = Gtk.Button(label=_("Clear"))
         clear_button.connect('clicked', self.clear_clicked)
-        bbox.pack_start(clear_button, False, False, 6)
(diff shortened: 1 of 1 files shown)


DateCalculator/DateCalculator.py
```diff
@@ -73,45 +79,56 @@ def build_gui(self):
         self.top = Gtk.Box(orientation=Gtk.Orientation.VERTICAL)
         self.top.set_border_width(6)
 
-        self.entry1 = self.__add_text_view(_("Expression 1"))
-        self.entry2 = self.__add_text_view(_("Expression 2"))
-        self.result = self.__add_text_view(_("Result"))
+        self.entry1 = self.__add_entry(
+            _("Expression 1: Reference Date or Date Range"), _("a valid Gramps date")
```
hgohel, line 83: Does "Expression 1" (and 2) help any more? With the proper description of the input field, what value does it add?

stevenyoungs: Two of the error messages still refer to "expression 1" or "expression 2".
If they are reworded then I agree it is clearer to remove "Expression 1:" and "Expression 2:" from the labels.

I've slightly reworded the relevant error messages and removed the label prefix.
I think the error messages could still be improved - ideas welcome!

hgohel: I see your dilemma. Ideally Gramps UI would have field validation and in-place error indication like a red border on error or a tooltip indicating the error. Without that the error message has to repeat the field description to be completely accurate. Thanks for making the change.

giotodibondone:
Updated the [wiki](https://www.gramps-project.org/wiki/index.php/Addon:DateCalculator), please give it a quick read over to verify the accuracy.

Thanks for the changes, much easier to use with the copy button.

🎅🤶❄🎄

PR #623, opened 2024-12-26
update avatars

DaveSch-gramps:
update dark, light and cartoon themes
add Updated theme
add generic male, female, other and unknown folders with symbol options

Bug #0013540


GaryGriffin:
There is an issue when building this. I get the error:

GraphView/po/uk-local.po:82: 'msgid' and 'msgstr' entries do not both begin with '\n'

Please fix before I can merge

PR #625, opened 2024-12-28
Form addon: Let form act as a drag and drop target

stevenyoungs:
Configure the form Details tab to act as a drop target, allowing a person or family object to be dropped.
The three form section types behave as follows:

- PersonSection: the dropped person is set as the person
- FamilySection: the dropped family is set as the family
- MultiSection: the dropped person is added as a new bottom row.

Note: you must drop on the headers, labels or whitespace. You cannot drop on an active edit widget

Drag in progress
image

After drop
image


GaryGriffin:
I thought RenTheRoot was rewriting the forms gramplet. Do we want to be pushing updates on the current code given this upcoming change? 

@Nick-Hall - thoughts?

Nick-Hall:
We should ask @rentheroot for her opinion.  This seems like a good idea though.

rentheroot:
I can confirm this will cause no issues with my updates to the form addon. I also think it would be a great addition!

PR #626, opened 2024-12-28
Gramps Web Sync improvements

DavidMStraub:
~Do not merge yet.~


GrampsWebSync/const.py
```diff
@@ -1,30 +1,13 @@
 """Constants for Gramps Web Sync."""
 
+from __future__ import annotations
 
-class TypeMeta(type):
-    """Workaround for missing typing module in Gramps AIO."""
```
PQYPLZXHGF, line 5: @giotodibondone removing work around for Gramp AIO about **import typing** module as per https://gramps-project.org/bugs/view.php?id=12073 in case no one answers you...

@DavidMStraub  Love the Gramps Web Sync addon and Gramps web, Thanks 🐱‍🏍💯

giotodibondone: @PQYPLZXHGF Thanks , re-closed 12073 

@DavidMStraub Great work with the addon and gramps web.

DavidMStraub:
This is ready for beta testing! Last missing piece before merging is to handle token refresh, i.e. being able to handle syncs that take longer than 15 minutes.

DavidMStraub:
Done. Will set to ready for review after adding a description of the changes in ~12 hours.

DavidMStraub:
Here comes the summary.

This is a major overhaul of the Gramps Web Sync addon with the following improvements:

- General code cleanup and improvements of type hints
- Add extensive debug logging
- More meaningful error messages
- New sync mode "merge" (does not delete any objects but restores missing objects on both side)
- If the server side has Gramps Web API v2.7+ (to be released today), leverages new background transaction handling, which allows synchronizations taking arbitrarily long, without server timeouts. A progress bar is displayed and even token refresh is handled if the process takes longer than 15 minutes
- Twp progress bars for media file sync are added (upload/download)
- Updated German translation

Overall, I think this addresses most of the pain points of the existing addon (timeouts, unclear progress status, hard to debug, partially translated) :rocket:

DavidMStraub:
@GaryGriffin FYI I also added one line to `make.py` because otherwise I got an error if `...-temp.po` is empty and the file is not created by `msggrep`.

dsblank:
Note that the ability to use the "|" operator for type hints in Python, signifying a union type, was added in Python version 3.10 as part of PEP 604. That implies that you'll need Python 3.10 or greater. Is that true for gramps core?

dsblank:
Gramps core seems to require 3.8 or greater: https://github.com/gramps-project/gramps/blob/master/setup.py#L32-L33

Nick-Hall:
> Gramps core seems to require 3.8 or greater

Gramps v6.0 core will probably require 3.9 or greater.

Our current policy is to support all versions that are not end-of-life at the time of release.

DavidMStraub:
I added `from __future__ import annotations` which makes the PEP 604 syntax (and other typing improvements) available for previous Python versions. This addon works with Python 3.9. Probably also with Python 3.8, but I don't worry about supporting EOL versions.

GaryGriffin:
Just making sure I understand your request - you want me to merge this now but not publish, since it is beta. Or are you asking that I NOT merge now while in beta?

DavidMStraub:
Yes, it can be merged! Just wanted to make you aware that I modified 1 file outside of my addon folder, but I thought it wasn't worth opening a separate PR for it.

-------- Ursprüngliche Nachricht --------
Am 31.12.24 18:25 um GaryGriffin schrieb :

DavidMStraub:
Thanks! Just to be sure: publishing to `addons` is done without me having to trigger anything, right?

GaryGriffin:
There are 2 steps here. 
1) Merge PR - anyone can do this, but recommended not by author. I often do this after the PR has been committed and there is time (I personally default at 1 week) for review/comments. 
2) Publish for Addon Manager - a subset of people can do this, currently I am doing this more often than others. This creates the tgz and json files. 

Normally, I do these steps together, so once the PR is merged, it gets published to the Addon Manager. 

However, in your case, you said you have a beta. In this case, I merged (so changes can be downloaded manually) but did NOT Publish (so all users do NOT see this update). If you intended this to be what the Addon Manager uses for all users, then I would need to Publish. Please let me know if I misunderstood your intent.

DavidMStraub:
Thanks for clarifying! Sorry, my comment was misleading. Yes, the merged version is intended for all users and it would be great if you could publish it, thanks!

GaryGriffin:
Published GrampsWebSync updates. Let me know if there are issues.

PR #611, opened 2024-10-11
Add more Ukrainian translations

Illia-M:
Continue add more Ukrainian translation for addons

PR #627, opened 2024-12-29
Add more ukrainian translations and update ukrainian related events in HistContext

Illia-M:
And fix broken translation for GraphView here https://github.com/gramps-project/addons-source/pull/611#issuecomment-2564580066


Fedik:
Thank you!

PR #628, opened 2024-12-29
Form: Sync Event and Citation dates on save

stevenyoungs:
Currently the date of the event object created by the form is set based on the form.
The date of the citation object is sometimes set.
This PR aligns the behaviour so that the citation date is always set

Currently there are two different behaviours:
1. if the form definition defined the date
  the event and citation objects had the same date as the user is not able to edit the date in the form
  therefore, they were always synced
2. if the form definition did not define a date
  the event date was set based on the form dialog. 
  the citation date was never set.

With this change, the citation date is aligned to the event date when saving the form

### DIFF
--- a/Form/editform.py
+++ b/Form/editform.py
@@ -326,7 +326,6 @@ class EditForm(ManagedWindow):
         if form_date is not None:
             date_text.set_text(displayer.display(form_date))
             self.event.set_date_object(form_date)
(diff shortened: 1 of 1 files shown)


GaryGriffin:
Will merge together with 265 after RenTheRoot comments. That way users will not have to update gramplet twice within days.

PR #629, opened 2024-12-30
Tentative fix for crash in Form editor

stevenyoungs:
Gramps 5.2, introduced a regression when closing the EditForm dialog, if an image from the gallery had been selected.
This is logged in [mantis](https://gramps-project.org/bugs/view.php?id=13326) and discussed in the forum [here](https://gramps.discourse.group/t/repeatable-unexpected-error-gallery-in-forms/6436/8) and [here](https://gramps.discourse.group/t/error-gallerytab-object-has-no-attribute-iconlist/6675/7).

**Minimal steps to reproduce**
1. install the Forms addon
2. configure at least one Source to be a form
3. create a new form
4. in the form dialog, switch to the gallery tab
5. add an image
6. select the image you just added
7. click Cancel on the Form dialog

It can also be reproduced by opening an existing form with an image and following steps 4, 6 and 7

**Investigation**
I suspect the problem is related to the order in which `EditForm.close()` is destroying objects. Namely `self.gallery_list.clean_up()` is called, leaving the GalleryTab object in an "unexpected" state, before `ManagedWindow.close(self)` is called (which in turn triggers calls into GalleryTab, resulting in the error)

**Tentative Fix**
This PR moves the call to `self.gallery_list.clean_up()` into `EditForm.clean_up`. At face value this seems like a sensible thing to do, and it fixes the error. I'd appreciate a review from someone that understands the object lifetimes of a ManagedWindow

Gramps 5.1.6 works correctly. The modified code is unchanged since 5.1.6.

### DIFF
--- a/Form/editform.py
+++ b/Form/editform.py
@@ -324,7 +324,7 @@ class EditForm(ManagedWindow):
         if not event.get_handle():
             form_reference = get_form_reference(form_id)
             if form_reference:
-                ref_entry = self.widgets['ref_entry']
+                ref_entry = self.widgets["ref_entry"]
                 ref_entry.set_text(form_reference)
 
         # Set date
--- a/Form/editform.py
+++ b/Form/editform.py
@@ -384,6 +384,9 @@ class EditForm(ManagedWindow):
         self.close()
         self.callback()
 
+    def clean_up(self):
+        self.gallery_list.clean_up()
+
     def close(self, *args):
         """
         Close the editor window.
@@ -395,7 +398,6 @@ class EditForm(ManagedWindow):
         self._config.set("interface.form-horiz-position", width)
         self._config.set("interface.form-vert-position", height)
         self._config.save()
-        self.gallery_list.clean_up()
         ManagedWindow.close(self)
 
     def help_clicked(self, obj):


hgohel:
@stevenyoungs I'm no expert on this code and this is my understanding simply by reviewing the source. It looks like [GalleryTab handles its own cleanup](https://github.com/gramps-project/gramps/blob/785c739c21584e5a4a75c833cae2bda79d6d4316/gramps/gui/editors/displaytabs/gallerytab.py#L646), so there should be no need to call it explicitly. Other places which create a GalleryTab in Gramps core code don't call clean_up explicitly either, and are hopefully not leaking resources.

One relevant change in recent years that might need a quick review is [Gramps PR#1461](https://github.com/gramps-project/gramps/commit/37395da262ef19b1702a5986b95ea1014252a888)

hgohel:
@stevenyoungs Following up on my previous comment: Python's garbage collector should clean up objects as they go out of scope so although GalleryTab has a clean_up method, I am not sure why it's necessary and in what instances it should be called explicitly. As you said, someone more knowledgeable should review.

jralls:
Python's GC only manages memory that Python allocates. [This SO](https://stackoverflow.com/questions/71167320/memory-leak-in-my-pygobject-project-but-seems-to-be-outside-of-python) shows that it's pretty easy to get PyGobject to leak like a sieve and the answer shows both a way to diagnose it and how to write code that gives PyGobject control over the memory allocated by wrapped GObjects.

stevenyoungs:
> @stevenyoungs Following up on my previous comment: Python's garbage collector should clean up objects as they go out of scope so although GalleryTab has a clean_up method, I am not sure why it's necessary and in what instances it should be called explicitly. As you said, someone more knowledgeable should review.

and conversely, if I remove the explicit call to `self.gallery_list.clean_up()`, which this PR adds, then `GalleryTab.clean_up` has not (so far) been called. It could however just be that the gc has not yet run....

stevenyoungs:
> Python's GC only manages memory that Python allocates. [This SO](https://stackoverflow.com/questions/71167320/memory-leak-in-my-pygobject-project-but-seems-to-be-outside-of-python) shows that it's pretty easy to get PyGobject to leak like a sieve and the answer shows both a way to diagnose it and how to write code that gives PyGobject control over the memory allocated by wrapped GObjects.

An interesting read - thanks for the link @jralls!
Alas, I'm on Windows :-(

stevenyoungs:
> > @stevenyoungs Following up on my previous comment: Python's garbage collector should clean up objects as they go out of scope so although GalleryTab has a clean_up method, I am not sure why it's necessary and in what instances it should be called explicitly. As you said, someone more knowledgeable should review.
> 
> and conversely, if I remove the explicit call to `self.gallery_list.clean_up()`, which this PR adds, then `GalleryTab.clean_up` has not (so far) been called. It could however just be that the gc has not yet run....

the comment in `GrampsTab.clean_up`, which is what `super(ButtonTab, self).clean_up()` ends up calling, implies that it has to be called in order for gc to work.

        """
        Remove any instance variables from scope which point to non-glade
        GTK objects so that the class can be garbage collected.
        """

hgohel:
> the comment in `GrampsTab.clean_up`, which is what `super(ButtonTab, self).clean_up()` ends up calling, implies that it has to be called in order for gc to work.

I tend to agree.

When instances of GalleryTab are created in Gramps core, they are being tracked for deletion, but I'm not sure when that happens, and whether the clean_up() method is called. Do you have an understanding of that?

GaryGriffin:
Is this PR approved to publish? The comments (to me) seem to imply that the change is correct, but before publishing, I want to confirm that GC is done correctly by this PR.

Nick-Hall:
Yes, this change is correct.

Sometimes GTK objects are not freed up so we use the `clean_up` method to set references to them to `None`.  This is all that is needed for the garbage collector to work.

stevenyoungs:
Great. With that confirmation @GaryGriffin it's ready to merge in my view. I've been running the fix locally since I created the PR with no problems seen.

PR #632, opened 2025-01-12
DateCalculator: Make the enter key trigger calculation

stevenyoungs:
Pressing the enter or keypad-enter keys in either of the two entry widgets now acts the same as clicking the Calculate button.

Typically your hand will be on the keyboard to enter the second date value. Pressing enter can be a quicker and more convenient way to trigger calculation than existing methods:
1. moving your hand back to the mouse, moving the mouse over the Calculate button and clicking it. 
2. pressing tab twice to move the input focus to the Calculate button and press enter
Both these methods continue to work

### DIFF
--- a/DateCalculator/DateCalculator.py
+++ b/DateCalculator/DateCalculator.py
@@ -3,7 +3,7 @@
 # Copyright (C) 2010       Jakim Friant
 # Copyright (c) 2015       Douglas S. Blank <doug.blank@gmail.com>
 # Copyright (c) 2020       Jan Sparreboom <jan@sparreboom.net>
-# Copyright (c) 2024       Steve Youngs <steve@youngs.cc>
+# Copyright (c) 2024-2025  Steve Youngs <steve@youngs.cc>
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


emyoulation:
I did a few tests calculations.

For some reason it is only showing Years and Months in results, **but not Days**, for some calculations.

image

Noted that the Height for the detached gramplet in the .gpr.py needs to be increased to show the buttons. I increased it to 300.

The width is too narrow for the default "Enter an expression in the entries above and click Calculate." text in the Result text box.  Perhaps that field could be 2 rows in height with text wrapping? (Increasing the number of rows would necessitate increasing the detached height.)  

Note that Expression1 and Expression2 were the previous label.  So "expression" may no longer be intuitive. Plus... 2 values have to be entered, so a singular form is incorrect.  Perhaps it could be "Enter date values in the text boxes above and click Calculate."

stevenyoungs:
@emyoulation Thanks for the feedback!

> For some reason it is only showing Years and Months in results, **but not Days**, for some calculations. seems to be for date +/- date calculations. e.g. 22 Jan 2025 -13 Jul 1980

I've looked into this one. The result of subtracting the two dates in gramps is a span (class `Span`). The default precision of a `Span`, used when converting it to a string, is 2. This means "only most two significant levels (year, month, day)". This explains the result you see.
The Span class defaults the precision to 2 and it's a hard coded value. You can however specify the precision when converting the `Span` to a string. I'll create a separate PR to change this to 3 for DateCalculator.

stevenyoungs:
The other points are covered in #636

PR #635, opened 2025-01-19
GraphView: Add configuration option for person limit

thcrt:
This PR adds an option to the GraphView chart configuration menu that lets users set or disable a limit to the number of individuals that may be shown on an 'All connected' chart.

The default value of this limit remains at 1000. It can be set to any value between 0 and 50000. When set to 0, the limit is completely disabled, and all connected individuals will be shown, no matter how many.

### Background

The GraphView addon sets a limit on the number of people displayed at a time, to avoid performance issues. The limit is currently hardcoded to 1000. It exists to avoid situations where a user, experimenting with the 'All connected' view, causes their computer to slow or freeze due to the performance impact of generating a GraphView chart of families with a high number of individual members.

Initially, the limit was a consequence of Python's recursion limit of 1000. When the GraphView addon was updated to generate charts without recursion, it was removed, but was then reinstated to mitigate said performance issues (see [0012225:0062187](https://gramps-project.org/bugs/view.php?id=12225#c62187)).

### Rationale

Keeping the limit in place makes sense as an easy way to avoid instability, but not every user will have hardware of the same capabilities. On my PC, which isn't particularly powerful, I can generate a >1.4k individual chart with ease. Given that this is just a matter of performance, and there's no actual hard limiting factor, it therefore makes sense to let users decide the limit for themselves.

### Screenshots

I haven't attached screenshots of an extremely large chart, since I didn't want to expose my own personal information and I'm not aware of an example dataset that contains >1000 connected individuals.

image
_The configuration option in the menu_

image
_The result of setting the limit to 5_

image
_A warning message shown when the number of individuals found exceeds the limit_

----


GraphView/graphview.gpr.py
```diff
@@ -3,7 +3,7 @@
     name  = _("Graph View"),
     category = ("Ancestry", _("Charts")),
     description =  _("Dynamic and interactive graph of relations"),
-    version = '1.0.139',
+    version = '1.0.140',
```
giotodibondone, line 6: Please reverse as you do not need to update this as the number is automatically incremented on release.

GraphView/po/template.pot
```diff
@@ -54,545 +54,550 @@ msgstr ""
...
 msgid "Relationship loop detected"
 msgstr ""
 
-#: GraphView/graphview.py:2652
+#: GraphView/graphview.py:2667
+#, python-brace-format
 msgid ""
-"You try to build graph containing more then 1000 persons. Not all persons "
-"will be shown in the graph."
+"The number of people in this graph exceeds the limit of {people_limit}. Not "
```
giotodibondone, line 401: template.pot is updated automatically based on the source file. So no changes needed to this file also

fxtmtrsine:
>  The result of setting the limit to 5

Screenshot shows 6 people, for programmers this makes sense 😊 but normal users well maybe adjust the result to match what they see?

fxtmtrsine:
It would also be nice if the warning message also indicated exactly how many people it tried to display as "Not all people will be shown." really does not help with guessing how high I can increase the limit, so if it stated "Limited to xxxx people out of XXXX people,  xxxxx people will not be shown." or something along those lines it would be much better than the vague message?

Would show unlimited people be better as a context menu option "Maximum people to be displayed" with two choices ?

_Maximum people to be displayed_:
* **Restrict to 1000 people**.
* _Unlimited_

GraphView/po/template.pot
thcrt: Resolved in [6586b32](https://github.com/gramps-project/addons-source/pull/635/commits/6586b323057573e31b376b2cea5f7d11cc5d5a18)

thcrt:
[f45b865](https://github.com/gramps-project/addons-source/pull/635/commits/f45b865363c8056baab3f23ffddfd2c6ddf23c66) changes the 'too many people' check to trigger on **greater than** rather than just **greater**, proving that the 2 hardest problems in programming are indeed naming things, cache invalidation and off-by-one errors :^)

It also updates the warning message to read:

> This graph would contain at least `{people_count}` people, which exceeds the limit of `{people_limit}` people. For performance reasons, at least `{people_missing}` people won't be shown. You can change this limit in the view configuration.

Note my use of _at least_. This is because the function in which we run this check, `DotSvgGenerator.add_connected()`, works by iterating over a deque of people, adding new connected people that it finds to the deque on each iteration. At the time that this check gets triggered, by the number of people already processed exceeding the limit, we know how many more people are left in the deque (i.e. people who are direct connections of people already added), but we don't know how many more connections those as-yet unprocessed people might have, that aren't yet in the deque. 

While it would be possible to restructure this slightly to run a full count, I fear this would defeat the purpose of the limit a bit, since we'd essentially be doing half the processing that the limit exists to avoid. I think it's better to just show a minimum value and return a good deal earlier.

thcrt:
> Would show unlimited people be better as a context menu option "Maximum people to be displayed" with two choices ?

Do you mean the chart configuration menu? I understand a 'context menu' as being a right-click menu, and I don't know if that would make sense to be added there.

I agree that using a value of `0` to disable the limit is a tad janky. I'd be happy to improve the UI but my experience both with Gramps addon development and with GTK is limited (read: zero).

I feel the best design would be a checkbox (to toggle whether a limit is applied) and a number picker or 'spinner' (to set the limit in question), where unchecking the checkbox greys out the number picker. I don't have any clue how to implement this, though, and from looking at how the configuration menu is currently built, it's not clear how difficult it would be to have the disabled/greyed-out status of the number picker update dynamically based on the state of the checkbox.

If anyone more experienced knows whether this would be a relatively simple addition, let me know and I'll start reading through the PyGObject docs.

fxtmtrsine:
> Do you mean the chart configuration menu? I understand a 'context menu' as being a right-click menu, and I don't know if that would make sense to be added there.

No I meant the [context menu](https://gramps-project.org/wiki/index.php/Addon:Graph_View#Context_menus) and specifically the "Empty section"  context menu positioned below the "Lines type" option (that you can use as an example and you can also see what I roughly mean in the Timeline Views addon  -  [blank section](https://gramps-project.org/wiki/index.php/Addon:TimelinePedigreeView#Blank_section) options for _Descendant Generations_  and _Ancestor Generations_ )  😊 
2025-01-21 18_25_39-Example Family Tree - Timeline pedigree - Gramps

SNoiraud:
Why are you adding this? We already have the "Generations" option in the search bar.

thcrt:
@fxtmtrsine 
> No I meant the context menu and specifically the "Empty section" context menu positioned below the "Lines type" option 

Ah, okay, thanks for clarifying :) 

In my view, it's probably better to have one canonical way & place to set the value. I think it's better suited to the configuration menu, since this is a setting that only really needs to be set once based on the capabilities of the user's hardware and their tolerance for slightly longer wait times. I don't imagine most people will want to frequently update the setting on the fly, rather, I'd expect it to be more of a 'set-and-forget' thing.

Of course, I'm not a veteran Gramps user, so if you think most people would expect and/or prefer it to be changed in a context menu, I'm happy to look into implementing it there. Just in my opinion, I don't think it's a great UI pattern to have the same thing be set in two different places.

@SNoiraud 
> Why are you adding this? We already have the "Generations" option in the search bar.

I'm not sure if you're referring to fxtmtrsine's example, or to my actual PR. If you're asking about the latter, you can see my paragraph on 'Rationale' in the [original PR description](https://github.com/gramps-project/addons-source/pull/635#issue-2797714718). :)

SNoiraud:
> I was thinking for a while that it would be nice to have straight lines with 90° angles

Did you try the orthogonal option ?

fxtmtrsine:
>  **this is the first time I'm noticing the 'Lines type' option.**  I was thinking for a while that it would be nice to have straight lines with 90° angles, but I didn't see it in the view configuration menu **and I didn't realise it was only available in the context menu.**

@SNoiraud The rest of the sentence mentions @thcrt  now knows after finding the option in the context menu to do _straight lines with 90° angles_ ( orthogonal ) .

PR #636, opened 2025-01-23
More DateCalculator Improvements

stevenyoungs:
@emyoulation made some comments on PR #632 after merge, which are addressed in this PR.
See his comment [here](https://github.com/gramps-project/addons-source/pull/632#issuecomment-2608328037)

1. Increase the precision of span results so that all values are shown in DateCalculator
By default gramps only shows the 2 most significant values from (years, months, days). For a date calculator it seems more appropriate to show all 3. Zero values are omitted from the result string
2. format the registration file, bump the version and update authors
3. Increase the default height of the gramplet so that all controls are visible when the gramplet is detached


GaryGriffin:
Since the gramps60 addon branch now exists, can you create a second PR for that branch. Normally I would just cherry-pick, but you have rewritten the GPR file and I cannot push that to the gramps60 branch since it is for the gramps52 branch.

emyoulation:
Can the whole 5.2 set be pushed to the 6.0 branch? 

Perhaps also have a dumbed-down (version blind) version of the make.py generate the old-style
addons-en CSV? Such a CSV file could be easily be imported into a spreadsheet for use as a checklist. 

It could show which need revisions for:
1) re-targeting Gramps version 
2) need a description improvement
3) need a help_url
3) need an author/maintainer

(Such a CSV list would be useful for a checklist of the built-in plug-ins too.)

Nick-Hall:
> Can the whole 5.2 set be pushed to the 6.0 branch?

The _gramps60_ branch contains everything in the _gramps52_ branch.  All addons have also been rebuilt for v6.0.

stevenyoungs:
> Since the gramps60 addon branch now exists, can you create a second PR for that branch. Normally I would just cherry-pick, but you have rewritten the GPR file and I cannot push that to the gramps60 branch since it is for the gramps52 branch.

@GaryGriffin please see PR #637

PR #638, opened 2025-02-05
Make PostgreSQL ready for Gramps 6.0

DavidMStraub:
I have not tested this yet. Will leave as draft until it's ready to merge.

### DIFF
--- a/PostgreSQL/postgresql.py
+++ b/PostgreSQL/postgresql.py
@@ -190,6 +190,23 @@ class Connection:
                               "WHERE table_name=%s;", [table])
         return self.fetchone()[0] != 0
 
+    def column_exists(self, table, column):
+        """
+        Test whether the specified SQL column exists in the specified table.
+        :param table: table name to check.
+        :type table: str
+        :param column: column name to check.
+        :type column: str
+        :returns: True if the column exists, False otherwise.
+        :rtype: bool
+        """
+        self.__cursor.execute(
+            "SELECT COUNT(*) FROM information_schema.columns "
+            "WHERE table_name = %s AND column_name = %s", 
+            (table, column)
+        )
+        return self.fetchone()[0] != 0
+
     def close(self):
         self.__connection.close()
 


PostgreSQL/postgresql.py
```diff
@@ -190,6 +190,23 @@ def table_exists(self, table):
...
+        :param table: table name to check.
+        :type table: str
+        :param column: column name to check.
+        :type column: str
+        :returns: True if the column exists, False otherwise.
+        :rtype: bool
+        """
+        self.__cursor.execute(
+            "SELECT COUNT(*) FROM information_schema.columns "
+            "WHERE table_name = %s AND column_name = %s",
```
dsblank, line 205: Does postgresql have this convention:

```suggestion
            "WHERE table_name = ? AND column_name = ?", 
```

DavidMStraub: Yes, that's what psycopg2 uses - it's like that in the rest of the addon as well.

DavidMStraub:
By the way - since `TEXT` is hard-coded in `DBAPI`, we cannot use [native JSON types](https://www.postgresql.org/docs/current/datatype-json.html), correct?

dsblank:
> By the way - since TEXT is hard-coded in DBAPI, we cannot use [native JSON types](https://www.postgresql.org/docs/current/datatype-json.html), correct?

One of the many aspects we learn as we port to different engines :smile: 

In order for Postgresql to use a different type, we'll have to have an abstraction for that in the base class. Like:

```
class DBAPI():
    json_data_field_type = "TEXT
```

and use that when we create the fields. There may be a couple of other abstractions if we have a generic SQL query method. I'll make a PR for the `json_data_field_type` in the base DBAPI class.

dsblank:
I may have misunderstood you question: I think Postgresql can use whatever type you want, as long as it acts like a string in Python.

DavidMStraub:
Actually, after reading up a bit, I realize I was making a wrong assumption - I thought the JSON type was more performant or offered more possibilities to query it, but it turns out it doesn't do much compared to TEXT, only enforcing JSON syntax, which we don't really need. So, nothing to do right now :relaxed: 

It will get more interesting/complicated if/when we start to query into the JSON structure itself, where each engine has their own syntax; like I had to do for the Sifts search library used in Gramps Web, e.g.

https://github.com/DavidMStraub/sifts/blob/main/src/sifts/core.py#L428
vs
https://github.com/DavidMStraub/sifts/blob/main/src/sifts/core.py#L573

dsblank:
> It will get more interesting/complicated if/when we start to query into the JSON structure itself

That is on the schedule for 6.1 (soon!). We should coordinate now to make sure that whatever we do, that it works across at least postgresql and sqlite.

Can you test this query in postgresql on new database structure:

```sql
SELECT json_extract(json_data, '$.primary_name.surname_list[0].surname') 
FROM person 
WHERE (json_extract(json_data, '$.gender') = 0) 
ORDER BY json_extract(json_data, '$.handle') DESC, json_extract(json_data, '$.gender');
```

That will be a good start. We'll have to check operators, but there is a test suite that I'll port to whatever approach we take.

DavidMStraub:
Sorry, but I have no capacity to check these things before the 6.0 release. Can we please not start new breaking database backend changes before 6.0 is fully released?

DavidMStraub:
`drop_column` still needs to be added

DavidMStraub:
Closed in favour of #725.

PR #647, opened 2025-02-17
fix Betawhatsnext "Seach" typo

emyoulation:


GaryGriffin:
For consistency, shouldn’t the template.pot commit in 649 be in this PR instead?

that way 647 is to address the typo completely and 649 would be translation only.

emyoulation:
Is there guidance on how this works?  The template.pot was generated automatically and is a new thing for me.

Nick-Hall:
I generated the `template.pot` in the _gramps60_ branch.  Then I aggregated the `pot` files and the ["Search Quota "](https://hosted.weblate.org/translate/gramps-project/addons/en/?checksum=a0f639d028aa0712) string is now available for translation on weblate.

PR #648, opened 2025-02-17
fix Betawhatsnext "seach" typo

emyoulation:


Nick-Hall:
Updated `template.pot` and rebased.

PR #649, opened 2025-02-17
Test of fr translation file

emyoulation:
translated fr.po with Perplexity.  Primarily to see if this is all that is necessary to make another language in the addons branch and the addons-fr.json

### DIFF
--- a/po/sk.po
+++ b/po/sk.po
@@ -11,7 +11,7 @@ msgstr ""
 "Project-Id-Version: GRAMPS 3.1.3\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2025-02-16 15:10-0800\n"
-"PO-Revision-Date: 2025-02-23 17:44+0000\n"
+"PO-Revision-Date: 2025-02-25 18:32+0000\n"
 "Last-Translator: Milan <mobrcian@hotmail.com>\n"
 "Language-Team: Slovak <https://hosted.weblate.org/projects/gramps-project/"
 "addons/sk/>\n"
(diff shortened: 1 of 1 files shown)


Nick-Hall:
We are in the process of switching to Weblate for the translation of addons.

emyoulation:
Yes. I read that. But wanted to experiment because the author of FTV has people wanting to Translate it. 

So I wanted to understand the manual process ... and see how it fights or cooperates with Weblate.

Note, you might want to take a quick peek at the source for the main FTV module. The developers uses an alternative translatable string marker to `_("text")`

emyoulation:
Yes. I read that. But wanted to experiment because the author of FTV has people wanting to Translate it. 

So I wanted to understand the manual process ... and see how it fights or cooperates with Weblate. 

Also wanted to see if `make.py` packages the translations .po's and template.pot file for the basic installs. Which seems necessary for people playing with translating locally.

Note, you might want to take a quick peek at the source for the main FTV module. The developers uses an alternative translatable string marker to `_("text")`

emyoulation:
@fxtmtsine

I understand your dislike of a translation file generated by AI.

But this is likely to be a throwaway file used to learn workflow. I have no idea if the translation will be incompatible with weblate. 

As a monoglot, the language that I come closest to understanding is French. So it drew the short straw as guinea pig. (Plus I know that there is an active community Geneanet. And they would have volunteers who would actually tell me when it was done badly. )

GaryGriffin:
@emyoulation 

I am trying to understand the value of this PR.  The Seach typo has been fixed in 6.0, and the go-forward addon translation process is weblate. Why add AI to the mix?

make.py can, with the right arguments, update template.pot .

emyoulation:
> Why add AI to the mix?

I fed the AI the template.pot and, since I cannot speak French and have not experience in the system, asked it to generate the  po file. (I did not want to expend a favor when I fully expect the experiment to fail. Probably because I will not have time to pursue it to completion.)  Perplexity explained some of what needed to added in the header.

Now I can watch the process of what happens next. 

But what about translation support before a plugin is accepted by the addons-source repository?  What if an end-user wants to contribute a po file early in the development cycle?  While the plugin is in alpha or beta (like FamilyTreeView is right now), can testers do some translations so their compatriots can test too? Then what happens to that work when the Weblate process takes over? 

And in the distribution side, what parts are delivered with each install? If my system is English only, do I also get the template.pot and the french po file?

PR #656, opened 2025-02-23
Fix typo in DNA Matches gramplet

Nick-Hall:


DNAMatches/dnamatches.py
GaryGriffin, line 1: I am not the author of this gramplet, but this does fix the typo.

Nick-Hall: OK.  Thanks.

PR #659, opened 2025-02-26
new generic avatars for the Graph View

DaveSch-gramps:
Taking a page from the new Family Tree View, I created the same generic avatar for each gender type letting the background color differentiate the person's gender.

Created dark, gray, and light avatars


DaveSch-gramps:
Feature request #0013669

PR #666, opened 2025-03-09
New Gramplet: Gram.py Script

dsblank:
This PR adds a new Gramplet (beta) that is a reimagining of SuperTools for Gramps 6.0. It can:

* create tabular data tables
* create simple analysis graphics: bar, pie, and histogram
* can work on entire data, selected data, or filtered data
* can edit primary objects

It is fast, as it builds on the JSON data dictionaries of Gramps 6.0. It has some major differences that I'll outline in discourse and/or wiki page.

### DIFF
--- a/GrampyScript/GrampyScript.gpr.py
+++ b/GrampyScript/GrampyScript.gpr.py
@@ -1,0 +1,35 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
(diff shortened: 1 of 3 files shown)


giotodibondone:
https://gramps.discourse.group/t/new-gram-py-script-for-6-0/7192

dsblank:
@GaryGriffin this is ready for inclusion whenever you like.

PR #672, opened 2025-03-13
Fix format string with unnamed arguments warnings

Nick-Hall:
Fixes [#13661](https://gramps-project.org/bugs/view.php?id=13661).

Note:  I used the string "Failure writing {target_path}: {message}" instead of "Failed writing {target_path}: {message}" because it already existed in DenominoViso.

### DIFF
--- a/D3Charts/AncestralFanChart.py
+++ b/D3Charts/AncestralFanChart.py
@@ -398,7 +398,7 @@ class AncestralFanChartReport(Report):
                 fp.write(outstr)
 
         except IOError as msg:
-            ErrorDialog(_("Failed writing %s: %s") % (self.desthtml, str(msg)),
+            ErrorDialog(_("Failure writing {target_path}: {message}").format(target_path=self.desthtml, messgae=str(msg)),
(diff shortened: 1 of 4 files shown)


D3Charts/AncestralFanChart.py
```diff
@@ -398,7 +398,7 @@ def write_report(self):
                 fp.write(outstr)
 
         except IOError as msg:
-            ErrorDialog(_("Failed writing %s: %s") % (self.desthtml, str(msg)),
+            ErrorDialog(_("Failure writing {target_path}: {message}").format(target_path=self.desthtml, messgae=str(msg)),
```
hgohel, line 401: @Nick-Hall Typo "messgae" instead of "message"? Also in line 648.

Nick-Hall: Yes, it looks like a typo.

GaryGriffin: Added PR to fix Ancestral Fan Chart and Descendant Indented Tree typo errors. Published PR.

PR #674, opened 2025-03-16
bug 13687: thumbnails module moved to gen.utils (5.1 and +)

romjeromealt:


romjeromealt:
Also a location issue for optional 'code' tag into Place attributes sequence.

romjeromealt:
> "Will cherry pick to gramps60 also."
No, this should be specific to gramps52
master and gramps60 need Gramps XML 1.7.2 version and also include two improvements

romjeromealt:
Thank you!

romjeromealt:
Gary,
There is maybe a packaging issue.
The list set on [MANIFEST](https://github.com/gramps-project/addons-source/blob/maintenance/gramps52/lxml/MANIFEST) is ignored. 'lxmlgramplet' needs these files. Only 'etreegramplet' will be supported because it only uses standard python libs, without additionnal files.

romjeromealt:
> make.py build will increment version when published.
What about when there is major update? 
versioning "x.y.z" will only generate a "z+1"...

GaryGriffin:
If you change the version number for major, I will process that as a valid change. You are right that the publish process just increments to z+1. 

It is not technically an issue if you increment version, since it will be incremented AGAIN in publishing. But I try to let devs know that this is not needed by them. This reduces the dev changes to repeated version changes.

GaryGriffin:
Looks like the change to the make.py for Weblate translations broke the MANIFEST inclusion. Will file a bug on this.

romjeromealt:
> "_include_in_listing=False was to prevent publication of an addon._"
But today, it is now possible to limit diffusion (all users, developpers, geeks, etc.)
In the past, I use the not include in the listing for limiting large diffusion for this set of addons, only oriented XML. Could be useful, but maybe not for aunt Martha or all users.

Nick, does it mean that files set into MANIFEST are ignored during packaging before there is the include_in_listing=False (attribute and value)?

romjeromealt:
Oh, I also added a custom 'setup.py' file, for experimentation around packaging build, in the past! https://github.com/gramps-project/addons-source/blob/maintenance/gramps52/lxml/'setup.py'. The first version is noted : May 14, 2015 - Initial population from SourceForge SVN...

romjeromealt:
>"_Issues
    Help buttons for both Gramplets do not go to this page._"

https://www.gramps-project.org/wiki/index.php/Addon:Lxml_Gramplet#Issues

Should I try to make a support for help button on these basic gramplets?

Nick-Hall:
> Nick, does it mean that files set into MANIFEST are ignored during packaging before there is the include_in_listing=False (attribute and value)? Should be renamed MANIFEST.in?

> Oh, I also added a custom 'setup.py' file, for experimentation around packaging build, in the past!

We still support the [MANIFEST](https://github.com/gramps-project/addons-source/blob/maintenance/gramps60/make.py#L245) file and custom [setup.py](https://github.com/gramps-project/addons-source/blob/maintenance/gramps60/make.py#L234) in the _gramps60_ branch.

The `include_in_listing=False` prevents the addon being listed, but not from being built.  It is also used in the `init` command, but I'm not sure why.

romjeromealt:
#681

GaryGriffin:
You have a typo in your MANIFEST file. It should have the folder name as a prefix for each file. Like:

lxml/xml.css

PR #675, opened 2025-03-16
lxml and ElementTree gramplets update/upgrade (1.7.1 -> 1.7.2)

romjeromealt:
* also need to backport on gramps60 branch
https://gramps-project.org/bugs/view.php?id=13687


GaryGriffin:
Please rebase to maintenance/gramps52 . I can also cherrypick to maintenance/gramps60

romjeromealt:
To rebase to maintenance/gramps52 does not make sense as this is specific to 6.0 and later.
Currently only to master, as PR https://github.com/gramps-project/gramps/pull/2021 is still pending.

GaryGriffin:
So that I understand, this PR is for gramps60 branch but cannot be applied until the change to gramps core is applied (and released?)? 

If so, should this be rebased to gramps60 once the core change is applied (or released?)? 

My understanding is that any push to master is not the right process, given how we migrate from release branches.

romjeromealt:
> So that I understand, this PR is for gramps60 branch but cannot be applied until the change to gramps core is applied (and released?)?

Sure, we could apply it before the change. But it was rather about testing. One will not be able to use it with '_example.gramps_' because of the 1.7.1 and 1.7.2 mix-up.

On gramps52, there was an unfixed issue around attributes sequence (order) into Place objects (_was optional_). No one saw the issue before, and 'example.gramps' does not contain any code attribute on Place. The thumbnails module also moved (maybe after gramps 5.1.4) to an other location. So, minor fixes for gramps52 only.

It was already fixed on master branch (before gramps60 branch creation...). And an experimental "privacy" support has been added (new feature, which might be backported to gramps52, if missing). So, to maintain two (or 3) branches is a little bit complicated. 

All other addons (except maybe those around DB and file formats) do not have problem for switching or backporting features or fixes. 

These gramplets are +/- POC for Gramps XML handling with multiple small features and experimentations. The only one limitation is maybe on Gramps XML schemas, so on update and upgrade. I did some tests for GEPS045 (which has not be merged for gramps60) on master branch. This is why the master branch was more 'tested, advanced' than gramps52 version (for gramps42, gramps51, gramps52).

romjeromealt:
With #679, gramps60 and master should be the same for this set of gramplets (etree, lxml and schema). Just gramps52 will still use 1.7.1 schema version.

PR #676, opened 2025-03-16
add help_url attribute support on two tools (associations and RelID)

romjeromealt:


GaryGriffin:
You dont need to update version in RelID. That will happen when published.

romjeromealt:
Done.
Sorry, it seems to happen during testing for a local rebuild.

GaryGriffin:
There is an error in the GPR file. The help_url is listed twice. 

python3 make.py gramps60 listing AssociationsTool 
Building listing for 'pt_PT'...
  File "/Volumes/Storage/Development/addons-source/make.py", line 927, in <module>
    code = compile(
           ^^^^^^^^
  File "AssociationsTool/associationstool.gpr.py", line 50
    help_url="Addon:Check_Associations",
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
SyntaxError: keyword argument repeated: help_url

GaryGriffin:
I reverted the Merge (hopefully that is the easiest solution).

The associations tool already has the help_url, so no change needed, I think. 

The RelID (since it is not in listing) does not and needs it added. Nick added help_url to all gramplets when initially created for 6.0 IFF they were in the listing.

PR #678, opened 2025-03-16
Some fixes and enhancements for the make file

prculley:
Been looking at the translation automation in make.
Big one, it looks like the extracted po files don't contain .pot msgid strings that should be satisfied from the Gramps main translations.  You stripped those out of the overall pot, to reduce translation work, good.
But when you did the extraction of main po files to individual addons, the Gramps strings were not included.

A smaller issue, Windows system (mine at least) doesn't include a "touch" command.
Also on Windows, you have to specify utf8 when opening a file or you get errors due to default code page.

I also added the ability to do a single file "as-needed", but may still want to enhance further with updated po etc. when the Weblate changes settle down.

I suspect that some of the other commands will fail for the 'all' when they hit the new po directory, but have not yet tested and fixed these.

### DIFF
--- a/make.py
+++ b/make.py
@@ -275,7 +275,7 @@ def strip_header(po_file):
     out_file = ""
     if not os.path.isfile(po_file):
         return out_file
-    with open(po_file, "r") as in_file:
+    with open(po_file, "r", encoding="utf-8") as in_file:
         for line in in_file:
             if not header:
                 out_file += line
@@ -289,8 +289,8 @@ def aggregate_pot():
     Aggregate the template files for all addons into a single file without
     strings that are already present in core Gramps.
     """
-    args = ["touch", "po/template.pot"]
-    call(args)
+    f = open("po/template.pot", "w")
+    f.close()
 
     args = ["xgettext", "-j", "-o", "po/template.pot"]
     args.extend(glob.glob("*/po/template.pot"))
@@ -514,7 +514,9 @@ elif command == "update":
         f'"{addon}/po/{locale}-local.po" '
     )
     # Get all of the addon strings out of the catalog:
-    system(f"touch {addon}/po/{locale}-temp.po")
+    f = open(f"{addon}/po/{locale}-temp.po", "w")
+    f.close()
+
     system(
         f"msggrep --location={addon}/* "
         f'"{addon}/po/{locale}-global.po" '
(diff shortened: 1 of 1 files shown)


make.py
```diff
@@ -1023,11 +1032,24 @@ def register(ptype, **kwargs):
...
-        print (addon)
+    check_gramps_path(command)
+    # add Gramps po to each of the base addon po creating "*_g.po" files
+    for g_po in glob.glob("*.po", root_dir=GRAMPSPATH + "/po"):
+        lang = g_po[:-3]
+        if not os.path.exists(f"po/{lang}.po"):
+            print(f"No language file for {lang} in po/ ?")
+            continue
+        system(
+            f'msgcat --use-first po/{lang}.po {GRAMPSPATH}/po/{lang}.po -o "po/{lang}_g.po"'
```
Nick-Hall, line 1043: Why do you need to do this?

The [get_addon_translator](https://github.com/gramps-project/gramps/blob/master/gramps/gen/utils/grampslocale.py#L662) method returns a translator that looks in both the `gramps` and `addon` domains for translations.

Nick-Hall:
> But when you did the extraction of main po files to individual addons, the Gramps strings were not included.

This is intentional.  The addon translators should  look in both the `gramps` and `addon` domains.

Nick-Hall:
> A smaller issue, Windows system (mine at least) doesn't include a "touch" command.

I thought that might be the case, but I saw that it was used elsewhere.  My original implementation of `strip_header` used a "sed" command, but I anticipated a problem there. :)

make.py
prculley: I have a script that reports the percentage of strings done for each addon/lang.  It was consistently showing less than 100%, which led me to investigate what was missing.  The Gramps strings.  I studied the make "update" code and realized that the new process was not including the Gramps strings, where the old process was.

I confess that I did not test to see if the addon was actually using the Gramps strings when an individual string was missing.  So I assumed that we needed to emulate the old process.  I'll run some more tests to make sure I understand all this, and then rework this PR.  When I have some more time (life still happens!).

Nick-Hall: I've added a few unit tests in PR [#2023](https://github.com/gramps-project/gramps/pull/2023) just to check that the translator was working as expected.

It actually looks in `gramps.mo` first and then uses `addon.mo` as the fallback.  Any strings from Gramps core that end up in an `addon.mo` file will be ignored.

GaryGriffin:
Are all of the questions resolved such that this is ready to merge? The list of changes seem to mostly be about replacing 'touch' and some lint changes. The specific question from Nick seems to be resolved.

Nick-Hall:
> Are all of the questions resolved such that this is ready to merge?

Yes.  I'll merge this now.

PR #681, opened 2025-03-20
[lxml and etree gramplets] Add an audience attribute and update status

romjeromealt:


GaryGriffin:
Do you really need a setup.py file? 

You need to change MANIFEST to use syntax folder/filename, like:

lxml/lxml.css

Otherwise the publish step does not find the files and ignores them.

romjeromealt:
Thanks!

PR #682, opened 2025-03-21
Fix files locations into MANIFEST, minor backported improvement and delete setup.py

romjeromealt:

PR #683, opened 2025-03-21
Fix files locations into MANIFEST, minor backported improvement and delete setup.py

romjeromealt:
was #682


GaryGriffin:
Published for 5.2 branch. Download is updated and contains files from MANIFEST. Listing is NOT updated as not set.

romjeromealt:
Thanks! Yes, it was the EXPECTED behavior... Sure, it was also possible to backport changes on attributes for these gramplets, but does it makes sense to provide them into listing now, for gramps52 branch ? There is an option which will filter available addons on plugins manager. One may set the option for only looking at new addons. So, even set as developer and experimental, someone curious might try it, just because "it is new!". To list it, like all addons, for the first time on gramps60 was maybe an overkilled security, but as wroke on the others PR, there was a switch between gramps xml version (1.7.1 to 1.7.2). So, to provide a larger support for gramps60 and to keep the "old" manual method for the previous branches, sounded like a safe solution.

GaryGriffin:
Since I stated (incorrectly) earlier that setting include_in_listing=False also prevented the tarball from being created, I wanted to be clear what actually happened. 

Personally, I agree that 5.2 should not be in listing and 6.0 should be in listing. That makes the most sense.

PR #684, opened 2025-03-21
Rename some python "documentation" scripts to textual

romjeromealt:
Some super*.py files are added to the lxml package. They are only there as documentation scripts either for dict or list addicts. To move to textual extension might be useful for marking them as a group of *.txt and not active python scripts for etree gramplets or lxml gramplet.
Also, I still wonder if they should be included into package. They are very large for a simple documentation like. As they are stored/hosted into the lxml repository, maybe either to remove them (still keep the revisions on previous branch) or find a solution to exclude them for production (package) and keep them as source for documentation.


GaryGriffin:
My initial reaction is that these should be made available on the documentation wiki and not in the package. If they are not needed to run the gramplet, then it doesnt feel like they should be in the installed package. 

Since they are doc-related, they could be in the wiki as downloads with descriptions. 

Secondary option - leave them in the repo but remove from the download. You will need to change the extension to something not *.py nor *.txt in this case, to avoid them being included in the tarball.

romjeromealt:
> "You will need to change the extension to something not *.py nor *.txt in this case, to avoid them being included in the tarball."

Mardown extension (.md) seems to be an alternative to .txt?

PR #687, opened 2025-03-22
Add WebSearch addon for Gramps 5.2

jurchello:


GaryGriffin:
Initial comments
1) Some Geneanet links are https and some are http
2) Making changes in configure: tooltip says changes are saved immediately, but they don't take effect until the next restart. This is confusing for the user.
3) The amount of printing to the console is very large. Is there a way to reduce?

jurchello:
@GaryGriffin First of all, thank you for the detailed response regarding the suggestions and recommendations for improving my application. The same response is duplicated in another pull request.
According to your recommendations, I made a major update in the code.

Regarding the suggestions:

> Some Geneanet links are https and some are http

I changed all http to https where possible. I also checked all domains in all csv files and removed those that return 404 or 522 errors. Additionally, I added the links you suggested in your recommendations.

> Making changes in configure: tooltip says changes are saved immediately, but they don't take effect until the next restart. This is confusing for the user.

I already have a similar task in my planned tasks: https://github.com/jurchello/WebSearch/issues/48 Links list (and other settings) should be updated right after settings are saved. How critical is it for the release? Can we postpone it a bit? I plan to do it, but it might significantly impact the code (I'm not sure yet).

> The amount of printing to the console is very large. Is there a way to reduce?

Yes, I forgot about this. I have already removed the unnecessary logs, leaving only errors. I tested the application, and there are no logs, including errors.

> Should the gpr help_url now point to the wiki page rather than the README.md?

Changed it to help_url="Addon:WebSearch".

> Wrong target version. In the gpr you have:

Fixed.

> Geneanet is in the de and fr list. In both it sets lang=fr. There is also a nl version for this link, which is quite different. There is no version where lang=en, which seems strange as I use this in en.

Fixed. I checked all the links for matching locales in the links and csv files. There are still a few mismatches in fr-links.csv, but it seems justified based on my understanding. If there are any errors, genealogists from those regions can correct me.

> You might want to add the GOV gazetteer (for places). A sample use would be: https://gov.genealogy.net/search/name?name=grosselfingen for the string 'grosselfingen'. Website supports en, ru, fr, de.

Added it to 4 regions with different locales.

> Wow, over 150 flags. Have you thought about other alternatives than loading 150+ images for every possible flag? Load/cache on demand (this could have issues for people concerned about privacy, requiring an upload at runtime)? Load only those flags that have an existing CSV link?

This is the only suggestion I disagree with. If you look at the size of the icons, all together they only take up 153.3kB. That’s almost nothing. I intentionally compressed them all to 16 pixels. This way, the user who adds their own CSV will immediately see their flag. Imagine that every region would separately request these flags to be added. That’s extra requests from users, extra work for me, and even more work for you to approve these pull requests.

> de google link is bad. It generated: https://www.google.de/%23hl=de&q=%22Griffin%2C%2BGary%22

Fixed.

> Can you run black and pylint before checkin?

Done. I made a lot of changes to the code, but still couldn't fix all the suggestions, I think it’s impossible. But there are definitely no critical errors, and my result is:
Your code has been rated at 9.42/10 (previous run: 9.27/10, +0.15)

PR #694, opened 2025-03-31
5.2 release

jurchello:


jurchello:
https://github.com/jurchello/WebSearch/compare/v0.43.27...v0.58.40

GaryGriffin:
Do you realize that when this is published, the version changes to 0.58.41 ? Since you are managing to version number on both your branch and gramps branches, the chance of sync problems is real.

jurchello:
@GaryGriffin Yes, I understand. And the need to synchronize does indeed cause me some difficulties. I don't plan to make many changes in the future, except for fixing some bugs or implementing user requests. Nevertheless, what workflow do you suggest? If I see that the project will require frequent updates, I will consider a different workflow that avoids the need to synchronize in different places. Are you suggesting working exclusively in the fork and closing my own project?

jurchello:
Thank you!

PR #695, opened 2025-03-31
6.0 release

jurchello:


jurchello:
https://github.com/jurchello/WebSearch/compare/v0.43.27...v0.58.40

GaryGriffin:
Do you realize that when this is published, the version changes ? Since you are managing to version number on both your branch and gramps branches, the chance of sync problems is real.

Before I merge, I just want to be sure that you understand that publishing increments version.

GaryGriffin:
@Nick-Hall This has a lot of po/<lang>-local.po files updated. Will this be okay with Weblate translation process that Addons now use? Do I need to publish differently given these po files (make.py aggregate-pot or extract-po)?

jurchello:
@GaryGriffin Yes, I understand. And the need to synchronize does indeed cause me some difficulties. I don't plan to make many changes in the future, except for fixing some bugs or implementing user requests. Nevertheless, what workflow do you suggest? If I see that the project will require frequent updates, I will consider a different workflow that avoids the need to synchronize in different places. Are you suggesting working exclusively in the fork and closing my own project?

jurchello:
> Do I need to publish differently given these po files
You can publish as you see fit. I was asked by users to update the settings, which affected the translations. And I did. Translation issues are outside my area of ​​expertise.

GaryGriffin:
I dont have a great suggestion. I suspect having two places where people can download (and are potentially out of sync) will cause user confusion on versions. 

There are 3 ways I have seen:
1) gramps/addons-source repo. Full support within Addon Manager, so easy to use.
2) You create a private repo. More work for you, but still has Addon Manager support.
3) You create a tarball for users to manually upload. Little work for you, more work for users, no Addon Manager support. 

People generally choose one or the other for released gramplets. People also use their own tarballs if the gramplet is not ready or the owner does not want to long-term support it.  

I have seen many people use tarballs and once they have user feedback and testing, they put into the addons-source repo. I have done this. This is what you have done. You need to decide which distribution is easier for you.

GaryGriffin:
Sorry - the publish question was for Nick since the commands that I have to use to publish might change due to the translation stuff. I am still learning the intricacies of it as it changed for 6.0 (using Weblate) vs 5.x . 

This is not a question about whether to publish, but the exact commands I have to use to publish given the po file changes. Think of this as behind the curtain details that I am still learning.

Nick-Hall:
> This has a lot of po/-local.po files updated. Will this be okay with Weblate translation process that Addons now use? Do I need to publish differently given these po files (make.py aggregate-pot or extract-po)?

@GaryGriffin Updating local `po` files in v6.0 is a waste of time - they will be overwritten by the Weblate translations.

The `template.pot` should be aggregated so that the Weblate translators see the new strings.  I can do this for you.

PR #696, opened 2025-03-31
Dialogs on lxml gramplet

romjeromealt:
* Update UI (_more user friendly_)

5.2 : https://gramps-project.org/wiki/images/4/47/Lxml-Gramplet-detached-addon-51.png

PR for 6.0 : https://gramps-project.org/wiki/images/3/34/Lxml-Gramplet-detached-addon-60.png

* Move the user stuff (file selection and the RUN button) to the top of the dialog, on the same horizontal line
* Add a tooltip
* Fix a console warning on a child dialog (pop-up on File Selector)
* Make ICONS more large and fix a STOCK_ICON under my config


GaryGriffin:
lxmlGramplet.py is the only file with changes, but there are a lot of commits to other files.  Does this need rebased?

Line 154: ("Please, select a Gramps XML file and\n clic on the Run button.")) . 
clic --> click
Should we say Please? I have not seen that anywhere else.

line 773: self.text or self.text_page ?

romjeromealt:
> lxmlGramplet.py is the only file with changes, but there are a lot of commits to other files. Does this need rebased?

Yes, but I was not able to find how to do that, from the Github web interface.

romjeromealt:
> Line 154: ("Please, select a Gramps XML file and\n clic on the Run button.")) .
> clic --> click
> Should we say Please? I have not seen that anywhere else.

oups, "clic" is in french... I will also remove the please.

GaryGriffin:
I am not a git wizard, but I think you need to re-clone the 6.0 branch. Then your local branch will be in sync with the git branch. The changes should then be relative to that new clone, and list only the one file.

romjeromealt:
> line 773: self.text or self.text_page ?

I did a mistake with self.text because it should be only the GtkBuffer and not a list. The first idea was to pass many information into to the TextView, but mixing string and list is not a good idea. So, that's why we can see:
```
        if isinstance(self.text, list):
            pass
        else:
            self.text.set_text(_('Parsing file...'))
```

on line 377 and on others places

self.text_page set as a list (_the mistake on naming, was self.text=[]_) into `PrintMedia()` on line 756 will be used as `LOG.info()` into `__write_gallery()` on line 790
```
                self.text_page += Html('img', src=str(thumb), mtype=str(mime))
                self.text_page += fullclear
                self.text_page += Html('a', str(description), href=str(src), target='blank', title=str(mime))
                self.text_page += fullclear

        LOG.info(self.text_page)
```
When I tried to pass data (eg., content of the JSONL) into TextBuffer, I saw that there was a mix-up (either str or list) with the TextBuffer! It was for printing the content of the gallery into HTML page (Gramps API). The logging does the job, whatever passed. Maybe just the HTML instance should not be added into the TextBuffer (self.text). 

So, to set a proper name on this piece of code was, for me, the right solution. Sure, I made some testing before this change because it sounds like something that could break the code. 

As written, it is for the Media list into the HTML instance (and page), or for the logging. The only one issue was at the end of the run, when self.text gets an HTML instance (list). So, when we run the gramplet a second time, we can get a problem when the code use : `self.text.set_text(str(...))`.

romjeromealt:
It seems that I was able to rebase by making a pull request on my side (was PR#3) between my branch and the "origine" (gramps-addons). The title and the description of the commit sounds strange, but there is only the changes on lxmlgramplet.py.
So, once rebased from my PR, I made a new PR for gramps-addons on #697

PR #697, opened 2025-03-31
Dialogs on lxml gramplet

romjeromealt:
* minor re-organisation into Gtk widgets call (Vertical and Horizontal boxes)
* polish GUI stuff
* GtkDialog mapped without a transient parent was discouraged
* Fix ICON stock and image size on file selector
was Gtk-CRITICAL **: gtk_image_set_from_gicon: assertion 'GTK_IS_IMAGE (image)' failed
* Use LARGE ICON instead of label for the run button
* Move Gtk Entry (file path) and buttons (selector and run), on the same horizontal line.
* Now, it "prints" JSONL content via the TextBuffer on gramplet. So, we can also read and copy from gramps UI. The ERROR dialogs, inherited from lxml (python), are not "children dialogs" into gramps UI manager. So, they can pop-up if the parsing failed. A proper solution should exist, but as they will raise an error anyway, we only get them during debug or when need a fix (either on code or on user side).

### DIFF
--- a/lxml/lxmlGramplet.py
+++ b/lxml/lxmlGramplet.py
@@ -144,40 +144,38 @@ class lxmlGramplet(Gramplet):
         self.entry.set_text(os.path.join(self.__base_path, self.__file_name))
 
         self.button = Gtk.Button()
-        image = Gtk.Image.new_from_icon_name(Gtk.STOCK_OPEN, Gtk.IconSize.BUTTON)
+        image = Gtk.Image.new_from_icon_name(Gtk.STOCK_FIND, 6)
         self.button.add(image)
+        #self.button.set_size_request(40, 40)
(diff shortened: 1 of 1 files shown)


romjeromealt:
* Update UI (_more user friendly_)

5.2 : https://gramps-project.org/wiki/images/4/47/Lxml-Gramplet-detached-addon-51.png

PR for 6.0 : https://gramps-project.org/wiki/images/3/34/Lxml-Gramplet-detached-addon-60.png

* Move the user stuff (file selection and the RUN button) to the top of the dialog, on the same horizontal line
* Add a tooltip
* Fix a console warning on a child dialog (pop-up on File Selector)
* Make ICONS more large and fix a STOCK_ICON under my config

PR #704, opened 2025-04-06
added font name & size settings to DescendantsLines report

n-oersch:
This adds the option to resize the text and change the font of the DescendantsLines report. The default text size is so small, that after printing the report on a bigger scale the text is very blurry
The font name in windows is the name of the `.ttf` file at `\Windows\Fonts\..`, in Linux you can check `fontconfig`.


GaryGriffin:
On Mac, you can check Font Book.

One comment - if the user enters a font name that does not exist, there is no warning that it was ignored. This might be useful.

PR #705, opened 2025-04-09
Bug 13399: change lower case names `graphview-show-id` keys

emyoulation:
since ConfigParser forces lower case names for keys; harmonize the internal labels to follow

Addresses
https://gramps-project.org/bugs/view.php?id=13399


GaryGriffin:
Do you want this also cherry picked to gramps60 branch? 

I tested this on 6.0 and the problem exists there. This change fixes it in 6.0 .

emyoulation:
Yes.  It should be suitable for both 5.2 and 6.0

Thank you

GaryGriffin:
Published for both gramps52 and gramps60

PR #709, opened 2025-04-20
[gramps52] etree gramplet - Fix whitespace issue with gunzip; refactor and polish

romjeromealt:
* Move to a more 'modern' UI
* Extend experimentations around timestamp and Event object
* Fix a mistake on whitespace handling and gunzip via shell
* Workaround on preview area, by using 6 characters for values (counters), a separator and tabs
* more testing under Mac OS and Windows OS
* Improve messages and add a tooltip
* Python 3.9 and + support


romjeromealt:
See also #707

GaryGriffin:
Still does not work on Mac (gramps5.2). I entered the filename for the Examples tree.

I got the Console Message: 
`gunzip: can't stat: Support/gramps/gramps52/plugins/lxml/etree.xml (Support/gramps/gramps52/plugins/lxml/etree.xml.gz): No such file or directory`

and the Error Dialog:
Screenshot 2025-04-21 at 7 59 42 AM

GaryGriffin:
You need to change the gunzip line to:

`os.system(f'gunzip < "{entry}" > "{filename}" ')`

The filename variable on Mac likely contains a space character.

romjeromealt:
Oh, thank you, Gary, once more.
Right, the filename path is pointing to the USER_PLUGINS directory, which may also have a whitespace character. During testing under linux, I did not switch to an alternate path. As the default was `$HOME/.gramps/gramps52/plugins` I forgot to cover this case. My bad, sorry.

PR #710, opened 2025-04-22
[6.0] Refactor, fix for lxml and etree gramplets

romjeromealt:
was PR #707


lxml/etreeGramplet.py
```diff
@@ -99,60 +94,64 @@ def epoch(t):
 
 class etreeGramplet(Gramplet):
     """
-    Gramplet for testing etree (python 2.7) and Gramps XML parsing
+    Gramplet for testing etree (Python 2.7) and Gramps XML parsing
```
giotodibondone, line 97: Is it really for Python 2.7?

romjeromealt: oups, you are right, it has been started some years ago with Python 2.7.
Now, the code is for Python 3. Note, maybe I learned more about Gramps behavior (_and python coding_) by generating this set of gramplets, than reading most documentations on the wiki!

GaryGriffin:
etreeGramplet.py still refers to issue with white space in filename. 

`line 288: Filenames issue or white space character`

Is there still an issue with a white space character in the filename?

romjeromealt:
> Is there still an issue with a white space character in the filename?

I hope there is no more issues with a white space character...
In fact, it is a global Error message on a "try/except" bloc
```
        try:
            tree = ElementTree.parse(Path(filename))
            self.parse_xml(tree, Path(filename))
        except FileNotFoundError as e:
```
One could get a `FileNotFoundError` (whatever cause), either when ElementTree will parse the copy of our Gramps XML file, or during any action on parse_xml(). On `etree gramplet`, there is only some iterations and counting. None of them should raise a `FileNotFoundError`, but I am not able to test all possible cases generating such Error. So, it is a fallback if this error occurs.

Also for consistency, I will re-use the same translated string as `lxml gramplet`.
`ErrorDialog(_('Parsing issue'), _('Cannot parse content of "%(file)s"') % {'file': filename}, parent=self.uistate.window) `?
Thanks!

giotodibondone:
Congratulations 👍 

Spotted in the configuration tab for etree that it still shows "editions"  that I [believe you changed](https://github.com/gramps-project/addons-source/pull/707#issuecomment-2799884129) at one point and is now back?

2025-04-25 09_48_34-Configure Dashboard - Dashboard - Gramps

romjeromealt:
> Spotted in the configuration tab for etree that it still shows "editions" that I [believe you changed](https://github.com/gramps-project/addons-source/pull/707#issuecomment-2799884129) at one point and is now back?

oups, I missed this one... it is the translated 'option' string.
https://github.com/gramps-project/addons-source/blob/052cb580dbc2bf06ab71ac71b4889752ff65787f/lxml/etreeGramplet.py#L198
https://github.com/gramps-project/addons-source/blob/maintenance/gramps60/lxml/etreeGramplet.py#L203
The previous one was on the GtkTextView
https://github.com/gramps-project/addons-source/blob/052cb580dbc2bf06ab71ac71b4889752ff65787f/lxml/etreeGramplet.py#L477
there is also an other one, on console only!
https://github.com/gramps-project/addons-source/blob/052cb580dbc2bf06ab71ac71b4889752ff65787f/lxml/etreeGramplet.py#L509

Note, I have an alternate branch for testing some options on this set of gramplets. I thought to only add options on `lxml gramplet`, but I will also fix the strings for `etree gramplet` on this branch too.

PR #714, opened 2025-04-23
Update WebSearch to version 0.81.73 for Gramps 6.0

jurchello:


GaryGriffin:
If the AI Provider = disabled, why do I still get the console messages:

```
⚠ The 'requests' module is missing. Install it using: `pip install requests`.
⚠ OpenAI module is missing. Install it using: `pip install openai`.
```

Shouldn't these be limited to the mistral and OpenAI being enabled?

jurchello:
you are right, it would be better if they will be logged only if AI is enabled

jurchello:
Done

PR #715, opened 2025-04-25
Update FilterRules addon for Gramps 6.0

stevenyoungs:
Simplify the code and update for gramps 6.0

### DIFF
--- a/FilterRules/ageatdeath.py
+++ b/FilterRules/ageatdeath.py
@@ -106,8 +106,7 @@ class AgeAtDeath(Rule):
 
         max_age = int(self.list[1])
 
-        for person_h in db.iter_person_handles():
-            person = db.get_person_from_handle(person_h)
+        for person in db.iter_people():
             birth_ref = person.get_birth_ref()
             death_ref = person.get_death_ref()
             if birth_ref and death_ref:
@@ -118,7 +117,7 @@ class AgeAtDeath(Rule):
                 if birth_date.is_regular() and death_date.is_regular():
                     age = death_date - birth_date
                     if cmp(age[0], max_age):
-                        self.selected_handles.add(person_h)
+                        self.selected_handles.add(person.handle)
 
     def apply_to_one(self, db: Database, person: Person) -> bool:
         """Check if the filter applies to the person."""
--- a/FilterRules/ageatdeath.py
+++ b/FilterRules/ageatdeath.py
@@ -97,6 +97,6 @@ class AgeAtDeath(Rule):
                     elif leg == "greater than" and age[0] > max_age:
                         self.ref_list.add(person_h)
 
-    def apply(self, db, person):
+    def apply_to_one(self, db, person):
         """Check if the filter applies to the person."""
         return person.handle in self.ref_list
--- a/FilterRules/ageatdeath.py
+++ b/FilterRules/ageatdeath.py
@@ -2,6 +2,7 @@
 # Gramps - a GTK+/GNOME based genealogy program
 #
 # Copyright (C) 2020    Matthias Kemmer
+# Copyright (C) 2025    Steve Youngs
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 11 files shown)


stevenyoungs:
@dsblank another filter related fix that it would be good to have you review.

stevenyoungs:
There are other rules in this addon that need similar updates to make them compatible with gramps 6.0

FilterRules/ageatdeath.py
```diff
@@ -97,6 +97,6 @@ def prepare(self, db, user):
                     elif leg == "greater than" and age[0] > max_age:
                         self.ref_list.add(person_h)
 
-    def apply(self, db, person):
+    def apply_to_one(self, db, person):
         """Check if the filter applies to the person."""
         return person.handle in self.ref_list
```
stevenyoungs, line 102: @dsblank this is the minimal change to allow the rule to work again.
Is there any advantage to be gained by converting `self.ref_list` to `self.selected_handles`, other than consistency?

dsblank: Yes, that is the minimal change. I think the name is fine... there are all kinds of names used throughout gramps-addons and gramps.

dsblank:
Thank you @stevenyoungs !

FilterRules/ageatdeath.py
dsblank: Oh, I see. You are correct, `selected_handles` is a special name. Can you update all of these?

stevenyoungs: I'll look a bit later today

stevenyoungs: > Oh, I see. You are correct, `selected_handles` is a special name. Can you update all of these?

If a rule matches only a single object (e.g. `IsActivePerson`) does it still make sense to add the handle to a property called `selected_handles`? I feel like it should allow the `Optimizer` class to kick in and do a set operation rather than calling `apply_to_one` multiple times.
My doubt comes from the core rule `HasGrampsId` which can also only match at most one object and does not use `selected_handles`. It could call `db._get_raw_person_from_id_data` and populate `selected_handles`

dsblank:
> My doubt comes from the core rule HasGrampsId which can also only match at most one object and does not use selected_handles. It could call db._get_raw_person_from_id_data and populate selected_handles

You're right that it doesn't: 

```python`
    def apply_to_one(self, db: Database, obj: PrimaryObject) -> bool:
        """
        apply the rule on the obj.
        return true if the rule passes, false otherwise.
        """
        return obj.gramps_id == self.list[0]
```

and you are correct that we could call `db._get_raw_person_from_id_data()` in stick the handle in selected_handles. We should do that. I was hesitant to change any logic in the first pass of the optimizer. But let's find ones like that (including `isActivePerson`) and optimize those sooner than later.

My hope is that in Gramps 6.1 that we have a `db.select_people()` that would be able to search and add handles for things that require a db scan. Then we will have filters that never require a Python-level sweep through all data.

stevenyoungs:
> But let's find ones like that (including `isActivePerson`) and optimize those sooner than later.

I've updated `IsActivePerson` as a proof of concept that I'm making the right changes.
It's worth doing. `IsActivePerson` execution time reduces from 1.3 seconds to 0.57 seconds on the example tree on my PC. Completely unscientific, based on 1 run!

stevenyoungs:
I was just surprised at the difference for a rule which is basically a single comparison operation i.e. as simple as you can get.
I think there is scope to further improve the `prepare` method, which remains O(n), if the logical_op is "and". In this situation, if `prepare` knew the existing partial result set it might be able to optimize the work it did i.e. given a filter which is 
rule1 and rule2
if rule2's prepare knew the result of rule1 it may be able to optimize it's own work.

Does the optmizer have short circuit logic? i.e. in the above example if rule1 returns an empty set, there is no need to evaluate rule2 (if the logical_op is "and")

stevenyoungs:
All rules have been checked, updated for compatibility with gramps 6.0 and many have also been amended to allow the filter optimizer class to exploit them.
A couple of bugs were fixed along the way.

Ready for review (again!).

dsblank:
>Does the optmizer have short circuit logic? i.e. in the above example if rule1 returns an empty set, there is no need to evaluate rule2 (if the logical_op is "and")

Yes, the optimizer will use the computed select_handles with an AND.

PR #721, opened 2025-04-30
Small fix to web sync addon for 6.0

DavidMStraub:


GaryGriffin:
You do not need to increment version. That is done automatically (x,y,z --> x,y,z+1) when published.

DavidMStraub:
Good to know, thanks!

PR #722, opened 2025-04-30
Update FilterRules2 for compatibility with Gramps 6.0

stevenyoungs:
Update rules for combability with Gramp 6.0
Apply black formatting
Add some typing


stevenyoungs:
I'm not a regular users of these rules so help with testing would be appreciated.
I've tested that all rules execute without error on the example gramps database.

PR #723, opened 2025-05-02
Fix error that prevented loading Descendant Book Report

hgohel:
Fixes [#13705](https://gramps-project.org/bugs/view.php?id=13705)

This change resolves syntax errors introduced by a prior submission which prevented the book report from loading.


GaryGriffin:
I added more issues with this report in the bug report - 
https://gramps-project.org/bugs/view.php?id=13705#c68791 

Some are Mac-specific, others are not. 

I also get crashes (reported  in bug [#13815](https://gramps-project.org/bugs/view.php?id=13815) ) which prevents more testing on Mac.

PR #727, opened 2025-05-09
Add one time retry on exception when file upload

Illia-M:
In some cases, file uploads may fail due to temporary network issues or a restart of one of the pods running Gramps Web. When this happens, the entire sync process fails, requiring a full restart. This involves refetching remote data, comparing changes, and attempting to re-upload missing files — a time-consuming process, especially with large trees (as in my case).

This PR adds a one-time retry on network-related exceptions during file upload to mitigate such temporary issues and improve reliability.


DavidMStraub (approved):
Looks good to me!

PR #728, opened 2025-05-09
Form gramplet: Fix malformed form_us.xml file

stevenyoungs:
A closing `</form>` tag is missing from the file. This causes the Form gramplet to fail to load.

Fixes bug [13820](https://gramps-project.org/bugs/view.php?id=13820)

### DIFF
--- a/Form/form.py
+++ b/Form/form.py
@@ -37,6 +37,10 @@ import xml.dom.minidom
 # ---------------------------------------------------------------
 from gramps.gen.datehandler import parser
 from gramps.gen.config import config
+from gramps.gui.dialog import ErrorDialog, WarningDialog
+import logging
+
+LOG = logging.getLogger(".FormGramplet")
 
 # ------------------------------------------------------------------------
 #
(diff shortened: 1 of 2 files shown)


Form/form.py
```diff
@@ -121,7 +125,17 @@ def __init__(self):
...
             full_path = os.path.join(os.path.dirname(__file__), file_name)
             if os.path.exists(full_path):
-                self.__load_definitions(full_path)
+                try:
+                    self.__load_definitions(full_path)
+                except Exception as e:
+                    WarningDialog(
+                        _("Failed to load Form definition file:\n%s\n") % full_path,
+                    )
+                    LOG.warning(
```
hgohel, line 134: Would logging a warning provide sufficient protection or would Gramps end up in the same situation as this bug in case of a malformed definition?

stevenyoungs: I was not 100% sure of the best approach to take, so ended up with both!
`WarningDialog` lets the user know that something is wrong and gives them the full path of the problematic file. This avoids scaring the user that their data is at risk and also gives developers the file to investigate.
`LOG.warning` gives more information (the exception message) to help developers identify the exact problem (although it was only minimally helpful for this problem)

Just catching the exception is sufficient to protect Gramps from crashing; we do not need to show or log anything - but I did not think that was the correct approach. The user would then be wondering where their form definitions had gone and may, wrongly, think that they had lost data.

I'm happy to adjust the approach based on feedback.

PR #730, opened 2025-05-14
Prerequisites checker updates

prculley:
Some cleanups (linting) and some updates for gramps6.0.x


GaryGriffin:
For Win and Mac, since they are released as bundles (ORJSON is not in the 5.2 bundle and is in the 6.0 bundle), not sure this makes sense to backport to 5.2 . 

5.2 has min python version of 3.8 whereas 6.0 has min python of 3.9 . This is another reason to not backport.

emyoulation:
But that's exactly the point of backporting the ORJSON test to 5.2 addon.

ORJSON is not in some of the Linux flavors. The ORJSON test probably needs to advise 5.2 users of those flavors to install before they try to upgrade.

Because ORJSON is needed to run Gramps 6.0, a 6.0 Gramplet that tests ORJSON seems superfluous. (Unless a version test is needed to suggest upgrading versions from a less capable version.)

PR #736, opened 2025-05-30
DescendantSpaceTree initial commit

tpoindex:
This PR adds DescendantSpaceTree addon, generating an HTML page with an interactive Space Tree of descendants.

Initial discussion at: https://gramps.discourse.group/t/descendantspacetree-addon-initial-development/


GaryGriffin:
MANIFEST should have the addon directory name first. Like:

DescendantSpaceTree/proto/*
DescendantSpaceTree/webfiles/css/*
DescendantSpaceTree/webfiles/js/*
DescendantSpaceTree/webfiles/html/*
DescendantSpaceTree/webfiles/*

GaryGriffin:
Built the download file (to verify MANIFEST) and tested. Occasionally addon crashes on Mac (Gdk error) when changing Center Person. Bug submitted.

https://gramps-project.org/bugs/view.php?id=13872

tpoindex:
> Some message typos: 
> You will need to regenerate template.pot file.

Fixed in #faab230

GaryGriffin:
Are there additional issues to resolve? If not, you can remove the Draft state for further processing. 

Is the Mac occasional crash (diagnosis by jralls) still an open issue? I have not been able to reproduce it.

tpoindex:
I think it's ready for next step, changing status to Ready for review.

I see that the Mac/GTK issue is still open, but that bug seems like it would affect any use of PersonOption().

PR #738, opened 2025-06-01
Update WebSearch to version 1.10.7 for Gramps 6.0

jurchello:


GaryGriffin:
You need to update the MANIFEST file as the ai folder is not in the download that is created.  In your case, the MANIFEST file needs the line added:

WebSearch/ai/*

GaryGriffin:
Can you add an option to not use QR codes (like you have for AI). On a Mac, I always get:

❌ Error. QR codes are disabled. Install it using: `pip install qrcode[pil]`.
❌ Error. QR codes are disabled. Install it using: `pip install qrcode[pil]`.

I would like to select/deselect *something* in the config to avoid this every time I start gramps.

GaryGriffin:
If WebSearch is installed in the lower pane, then every time I start gramps I get the 2 errors.

jurchello:
@GaryGriffin done

jurchello:
@GaryGriffin No reasons. Looks like I do something wrong but I dont know what exactly. My project still has the po directory and .gitignore doesn't ignore po. So, everything looks good.



Gitignore list:
- .idea
- __pycache__
- config.ini
- data
- locale

jurchello:
I have only **uk_UA.po** translations. But anyway, after the first or second release I should add the same UA translations to Weblate, because it didn't use my **po** file. I dont know why.

Nick-Hall:
When you add or modify a translated string, you only need to update the `template.pot` file.  This then gets aggregated with all the other `template.pot` files into the `addons.pot` file for Weblate to translate.

The local `po` files are then extracted from the Weblate `po` files when we marge the Weblate translations.

GaryGriffin:
Process is different for 5.2 and 6.0 since Weblate is doing all translations starting with Addons 6.0 . 

5.2 - uses the po/xx-local.po for translations
6.0 - publish process updates po/xx-local.po based on Weblate translated strings.

What I think has happened for 6.0:
When previous version was published, the process created the de, fi, ... string translation files (eg po/de-local.po). These are in the repo. You dont have these in your local repo. So when you pushed your latest, these were marked to delete. I think the fix is for you to get these from the repo and leave them alone in your local repo.

You should only update po/template.pot (based on new strings that will need translated) and leave all the other files in that folder alone. The publish process will update the local.po files as needed. You will need to refresh your local repo po folder from the main repo before any PR to capture the latest translation strings. 

You can use Weblate to add the uk translations (for all addons). You should not use the uk-local.po file in 6.0 - Weblate creates this.

@Nick-Hall  can correct me if I have mis-spoken.

jurchello:
Got it. I will use the described above scenario in the next releases for 6.0+

GaryGriffin:
Merged and published. Recreated xx-local.po files from Weblate translations in the process.

PR #740, opened 2025-06-06
DRAFT: add some usage info and a workspace command to make.py

ahs3:
I kept trying to invoke "make.py --help" and nothing useful happened.  So, I fixed that based on the commentary in make.py and some code reading.  This will now at least provide some usage info.

More experimental is the "make.py workspace" command that was added.  The intent of this command is to do the initial git repository setup for development of an addon, or maintaining the addons repo.  Given a directory, it will clone the necessary repositories into the proper place under that directory, checking that forks have been made in a user's github account as well.  If this is useful, excellent.  If not, it is not essential to addons-source for the long term even if it is handy for some of us lazier folks.

### DIFF
--- a/make.py
+++ b/make.py
@@ -470,6 +470,120 @@ def usage():
     print("            files that are required.")
 
 
+def init_command(command, path, target_addon):
+    # Get all of the strings from the addon and create template.po:
+    pwd = os.environ["PWD"]
+    os.chdir(path)
+    if target_addon == "all":
+        dirs = [file for file in glob.glob("*") if os.path.isdir(file)]
+    else:
+        dirs = [os.path.join(path, target_addon)]
+    if (command == "init" and len(sys.argv) == 4) or \
+       (command == "workspace" and len(sys.argv) == 5):
+        plugins = []
+        try:
+            sys.path.insert(0, GRAMPSPATH)
+            os.environ["GRAMPS_RESOURCES"] = os.path.abspath(GRAMPSPATH)
+            from gramps.gen.plug import make_environment
+        except ImportError:
+            print(
+                "Where is Gramps: '%s'? Use "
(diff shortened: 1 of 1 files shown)


make.py
```diff
@@ -340,25 +400,87 @@ def extract_po(addon):
...
-    # # Get all of the strings from the addon and create template.po:
-    if addon == "all":
+def usage():
+    """
+    what are my options?
+    """
+    print(f"usage: {sys.argv[0]} <version> <command> [<options>]")
+    print("where:")
+    print("    <version>    => Gramps maintenance branch name: one of")
+    print("                    {gramps42 | gramps50 } gramps51 | gramps52 | gramps60}" )
```
hgohel, line 410: Typo: should be `gramps50 |` as opposed to `gramps50 }`

hgohel (changes requested):
Helpful change, thanks!

emyoulation:
Prompted Claude AI to make a GUI for a subset of the `make.py` utility option. This runs within Gramps as an 'expert' 'Utilities' Tool. Perhaps we could collaborate?

"Addon Pack and Ship" creates the "Project" directory within its own addon subdirectory in the Gramps User Directory.  This can be tested as a local "Project" before posting to your GitHub online repo. 

Only successfully registered addons can be 'packed'. (There is a filter which includes searching `authors` and `authors_email` registrations. (But `Maintainers` are not searchable as the Gramps registration does not seem to store those.)
image

Claude integrated the `make.py` into the addon instead of using it as a external module. Which creates an extra maintenance hassle.

`https://raw.githubusercontent.com/emyoulation/CuratedGrampsPlugins/master/gramps60/`
 `https://raw.githubusercontent.com/emyoulation/CuratedGrampsPlugins/master/gramps52/`

PR #742, opened 2025-06-17
DescendantSpaceTree:  Added two features, fixed resize problems

tpoindex:
new feature: option to open generated file in browser
new feature: theme option
bug fix: resize and reset graph on browser resize.  clarify modal web help 
bump version to 1.1.0 for new features
regenerate template.pot

### DIFF
--- a/DescendantSpaceTree/DescendantSpaceTree.gpr.py
+++ b/DescendantSpaceTree/DescendantSpaceTree.gpr.py
@@ -38,7 +38,7 @@ register(
         "as a Space Tree for efficient viewing, even "
         "with many descendants or generations."
     ),
-    version = '1.0.1',
+    version = '1.1.0',
     gramps_target_version="6.0",
     help_url="Addon:DescendantSpaceTree",
 )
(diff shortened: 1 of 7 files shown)


GaryGriffin:
Two comments
1) You do not have to increment the version in the gpr file. That is done automatically during publish. It will increment the lowest digit (x.y.z --> x.y.z+1). If you want the 'x' or 'y' digit incremented, then you need to make that change in the gpr.
2) I should probably be clearer about the use of Draft state. If the PR is known to be incomplete, it should be in Draft. If the author thinks it could be merged, but is waiting feedback, it should not be in Draft.

tpoindex:
>     1. You do not have to increment the version in the gpr file. That is done automatically during publish. It will increment the lowest digit (x.y.z --> x.y.z+1). If you want the 'x' or 'y' digit incremented, then you need to make that change in the gpr.

Right, since I added features, I incremented the Y value.  (semver.org specs # 7), although it could be argued this isn't really an API change.  I can easily go back to x.0.y and let publish increment the patch.

>     2. I should probably be clearer about the use of Draft state. 

Got it.  I'm still relatively new to Gramps development, so erring on the cautious side.

PR #744, opened 2025-06-21
Gedcom 7 import addon

DavidMStraub:
This is a draft of an import addon for Gedcom 7 files based on the [standalone library](https://github.com/DavidMStraub/gramps-gedcom7). The implementation is not complete yet, but ready for testing & feedback.

Two problems regarding the plugin (i.e., independent of the library):

- When using `ged` as extension, the plugin is always overruled by the core Gedcom (6) plugin. That's why I had to change the extension to `ged7`, but that's an ugly hack. Is there a way around it?
- I added `gramps_gedcom7` (with underscore) to `requires_mod` because that's the module that is required, but there are two problems with this:
    - The package is called `gramps-gedcom7` (with hyphen) and there is currently no way to specify that in the `gpr` file
    - There is currently no way to specify a target version of a pip package.

To install the required package, use

```bash
python -m pip install --upgrade gramps-gedcom7
```

### DIFF
--- a/ImportGedcom7/MANIFEST
+++ b/ImportGedcom7/MANIFEST
@@ -1,0 +1,1 @@
+ImportGedcom7/README.md
--- a/ImportGedcom7/README.md
+++ b/ImportGedcom7/README.md
@@ -1,0 +1,37 @@
+# GEDCOM 7 Import
+
+Imports [GEDCOM 7](https://gedcom.io/) files into Gramps.
+
+GEDCOM 7 is a substantial revision of the GEDCOM standard and is not backwards
+compatible with GEDCOM 5.5.1, so it needs an importer of its own. Gramps' built-in
+GEDCOM importer continues to handle 5.5.1 and earlier files.
+
+## Requirements
+
+This addon requires the [gramps-gedcom7](https://github.com/DavidMStraub/gramps-gedcom7)
+Python library, which does the actual parsing and conversion:
+
+```bash
+pip install gramps-gedcom7
+```
+
+The Plugin Manager can also install it for you when you install this addon.
+
+## File extension
+
+The addon registers the `.ged7` extension. GEDCOM 7 files produced by other
+applications usually carry the same `.ged` extension as GEDCOM 5.5.1 files,
+which Gramps routes to its built-in GEDCOM 5.5.1 importer — so **rename a
+GEDCOM 7 file to `.ged7` before importing it**.
+
+GEDCOM ZIP packages (`.gdz`) are not supported yet; unpack them and import the
+GEDCOM file they contain.
+
+## Status
+
+The underlying conversion covers the vast majority of GEDCOM 7 and has been
+tested against real-world files. The addon itself is registered as beta because
+it is new; as with any conversion tool, check the results after importing.
+
+Please report problems, ideally with a sample file, at the
+[gramps-gedcom7 issue tracker](https://github.com/DavidMStraub/gramps-gedcom7/issues).
--- a/ImportGedcom7/import_gedcom7.py
+++ b/ImportGedcom7/import_gedcom7.py
@@ -1,0 +1,7 @@
+from gramps_gedcom7 import import_gedcom, ImportSettings
+
+
+def import_data(database, filename, user):
+    """Import a GEDCOM file into a Gramps database with user context."""
+    settings = ImportSettings()
+    import_gedcom(input_file=filename, db=database, settings=settings)
--- a/ImportGedcom7/import_gedcom7.py
+++ b/ImportGedcom7/import_gedcom7.py
@@ -1,7 +1,116 @@
-from gramps_gedcom7 import import_gedcom, ImportSettings
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2025 David Straub
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+"""Import a GEDCOM 7 file into Gramps."""
+
+# -------------------------------------------------------------------------
+#
+# Standard Python modules
+#
+# -------------------------------------------------------------------------
+import logging
+
+# -------------------------------------------------------------------------
+#
+# Gramps modules
+#
+# -------------------------------------------------------------------------
+from gramps.gen.const import GRAMPS_LOCALE as glocale
+from gramps.gen.errors import DbError
+from gramps.gen.utils.libformatting import ImportInfo
+
+# -------------------------------------------------------------------------
+#
+# GEDCOM 7 library
+#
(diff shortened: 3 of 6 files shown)


Call-Me-Dave:
>  MIT license 

Wrong license to be part of Gramps. But ok as third party addon!

emyoulation:
In a Discourse "Idea" and related feature request, Woody notes that GedTool (VBA  software $35) uses a "mapping table" GEDCOM dialect lookup file to map which importer to use. This differentiates between 5.5.x and 7.x and originating software.

https://gramps.discourse.group/t/new-gedcom-import-module/7045

There are other formats that could also benefit from version recognition branching. The built-in vCard importer only recognizes v3.0 (and only a limited subset) but many Smartphones only export v2

DavidMStraub:
The format version (e.g. "7.0") is required in the header in the Gedcom 7 specificiation so detecting it is trivial, the problem is just that this would require a change in Gramps core.

GaryGriffin:
I guess I am stupid, but I cannot see how to test (on Mac). I copied the ImportGedcom7 folder to my gramps60/plugins folder.  Copied your maximal70.ged (to maximal70.ged7). Then went to FamilyTrees --> Import but gramps does not seem to find the new import type (ged7).

I tried both with and without the standalone (gramps-gedcom7) library - expected error on import without it. No warnings/errors in the console, so I suspect it is not installed correctly.

DavidMStraub:
Does GEDCOM 7 show up in the "Select file type" drop down in the file dialog? (Note: I was first confused because I had to scroll that dialog to see the last, new entry.)

GaryGriffin:
No, it does not. Below is the snapshot - there is no scrolling needed to see the full list. 

Screenshot 2025-06-23 at 12 21 27 PM

DavidMStraub:
Was anyone else able to test this? Perhaps it's a Mac issue?

I noticed that this behaviour (Gedcom 7 not offered in the file dialog) happens when the library is not installed or not detected.

GaryGriffin:
I was expecting an error if the library was not installed in the line:

`from gramps_gedcom7 import import_gedcom, ImportSettings`

but didnt get any. I tried with and without the library installed - no difference. I put the gramps_gedcom7 folder in the same plugins folder (since Mac uses a bundle)

DavidMStraub:
I think the reason there is no error is that the module is listed as `mod_required`, and if that module is not importable, Gramps skips the initialization of the plugin, so it is silently absent.

GaryGriffin:
D'ohhhh

I manually installed gramps_gedcom7 not using pip (given gramps on a Mac is a bundle, any additional packages need to be added to the bundle and not to the standard OS). So I did not load gedcom7 package. 

Once I also loaded gedcom7 v0.4.0, things worked.

DavidMStraub:
Did anyone have a chance to play around? How should I deal with the extension issue, should be advise users to rename their files?

Call-Me-Dave:
> How should I deal with the extension issue, should be advise users to rename their files?

Maybe adapt the suggestion at the end this article?
* [2014-09-28 GEDCOM Version Detection](https://www.tamurajones.net/GEDCOMVersionDetection.xhtml)
* https://github.com/FamilySearch/GEDCOM/blob/main/version-detection/version-detection.md

Another way is to look at how the [GEDCOM Extensions](https://gramps-project.org/wiki/index.php/Addon:GEDCOM_Extensions) addon handles exports and see the enhancement request to refactor it to support imports?
* [10550](https://gramps-project.org/bugs/view.php?id=10550) Refactor Gramps GEDCOM import so as to support GEDCOM extension addons

The GEDCOM extension outputs the file to *.ged2 and afterwards, you can change the extension from ".ged2" to ".ged" and treat as a normal GEDCOM file.

DavidMStraub:
Detecting that a file is GEDCOM7 is easy, the problem is that currently, when it has extension `.ged`, Gramps core will use the built in parser and ignore the addon.

Call-Me-Dave:
Then Gramps core needs to be updated to support the detection or feature [10550](https://gramps-project.org/bugs/view.php?id=10550) be implemented to monkey patch the Gramps GEDCOM import detection in your addon?

Call-Me-Dave:
> [The format version (e.g. "7.0") is required in the header in the Gedcom 7 specificiation so detecting it is trivial, the problem is just that this would require a change in Gramps core.](https://github.com/gramps-project/addons-source/pull/744#issuecomment-2993616394)

Which is what you stated above  ¯\_(ツ)_/¯

PQYPLZXHGF:
Deleted the internal GEDCOM 5 importer; but probably could have hidden it via the plugin manager; which could be a suggestion to users of the addon at least during development?  

Tested using the [familysearch ged 7 example files](https://gedcom.io/tools/#example-familysearch-gedcom-70-files) and [these](https://github.com/gedcom7code/test-files/tree/main/7)  so nothing from the wild yet.

Thanks for the addon ✨

giotodibondone:
Exciting news @glamberson has submitted [pr#1 ](https://github.com/DavidMStraub/gramps-gedcom7/pull/1) that needs testing ( not sure if it works with this addon yet ) . 

> This PR adds support for three GEDCOM 7 extensions that enable full compatibility with GRAMPS' advanced data model, addressing the long-standing issue of data loss when exchanging GEDCOM files with GRAMPS


Note the PR was withdrawn see https://github.com/DavidMStraub/gramps-gedcom7/pull/1#issuecomment-3133442884

DavidMStraub:
How do we proceed? I think this could be merged if it weren't for the problem with the extension. Should we go with `ged7` for now?

GaryGriffin:
> How do we proceed? I think this could be merged if it weren't for the problem with the extension. Should we go with ged7 for now?

This gramplet is a very thin wrapper for the external library. I suspect very little maintenance will be needed in the gramplet, as opposed to updates to the external library. 

**Decision 1** options:
1. Update core gedcom import for 6.1 to add recognizing gedcom version. 
- This eliminates the issue of the extension.
2. Release separate Gedcom7 import module as 3rd party. 
- Would there be a roadmap for migrating from 3rd party to core or more likely, having it subsumed into current core gedcom import? 
- Still need the extension name resolved. And it may need changed later if subsumed into core.

**Decision 2**: how do we deal with the gramps-gedcom7 external library? If it is added to the Win/Mac bundle for 6.1, then its maintenance is tied to core releases, which is not optimal. Alternately, if it is not distributed with 6.1, then Win/Mac users cannot use it. And Linux users will need to pip install the library to access.

I think we need both decisions before going further.

Call-Me-Dave:
> Should we go with ged7 for now?

@DavidMStraub  Yes 👍 should allow people to initially use and test it.

And for the next Gramps major (6.1.x or 7.x.x) release decide what should happen?

DavidMStraub:
> I think we need both decisions before going further.

I'm open to all options and waiting for the decision to be made.

GaryGriffin:
@Nick-Hall 
Nick, since we are now talking about 6.1 release, can you give guidance on if/how core may need changes in 6.1 for gedcom7 support. Or if we keep this in an addon with a different file extension.

DavidMStraub:
If desired, we could reuse/adapt this function that I wrote to handle version detection in Gramps Web API:

https://github.com/gramps-project/gramps-web-api/blob/c63fe99745bfe06bc8caeab6de3271db2d15a21b/gramps_webapi/api/resources/util.py#L1407-L1441

dsblank:
I'll make a PR (if there isn't already one) to allow different gedcom version importers in gramps master for us to discuss. 

Also we should fix these two issues:

> I added gramps_gedcom7 (with underscore) to requires_mod because that's the module that  is required, but there are two problems with this:
> 1. The package is called gramps-gedcom7 (with hyphen) and there is currently no way to specify that in the gpr file
> 2. There is currently no way to specify a target version of a pip package.

dsblank:
PR for the two issues mentioned above: https://github.com/gramps-project/gramps/pull/2250

dsblank:
PR for the issue of routing Gedcom 7 import to the right importer: https://github.com/gramps-project/gramps/pull/2251

PR #749, opened 2025-08-02
Add better error handling to GrampWebSync

hdholm:
@DavidMStraub 
The current GrampsWebSync essentially crashes when provided a bad username or password and even offers to send a crash report.  This seems less than ideal.  I'm neither a python novice nor an expert, but somewhere in between and haven't played with gramps plugins before.  So hopefully I'm coloring within the lines here.  Full disclosure I used an AI as an experiment to see how far that would get me, but I've personally reviewed all changes and take responsibility for them. 

I used Anthropic's Claude Sonnet 4 AI to produce a first pass using the prompt:
Using the code at https://raw.githubusercontent.com/hdholm/addons-source/refs/heads/maintenance/gramps60/GrampsWebSync/grampswebsync.py as a starting point modify the code to report errors such as login failures directly on the login page in class LoginPage and provide the abilty to retry with a different password and/or user name.
commit 6e37f19fd18279595d4c1420e12cb7c640b24c24

I followed up with a prompt:
Building on that code, remove the need for the test connection button, just use the next button.  But keep the logic for the
error messages.
ommit e508f4bdbbdbf4a682e4e713f61d0b561b823be7

The rest of the commits are not AI generated and are me cleaning up the AI's misunderstanding of how the credentials were managed and it's gratuitous changes in some places as well as  some cleanup - particularly error messages that didn't necessarily reflect the cause of the error. I tried to use pylint and made a number of changes it suggested, but the vast majority were in code that wasn't added here, so I mostly left that as is.

I haven't really played with translation before and I DID change some strings, so if there are things I need to do to accommodate that please let me know.

### DIFF
--- a/GrampsWebSync/grampswebsync.py
+++ b/GrampsWebSync/grampswebsync.py
@@ -238,7 +238,6 @@ class GrampsWebSyncTool(BatchTool, ManagedWindow):
         self.assistant.set_page_title(page, title)
         self.assistant.set_page_type(page, page_type)
 
-
     def handle_done_syncing_dbs(self):
         """Handle the completion of syncing the databases."""
         self.save_timestamp()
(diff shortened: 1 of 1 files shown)


DavidMStraub:
Hi Howard,

thanks for looking in to this and providing a fix and thanks as well for being transparent about your AI use! The important thing in the end is that you are convinced you've understood all the AI generated parts, take responsibility for them, and have tested that everything works.

I didn't have time to review your changes yet, but before doing so, I want to make sure you are aware of this issue:

https://github.com/DavidMStraub/gramps-web-sync/issues/39

When you mention the addon "crashes", this is **only** due to this issue and **not** due to the issue you commented on (https://github.com/DavidMStraub/gramps-web-sync/issues/23).

Therefore, I'd also prefer first fixing issue 39 (which is a serious bug) and tackling any usability changes (which 23 is about) in a separate issue.

(Again, I haven't looked at your code yet, but if this is a separation you could make, I would really appreciate it.)

hdholm:
I don't think it's the same thing.  The default behavior for a failed password isn't a python traceback as I could have sworn it was previously, but it does show an error box and report error button which is still less than ideal.
image

Regardless, the code here provides a somewhat better display of the error and most importantly allows you go immediately go back and fix whatever was wrong rather than forcing you to click through the error dialog and completely restart the plugin.

hdholm:
It occurred to me that having shown what the existing code does, I should show what it does now.  So for the visual learners.  The next button is only grayed out until a change is made in one of the fields.
image

Nick-Hall:
Thanks for the screenshot.

I notice that you are hard-coding the error colour to be "red".  This might not be correct for all cultures or perhaps people with a visual disability.

Consider adding the error CSS class to the label instead:
```
label.get_style_context().add_class("error")
```

If this is not possible for any reason you can use the symbolic Gtk colour `@error_color`.

hdholm:
@Nick-Hall thanks for that. I think that's been corrected now.

DavidMStraub:
I finally had time to review the code, and I think it's great!

Just for completeness, can you please also handle the following statuses:

- 503 -> tree is disabled
- 429 -> too many requests, try again in a few seconds

Thanks!

hdholm:
I added handing for the 429 and 503 statuses.  Of course, that change is only minimally tested as I don't have things set up currently to be able to man in the middle and inject errors into the network traffic.

GaryGriffin:
@DavidMStraub let me know when you approve this change and I will publish it.

DavidMStraub:
Thanks, approved!

PR #756, opened 2025-08-12
[GetGOV] Fix empty date ranges

levigeiser:
For empty date ranges send an empty string to the date parser instead of using a NoneType as date.

Fixes [#13937](https://gramps-project.org/bugs/view.php?id=13937).

### DIFF
--- a/GetGOV/getgov.py
+++ b/GetGOV/getgov.py
@@ -560,4 +560,4 @@ class GetGOV(Gramplet):
         else:
             date_str = ""
 
-        return parser.parse(date_str) if date_str else None
+        return parser.parse(date_str)


hgohel:
While this change might resolve the immediate issue, the call stack for the unhandled exception in __get_sortvals doesn't handle the case where the input parameter `date` is None. Once that's fixed, the unhandled report is eliminated, and this change won't be required.

Essentially, change [L313-314](https://github.com/gramps-project/gramps/blob/8c8b0227b7089c6372c7cec3e4feb2410c1c4b80/gramps/plugins/gramplet/locations.py#L313) in locations.py to check for None:

```
        if date is None or date.is_empty():
            return (None, None)
```

@Nick-Hall @GaryGriffin You are both familiar with the code; what is your opinion?

Nick-Hall:
The fix is correct.  The date field in the place reference should contain an empty date rather than `None`.

This was unlikely to cause a problem in v5.2 since empty dates were serialised and stored in the database as `None`.  When retrieving the object the `None` would be converted into an empty date.  This no longer happens in v6.0.

We should update the Check and Repair tool to fix any date field containing `None`.

hgohel:
> The fix is correct. 

Thanks for the review and approval @Nick-Hall 

Based on the explanation of the change in date value storage and retrieval, the change I proposed in locations.py to handle a date value of None ought not to be necessary in v6.0+, but it would avoid unhandled exceptions like in this bug. Do you think the change should be made in addition to this one?

> We should update the Check and Repair tool to fix any date field containing `None`.

Would that be in Gramps master branch?

Nick-Hall:
An update to the Check and Repair tool in the _gramps60_ branch should be sufficient.  We don't know what else may be affected by this bug.

PR #758, opened 2025-08-13
Translated string change in the Edit Exif Metadata gramplet

Nick-Hall:
Update a couple of translated strings as requested by Juha Mäkeläinen on Weblate.

Also fix a couple of "is not" issues that need to be "!=".


GaryGriffin:
This also fixes https://gramps-project.org/bugs/view.php?id=13708

Nick-Hall:
Thanks.  I've added the bug reference to the commit message.

PR #761, opened 2025-08-20
Fix registration file to point to the correct help URL for Media Browser addon

hgohel:
Help URL in the registration file was pointing to a different addon's help URL. This has been fixed so that it points to the URL for Media Browser's help URL.

Fixes [#0013959](https://gramps-project.org/bugs/view.php?id=13959)

### DIFF
--- a/MediaBrowser/MediaBrowser.gpr.py
+++ b/MediaBrowser/MediaBrowser.gpr.py
@@ -38,5 +38,5 @@ register(
     gramplet="MediaBrowser",
     gramplet_title=_("Browser"),
     navtypes=["Person"],
-    help_url="Addon:MediaMerge",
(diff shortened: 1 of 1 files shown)


hgohel:
@GaryGriffin Since I don't have a working addon build environment, I haven't been able to this change, but the URL has been verified. Are you able to test?

GaryGriffin:
Confirmed that this fix works. As you stated, it does take an addon build env to test. 

BTW, your initial comment states it fixes 13959 but the link goes to 13955. Very tricky.

hgohel:
@GaryGriffin Thanks for validating the fix.

> BTW, your initial comment states it fixes 13959 but the link goes to 13955. Very tricky.

Fixed the URL in the comment. I am investigating the other defect as well; got mixed up.

PR #764, opened 2025-09-02
Fix NetworkChart addon crashing

syeysk:
Commits:
- use property name of family object instead of index

Issue: https://gramps-project.org/bugs/view.php?id=13860


hgohel:
Your first contribution to Gramps 🎉 Thanks!

Nick-Hall:
Looks good to me.  Thank you.

GaryGriffin:
Merged and Published. 
I cannot test this addon due to MacOS issues installing dependency `pygraphviz` that I have not yet conquered.

PR #765, opened 2025-09-04
Fix unhandled exception when omit duplicate option is used with index of names

hgohel:
The issue was that report_app_ref, index_of_dates, and index_of_places were only initialized when the dubperson (i.e. omit duplicates) option was enabled, but they were accessed unconditionally in methods like write_index_of_names, append_event, and write_report_ref. This lead to an `AttributeError` when later code tried to access them.

AI was asked to analyze the call stack and fix the issue. Validation and verification of fix was performed by me.

Generated-by: Cline v3.27.0; x-ai/grok-code-fast-1

Fixes #[12859](https://gramps-project.org/bugs/view.php?id=12859), #[12857](https://gramps-project.org/bugs/view.php?id=12857)

@Nick-Hall Please review acknowledgement of AI use

### DIFF
--- a/DescendantBooks/DetailedDescendantBookReport.py
+++ b/DescendantBooks/DetailedDescendantBookReport.py
@@ -353,6 +353,11 @@ class DetailedDescendantBookReport(Report):
         self.ca = CollectAscendants(self.database, self.user, self.title)
         self.ascendants = self.ca.collect_data(self.filter, self.center_person)
 
+        # Initialize attributes used for indexing and references
+        self.report_app_ref = {}
+        self.index_of_dates = {}
+        self.index_of_places = {}
+
         if self.dubperson:
             if len(self.ascendants) > 1:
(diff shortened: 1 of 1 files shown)


GaryGriffin:
@Nick-Hall Do you approve this to merge based on AI use?

hgohel:
@Nick-Hall This follows the guidelines discussed. Any concerns?

Nick-Hall:
I think that a change like this comes under the category of "Small code snippets that are non-copyrightable can be included without attribution.".  The change clearly cannot violate anyone's copyright, and is so small that identifying AI use is unlikely to be helpful in the future.

However, it is good to be transparent about AI usage, and the format of the "Generated-by:" tag looks reasonable to me.

I have no objections to this being merged.

hgohel:
@Nick-Hall Thanks for the feedback. It is helpful in clarifying AI usage attribution.

PR #770, opened 2025-09-19
Heatmap Web Report - Make it usable from CLI

soig:
Heatmap used to be called on CLI[1] but not anymore. See https://gramps-project.org/bugs/view.php?id=13990 : it's not registered to be used through the CLI.
This restores that feature

[1] See https://gramps-project.org/bugs/view.php?id=12813


soig:
(to be backported into main)

PR #777, opened 2025-12-06
Filter rule match person event with role

jjdup:
I would like to extends the event rule match person to add role selection.


GaryGriffin:
Can you add your Copyright to the py file. The existing one is from 2006 - I assume it came from whatever you used as a template.

PR #778, opened 2025-12-12
Fix PedigreeChart so it can be included in a Books Report

jfriant:
This plugin has been updated so it will work correctly when included as a "Books Report". A couple of calculations have been moved to be a part of the report generation since the Books Report might not have the menu options generated when the report is initially loaded. A little bit of code clean-up has been done to get rid of typos and misspellings as well.

This pull request is specifically targeted at the maintenance/gramps51 branch, since that is version being run by the user who requested this change, but the add-on GPR file has been updated so that it will work with any version after 5.1 and I will create a separate request for the maintenance/gramps60 to make sure it is updated there too.

### DIFF
--- a/PedigreeChart/PedigreeChart.gpr.py
+++ b/PedigreeChart/PedigreeChart.gpr.py
@@ -19,22 +19,31 @@
 #
 # $Id$
 #
-register(REPORT,
-         id="PedigreeChart",
-         name=_("Pedigree Chart"),
-         description=_("Alternate version of the traditional pedigree chart."),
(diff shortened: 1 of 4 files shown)


GaryGriffin:
Since you are doing gramps51 (and gramps60) you might want to also do gramps52, which is between those two releases.

jfriant:
The updates to this plugin have now been tested in 5.1, 5.2 and 6.0 and all three pull requests are ready for review for each of those branches.

PR #785, opened 2026-01-01
Improvements to the Pedigree Chart report

jfriant:
Add improvements from Waxtatect that adds extra styling to the person names and symbols for the events. Adds a fix for the arrow placement on the right side of the page.


GaryGriffin:
You do not need to increment the version number in the gpr file. This is done automatically as part of the publish phase.

jfriant:
Good to know about the version number in the GPR file.  I have set it back to the previous version number.

PR #786, opened 2026-01-01
Improvements to the Pedigree Chart report for 5.2

jfriant:
Mirrors PR #785 to add styling to the names. Add symbols for the events. Fix the arrow placement.


GaryGriffin:
You do not need to increment the version number in the gpr file. This is done automatically as part of the publish phase.

PR #789, opened 2026-01-07
Detailed Descendant Report with Images - Bug 14078

GaryGriffin:


GaryGriffin:
Fixes https://gramps-project.org/bugs/view.php?id=14078

PR #793, opened 2026-01-31
Virtual Keyboard v1.1.0 extended language support

emyoulation:
Layout sets for French and German

image

image

image


GaryGriffin:
Should you use unicode instead of the flag as a text string?

emyoulation:
The AI helping this wish-coder advised that Unicode was more likely to break. So I'm trying this as an experiment limited to 5.2 Gramps. Which should help keep complaints down to a dull roar if that was wrong.

Did a Black pass this time.

Thanks for the help!

PR #797, opened 2026-02-08
Virtual Keyboard v1.2.0 : External keymaps

emyoulation:
image

### DIFF
--- a/VirtualKeyboard/MANIFEST
+++ b/VirtualKeyboard/MANIFEST
@@ -1,0 +1,2 @@
+VirtualKeyboard/README.md
+VirtualKeyboard/layouts/*
--- a/VirtualKeyboard/README.md
+++ b/VirtualKeyboard/README.md
@@ -4,3 +4,42 @@ Available for docking in all views except: Relationships, Charts, and Geography
 <img width="1100" height="404" alt="image" src="https://github.com/user-attachments/assets/d0d7bcee-8e60-45e0-b558-2d5f37af2442" />
 
 <img width="754" height="404" alt="image" src="https://github.com/user-attachments/assets/659e9300-7c58-4e69-926d-122a1f4a66c8" />
+
+1.0.0
+- Virtual on-screen keyboard Gramplet for Gramps
+- Standard QWERTY keyboard layout
+- Basic editing controls (Backspace, Tab, Space, Newline, Clear)
+- OS Clipboard copy support
+- Dockable Gramplet with floating window support
+- Compatible Categories: Dashboard and all except: Charts and Geography
+
+1.1.0
+- Add multi-language keyboard sets (FR, DE added to GB) 
+- flag-based layout sets switching as "radio buttons"
+- add Tab character key
+- Introduce AltGr and Shift-AltGr layouts for supported layouts
+- Improve layout switching UI using toggle buttons
+- Expand Special character coverage for international text entry
+- Internal reorganization of keyboard layout definitions
+- GUI marked for Weblate Translation
+
+1.1.2
+- Suppress unintended Gramps tooltips over virtual keyboard controls
+- Refactor keyboard layout definitions in preparation for external layout files
+- Add Composed / Dead Keys layout for diacritics and extended characters
+- Improve internal validation and error handling for layout definitions
+
+1.2.0
+- external CSV for addon Keyboard mapping and layout sets definitions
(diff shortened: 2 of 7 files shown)


GaryGriffin:
Why add language keyboards? MacOS (and I think Win and Lin) have an Accessibility keyboard built into the OS that allows users to switch language and get a virtual keyboard of the specific language. This is much simpler than configuring the txt files needed for this addon, IMO.

And you dont need to use the Copy button and then the Paste operation to enter the string. Using the Accessibility keyboard enters the key directly into the edit field (not a buffer).

The RU keyboard in the OtherKeyboardLayouts.txt does not match the mapping in the RU keyboard in the MacOS Accessibility keyboard. This potentially adds a layer of confusion.

emyoulation:
Third item first, "OtherKeyboardLayouts.txt" is simply a scratchpad that I am sharing with users about future possibilities that have not yet been researched. To hopefully inspire user contributions. Having erroneous information about the layout is more like to prompt a response than matches.  

This VK gramplet is 1 piece of an evolving ESOL sample code suite. The [LocalTerm](https://github.com/emyoulation/LocalTerm/blob/master/README.md#localterm-gramplet) Glossary translator (which will soon interface with the Weblate API) is another. The  [Locale Checker](https://github.com/emyoulation/LocalTerm/blob/master/README.md#locale-checker-gramplet) is a yet another. It is to explore Gramps methods for determining installed languages and adapting contextually. Still have to determine which side of the Developer/Everyone audience they lean towards.

First item last, the Accessibility keyboards are a pure misery when trying to transcribe a foreign language that you don't speak.  It keeps popping up whenever you click a text box.  Getting in the way, shifting around windows onscreen all by itself. If you want to work in 3 languages at the same time, that is simply impossible.  But with Gramplets, that is no problem.  (A future expansion will allow mixed entry from both hardware and virtual keyboards.

There is a strategy behind creating these plugins.

GaryGriffin:
Questions:
1) No need to increment version from 1.1.1 to 1.1.2 in the gpr. This will be done when publishing.
2) Tried to add new keyboard but failed with no warning. Copied fr.csv to xx.csv and changed as follows. The layouts are all the same, only the set is changed.

```
@set
@id,XX_AZERTY
@language,XX
@language_iso3,xxx
@region,XX
@button,🇫 🇲
@label_native,Français
@label_en,xxFrench
@tooltip_native,Clavier français AZERTY
@tooltip_en,French AZERTY keyboard
@default_layout,AZERTY
```
3) Adding flag is awkward as symbol. Again suggest unicode which is more standard. Or at least document how to use unicode for flag.
4) Maybe adding example of transforming fr.csv to another language in HTML file would be useful?

emyoulation:
1) I'm manually incrementing the `.gpr.py` version to prevent Addon Manager from 'updating' and overwriting my development test code. It means automatic version incrementing causes a bit of hopscotch in the versioning of the public release numbers. But I don't seem any actual harm in that, merely a bit of confusion.

2) the next release should be better for building a non-french external CSV. (See №4) However, I have a teleconference about Weblate on Tuesday. So LocalTerm has priority in the next AI token allocation.

3) I am not wedded to the flags. And you will be able to use text labels too. It just seemed like the potential for dozens of layouts meant a small (or single-character) button was desireable. The label in the CSV will persist in being non-functional until I build a menuing option for the CSVs. Uincode is a luxury refinement that I just haven't had time (or spare AI tokens) to explore. 

4) It is just a bit early to write CSV files. So instructions will be changing as I simplify the workflow. The external files barely working with the superseding logic before running out of free AI tokens. When the tokens replenish, I still have more work: to change the module so the hard-coded and CSV format are copy/paste interchangeable and then then add a menuing feature allowing selection of external CSVs. 
Until then, we're limited to the French as a testbed sample. Just add your key mapping layout and add a layout selecting button to its layout sets. (The French set was implemented to test the 'superseding' functionality. As Patrice Legoux was the only end-user who had commented on Discourse. The En and De sets could be made into sample CSVs with just a little more non-intuitive effort. But they would have to be reworked when the internal and external formats are harmonized.) The only reason I did a PR was because the previous release needed additional symbols for entering a Polish name. I did not do a PR for added mapping because my prototype splitting the CSV out was at an "error generating" stage ... but was close. I don't want to bury you with constant PRs and `addons-source` republishing updates.

PR #800, opened 2026-02-16
GrampsWebSync: Add Cloudflare Zero Trust service token support

EnormousAxe:
## Summary

- Adds optional `CF-Access-Client-Id` and `CF-Access-Client-Secret` header support to `WebApiHandler` so users behind Cloudflare Zero Trust can sync with their Gramps Web instance
- Two new optional fields on the login page ("Cloudflare Access") are persisted across sessions via the existing config manager
- All HTTP requests route through a new `_headers()` method that merges any extra headers (e.g. Cloudflare Access) into each request

## Cloudflare setup

1. In the Cloudflare One dashboard, go to **Access controls > Service credentials > Service Tokens**
2. Select **Create Service Token**, give it a name, and choose a duration. Copy the Client Secret immediately -- Cloudflare only displays it once
3. In the Access policy for your Gramps Web application, add a rule with the **Service Auth** action so the token is accepted without an identity provider login
4. Enter the Client ID and Client Secret in the Gramps Web Sync login page under "Cloudflare Access (optional)"

## Test plan

- [x] Verify sync works without CF fields populated (no regression)
- [x] Verify CF credentials are saved and restored across sessions
- [x] Verify sync works with a Gramps Web instance behind Cloudflare Zero Trust using a valid service token

🤖 Generated with [Claude Code](https://claude.com/claude-code) Opus 4.6 with the prompt 

"Add Cloudflare Zero Trust service token support to GrampsWebSync. Add optional CF-Access-Client-Id and CF-Access-Client-Secret fields to the login page, persist them across sessions using the existing config manager, and include them as headers on all HTTP requests made by WebApiHandler."

### DIFF
--- a/GrampsWebSync/grampswebsync.py
+++ b/GrampsWebSync/grampswebsync.py
@@ -111,6 +111,8 @@ class GrampsWebSyncTool(BatchTool, ManagedWindow):
         self.config.register("credentials.url", "")
         self.config.register("credentials.username", "")
         self.config.register("credentials.timestamp", 0)
+        self.config.register("credentials.cf_client_id", "")
+        self.config.register("credentials.cf_client_secret", "")
         self.config.load()
 
         self.assistant = Gtk.Assistant()
@@ -128,11 +130,15 @@ class GrampsWebSyncTool(BatchTool, ManagedWindow):
         self.url = self.config.get("credentials.url")
         self.username = self.config.get("credentials.username")
         self.password = self.get_password()
+        self.cf_client_id = self.config.get("credentials.cf_client_id")
+        self.cf_client_secret = self.config.get("credentials.cf_client_secret")
         self.loginpage = LoginPage(
             self.assistant,
             url=self.url,
             username=self.username,
             password=self.password,
+            cf_client_id=self.cf_client_id,
+            cf_client_secret=self.cf_client_secret,
         )
         self.add_page(self.loginpage, Gtk.AssistantPageType.CONTENT, _("Login"))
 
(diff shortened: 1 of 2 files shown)


GaryGriffin:
Questions:
1) @DavidMStraub do you approve?
2) Previously the UserAgent: GrampsWebSync was added to address CloudFare issues. Should that be reverted? Or are both needed?
3) Please update to comply with [Howto Contribute to Gramps](https://gramps-project.org/wiki/index.php/Howto:_Contribute_to_Gramps#AI_generated_code) re: AI generated code.

DavidMStraub:
No, I don't think we should add specific configuration options and GUI elements just to accomodate specifics of a cloud provider. Where does it end? I would reject it.

EnormousAxe:
Updated the commit comment with the Claude version and prompt.

Understood on the specific use case limitation, I could generalize it for custom headers that would serve a lot more cases if it would be worth it.

Examples of use case.
  - Cloudflare Zero Trust — CF-Access-Client-Id / CF-Access-Client-Secret
  - Authelia / Authentik — Remote-User, Remote-Groups, or custom auth headers
  - Tailscale Serve / Funnel — custom identity headers
  - Traefik ForwardAuth — forwarded auth headers
  - Nginx auth_request — arbitrary auth headers passed upstream
  - API gateways (Kong, AWS API Gateway) — x-api-key or custom tokens
  - Any reverse proxy that injects or requires custom headers for authentication

DavidMStraub:
I disagree. This is not the right place to handle more complicated authentication schemes. The next step for this addon would be to use browser-based authentication with a callback, which would allow using OIDC authentication, but this has nothing to do with the changes proposed here.

This PR should be closed in my opinion.

EnormousAxe:
I appreciate the feedback, I'll look into those options. Request closed.

PR #801, opened 2026-02-16
Add DarkModePrefs addon for Gramps 6.0

slashmad:
## Summary
- add new `DarkModePrefs` GENERAL addon for Gramps 6.0
- provide a Preferences panel with `Auto`, `Dark`, and `Light` modes
- apply GTK dark preference and optional theme-name + CSS compatibility fixes
- include Flatpak-focused handling for `GTK_THEME` override visibility and behavior

## Included files
- `DarkModePrefs/darkmode.gpr.py`
- `DarkModePrefs/darkmode.py`
- `DarkModePrefs/darkmode_load.py`
- `DarkModePrefs/README.md`

## Validation
- syntax check: `python3 -m py_compile` on all addon Python files
- runtime smoke test performed in Fedora Workstation Flatpak (`org.gramps_project.Gramps`) with dark mode toggle working after removing conflicting `GTK_THEME` override

### DIFF
--- a/DarkModePrefs/README.md
+++ b/DarkModePrefs/README.md
@@ -1,0 +1,62 @@
+# DarkModePrefs (Gramps 6.0 addon)
+
+Addon for Gramps desktop that adds a **Dark Mode** preferences panel with:
+
+- `Auto` (follow GNOME system dark mode)
+- `Dark` (force dark)
+- `Light` (force light)
+- Optional GTK theme-name switching (Flatpak fallback uses `Adwaita` + dark flag)
+- Optional CSS compatibility fixes
+
+## Install (local plugin folder)
+
+Quick install:
+```bash
+./install-local.sh
+```
+
+Manual install:
+
+1. Create plugin folder:
+```bash
+mkdir -p ~/.local/share/gramps/gramps60/plugins/DarkModePrefs
+```
+2. Copy addon files:
+```bash
+cp darkmode.gpr.py darkmode_load.py darkmode.py ~/.local/share/gramps/gramps60/plugins/DarkModePrefs/
+```
+3. Restart Gramps.
+4. Open `Edit -> Preferences` and use the `Dark Mode` tab.
+
+## Fedora/GNOME notes
+
+- If `GTK_THEME` is set globally/shell-wide, Gramps may ignore theme changes from addons.
+- In Flatpak, `GTK_THEME` is often set via override; unset it with:
+  `flatpak override --user --unset-env=GTK_THEME org.gramps_project.Gramps`
+- In `Auto` mode, the addon follows `org.gnome.desktop.interface color-scheme`.
+- If Gramps is installed as Flatpak, host themes might not be visible in sandbox unless matching runtime theme extensions are installed.
+- If `Theme preferences` is also installed, it may conflict because both addons patch the same Preferences class.
+
+## Packaging as `.addon.tgz`
+
+From this folder:
+```bash
+tar -czf DarkModePrefs.addon.tgz darkmode.gpr.py darkmode_load.py darkmode.py README.md install-local.sh
+```
+
+Import this archive via Gramps plugin manager if you prefer archive-based install.
+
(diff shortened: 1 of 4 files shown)


GaryGriffin:
Questions:
1) We have a [Theme Preferences](https://gramps-project.org/wiki/index.php/Addon:ThemePreferences) addon. Doesn't that cover this already? 
2) If this does move forward, you need to update to comply with [Howto Contribute to Gramps](https://gramps-project.org/wiki/index.php/Howto:_Contribute_to_Gramps): General and AI generated code sections. 
3) Should also run _black_ on the python files.

slashmad:
No I couldn't get it to work on latest Fedora Workstation (wayland, gnome) hence why I created this.
I will read up on that.

GaryGriffin:
Would generating a PR to fix the Theme addon be simpler than a new addon? There is an existing user base of Theme addon - which will be less confusing than having 2 addons with the same purpose. 

@emyoulation : would appreciate your thoughts on this.

slashmad:
Yeah, would probably be better to focus on the existing addon - I'll check it out. 
In the meantime I'll use this locally.

emyoulation:
Although I've occasionally used dark modes using Themes addon on Gramps 6.0.6 using Windows 10, and frequently on Gramps 5.2 using Fedora Linux 37 (Workstation Edition) Gnome 43.9 and Wayland.  I have not yet used Gramps 6 on Fedora. So cannot give firsthand feedback. 

Also, @slashmad did not specify versions of Fedora and Gnome.  It might make a difference.

slashmad:
@emyoulation, I'm using Fedora Linux 43 (Workstation Edition) and GNOME 49.3 with Gramps 6.0.6.

slashmad:
You can check out this one, works locally.
https://github.com/slashmad/addons-source/commit/25ff466d

emyoulation:
Interesting.  Perhaps you could ask OpenAI Codex (provider: OpenAI; model/release: GPT-5, 2026-02-16) to detail exactly what was incompatible in the Themes addon which needed to be resolved?  It will be easier to address 1 item at a time to get Serge to validate.

slashmad:
Hey I'm just happy I'm not burning my eyes out (hate light themes). 
Obviously I'm in over my head but I was thinking that others may benefit from this since the theme addon didn't work  for me even if I forced a dark theme via terminal.

Here you have a lovely letter from the one and only:

"The original Theme Preferences addon is not API-incompatible with Gramps 6.0, but it is functionally insufficient for Fedora GNOME/Flatpak dark-mode behavior.

Exact gaps in the original addon:

1. No system “Auto” mode.
   - It only provides a manual “Dark Variant” toggle.
   - Code path: `Themes/themes.py` (`self.dark = Gtk.CheckButton(...)`, `dark_variant_changed`).

2. No GNOME color-scheme integration.
   - It does not read `org.gnome.desktop.interface color-scheme`.
   - It does not listen for `changed::color-scheme`, so it cannot react to live system theme changes.

3. No separate dark/light theme profiles.
   - It stores one theme key (`preferences.theme`) and one dark-variant bool.
   - There is no dark-theme + light-theme mapping with automatic switching.

4. GTK_THEME override handling is blocking but not actionable.
   - When `GTK_THEME` is set, controls are disabled with a generic tooltip.
   - In Flatpak this is common, and the original addon does not provide Flatpak-specific remediation guidance.

5. Flatpak desktop-preference edge case is not handled.
   - In this environment, host GNOME reports `prefer-dark`, but inside Flatpak `color-scheme` can resolve differently.
   - Original addon has no portal-based fallback (`org.freedesktop.portal.Settings`) to resolve host appearance preference robustly.

6. No compatibility CSS fallback for incomplete theme CSS.
   - This can leave some text/entry widgets with poor contrast in mixed theme/runtime situations.

7. Preference-panel monkey-patch conflict risk.
   - It monkey-patches `GrampsPreferences.__init__`; if another addon does the same, behavior depends on load order.
"

PR #803, opened 2026-03-02
Purpose of a new report : a table of ancestors

bforest76:
Purpose of a new report : a table of ancestors

This report shows the ancestors in a table of 4 columns
- the Sosa number
- the name of the ancestor
- the date and the place of the birth
- the date and the place of the death
Between 2 married persons the report also shows a row with the date and the place of the marriage.
A blank row separates the families to make the reading easier.

The report contains a few parameters like :
- the number of generations to display
- whether to have a page break between generations
- add a page with the number of ancestors per generation
- mask the name of the calendar in the dates 
- and the usual parameters about how to display names, dates, places...

The stylesheet purposes different styles between the males and the females, between the paternal branch and the maternal one...

### DIFF
--- a/AncestryTableReport/AncestryTableReport.gpr.py
+++ b/AncestryTableReport/AncestryTableReport.gpr.py
@@ -1,0 +1,16 @@
+register(REPORT,
+    id   = 'AncestryTable',
+    name = _('Ancestry Table'),
+    description = _("Produces a table of ancestry about a person"),
+    version = '1.0.0',
+    gramps_target_version = "6.0",
+    status = STABLE,
+    fname = 'AncestryTableReport.py',
+    authors = ["Bruno Forestier"],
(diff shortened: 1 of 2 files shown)


GaryGriffin:
Thanks for the new report. It looks quite useful.

It is best if you create the PR but not merge it. That allows people to test and provide feedback before publishing. And avoids the merged but not published-for-general-distribution possibility.

bforest76:
Gary,
Thanks for your advice, I will pay attention for the next time.
I use Github for the second time. The first one was many years ago. I don't
know the best practices. I'm learning as I'm experimenting.
Feel free if you want rollback some of my actions to do it in the right way.
I still have another problem. I made 2 commits on my computer but only one
was pushed. I'm looking forward...
Best regards.

PR #814, opened 2026-04-03
Add CensusCheck and CensusCheckUp quickviews to Forms addon

kulath:
Updated the CensusCheckQuickview (which checks descendants) and added a CensusCheckUpQuickview which checks ancestors for Gramps 6. You need to be in the person view to see the addons.

The quickview modules (and the .gpr.py) must be in the same folder as the Forms plugin, because they are designed to work with the Forms plugin (and they import the form.py).

The plugins intentionally only look at one level of descendants or ancestors, because they are intended to be used to guide research for searching and completing census for people. If it displayed everyone in the database it would be too much of a challenge to work out where to begin.

Fixes [Issue #0010388: [Forms addon] Add Census Listing report quickview](https://gramps-project.org/bugs/view.php?id=10388)

### DIFF
--- a/Form/CensusCheckQuickview.gpr.py
+++ b/Form/CensusCheckQuickview.gpr.py
@@ -16,6 +16,7 @@ register(QUICKREPORT,
          authors_email = ["guy.linton@gmail.com"],
          category = CATEGORY_QR_PERSON,
          runfunc = 'run',
+         help_url = "Addon:Census_Check"
   )
 register(QUICKREPORT, 
          id    = 'censuscheckupquickview',
@@ -29,4 +30,5 @@ register(QUICKREPORT,
          authors_email = ["guy.linton@gmail.com"],
          category = CATEGORY_QR_PERSON,
          runfunc = 'run',
+         help_url = "Addon:Census_Check"
   )
--- a/Form/CensusCheckQuickview.gpr.py
+++ b/Form/CensusCheckQuickview.gpr.py
@@ -1,0 +1,32 @@
+#------------------------------------------------------------------------
+#
+# Register the report
+#
+#------------------------------------------------------------------------
+
+register(QUICKREPORT, 
+         id    = 'censuscheckquickview',
+         name  = _("CensusCheck"),
+         description= _("Check whether any Census events are missing for a person and some of their descendents"),
+         version = '1.0.0',
+         gramps_target_version = '6.0',
+         status = STABLE,
+         fname = 'CensusCheckQuickview.py',
+         authors = ["Tim Lyons"],
+         authors_email = ["guy.linton@gmail.com"],
+         category = CATEGORY_QR_PERSON,
+         runfunc = 'run',
+  )
+register(QUICKREPORT, 
+         id    = 'censuscheckupquickview',
+         name  = _("CensusCheckUp"),
+         description= _("Check whether any Census events are missing for a person and some of their ancestors"),
+         version = '1.0.0',
+         gramps_target_version = '6.0',
+         status = STABLE,
+         fname = 'CensusCheckUpQuickview.py',
+         authors = ["Tim Lyons"],
+         authors_email = ["guy.linton@gmail.com"],
+         category = CATEGORY_QR_PERSON,
+         runfunc = 'run',
+  )
--- a/Form/CensusCheckQuickview.py
+++ b/Form/CensusCheckQuickview.py
@@ -1,0 +1,155 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2016, 2026       Tim G L Lyons
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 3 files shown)


kulath:
I have set the version in the grp file to 1.0.0, I don't know what it should be set to. Should it be the same as the main formgramplet.gpr.py file. In any event, the formgramplet.gpr.py file probably needs to have its version incremented so that this addition is caught as an updated gramplet.

kulath:
I previously got Claude to review this code:


Bugs / Logic Issues

1. probably_alive called with wrong arguments — In process_person, the "closed record" check calls probably_alive(person, database, max_age_prob_alive=100) without passing a reference year. This will use the current date rather than the census year, producing incorrect results. It should likely be probably_alive(person, database, census_list[key], max_age_prob_alive=100).
2. stab.set_link_col(4) is hardcoded — Column 4 is assumed to be the first census column, but this is only true if there are at least 4 columns (Name, Birth date, Death date + at least 1 census). If there are no census records, this will silently point to a nonexistent column. It should be stab.set_link_col(0) (linking to the person's name) or derived dynamically.
3. Children may be missed — The code collects children of person and children of each spouse's families, but it doesn't collect children from families where person is a parent alongside each spouse. If a blended family exists, some children could be skipped or duplicated inconsistently.


Code Quality Issues

4. Variable shadowing — In run(), the parameter is named person, but later for child in sdb.children(...) reassigns a loop variable child, and more critically person is never rebound — but the naming creates confusion when spouse and person are compared by .handle. This is fragile; using subject_person or target for the parameter would be clearer.
5. found_census is a list but used like a set — found_census is built as a list and checked with if key in found_census. Since duplicate census entries aren't meaningful here, this should be a set for both correctness (deduplication) and efficiency.
6. census_result built as a tuple via concatenation — Repeatedly doing census_result += ("value",) inside a loop is inefficient. Build a list and convert at the end: census_result = []; census_result.append(...); tuple(census_result).
7. form module imported without error handling — The try/except block handles the translator, but import form has no fallback. If the form addon isn't installed, the plugin crashes with an unhelpful ImportError. A guard with a user-facing message would be more robust.


Minor / Style

8. The commented-out imports at the top (# from branches.gramps40...) should be removed — they appear to be leftover debugging artifacts.
9. child_list deduplication using if child not in child_list is O(n²). A set of handles would be more efficient, though this is negligible for typical family sizes.
10. The $Id: comment block is an empty CVS/SVN relic and can be removed.

---------------------------

1. Is incorrect. The check is supposed to be against the current date. I changed the code to make the current date explicit to avoid any missunderstanding.

2. Is correct. the link col refers to either the wrong or an incorrect column. I changed to code to make sure the first column (the name) is treated as a link.

3. Is incorrect. The code does check for step children, and this is what is intended. When I asked Claude about this they revised their opinion, and checked whether I had intended to include step-children.

4. Is incorrect. I didn't understand the point Claude was making, and when I asked for clarification, they said that the code might be fragile if the function parameter person was later used inside the code. I decided to ignore this suggestion.

5. Is correct (but no change). I could use a set. And in fact in the obscure and possibly impossible event that there are two sources for the same census form (this might be impossible inside the Forms code), I would have two entries in the same found_census for the same form. However, this doesn't matter as I only check whether form_id is found anywhere in the census_list. Also, any performance impact will be tiny because there are so few sources for any person. Finally, the code has been working for me as it is for a long time so it is not worth changing.

6. Is correct (but no change). However with so few elements, any efficiency argument is moot. Also it makes to code unnecessarily complicated to use one Python construct (a list) and convert it at the end (to a tuple) compared with just using a single type. Finally, the code has been working for me as it is for a long time so it is not worth changing.

7. Is correct (but no change). It is not conventional in Gramps to build guards for missing imports, and since the code will be distributed inside the Forms addon, there is no need for a check.

8. Is correct. This was fixed.

9. Is incorrect. I don't want to use a set, because I want the children to be in the order they are in in the database.

10. Is correct. This was fixed.

GaryGriffin:
Can you add `help_url` to the GPR file? 

> I have set the version in the grp file to 1.0.0, I don't know what it should be set to. Should it be the same as the main formgramplet.gpr.py file. 

The version is fine at 1.0.0 . It does not need to be the same as the others.

> In any event, the formgramplet.gpr.py file probably needs to have its version incremented so that this addition is caught as an updated gramplet.

All 3 gramplet versions will be auto-incremented when published, since publish works at the directory level.

kulath:
I've added the help_url as requested by @GaryGriffin. However, I am not sure that it is much use, because only the main (Form) help file is accessible from the various menus in Gramps. Once the Quickreport is accepted, I will add a reference from the Form Wiki to this addon's wiki.

GaryGriffin:
> I've added the help_url as requested by @GaryGriffin. However, I am not sure that it is much use, because only the main (Form) help file is accessible from the various menus in Gramps. Once the Quickreport is accepted, I will add a reference from the Form Wiki to this addon's wiki.

Once the Quickreport is published, the Addon Manager will display the Wiki link for the report. Here is what the view *will* look like (a view of my local `PR testing` repo):
Screenshot 2026-04-04 at 10 28 22 AM

kulath:
> Once the Quickreport is published, the Addon Manager will display the Wiki link for the report. Here is what the view _will_ look like (a view of my local `PR testing` repo):

That's good. When I am running the quickview (just by inserting the code into the plugins directory), all I see is the Form Gramplet. I suppose there is some additional magic around how the addon is 'installed' into the application when it is properly published!

PR #815, opened 2026-04-08
Ensure unicode characters can be written to file

hgohel:
In some regions, e.g. Western Europe, Windows uses different encodings as default (e.g. cp1252) which doesn't handle unicode characters. To ensure that unicode characters can be written to files without an throwing an exception, we specify UTF-8 encoding when opening the file for write.

Fixes #[14057](https://gramps-project.org/bugs/view.php?id=14057)

### DIFF
--- a/MediaVerify/MediaVerify.py
+++ b/MediaVerify/MediaVerify.py
@@ -27,7 +27,6 @@
 #
 #-------------------------------------------------------------------------
 import os
-import io
 
 #-------------------------------------------------------------------------
 #
@@ -210,7 +209,7 @@ class MediaVerify(tool.Tool, ManagedWindow):
                 chooser.destroy()
                 return
         try:
-            with io.open(filename, 'w') as report_file:
+            with open(filename, 'w', encoding='utf-8') as report_file:
                 for title, model in zip(self.titles, self.models):
                     self.export_page(report_file, title, model)
         except IOError as err:


Nick-Hall:
We could also change this to use the built-in `open` function and remove the `io` import at some point.

hgohel:
> change this to use the built-in `open` function

Easy enough to make the change now. Also read in the docs that "Python 3.15 will make Python UTF-8 Mode default." Nice.

hgohel:
@GaryGriffin Please use squash-and-merge for this PR. Thanks.

GaryGriffin:
I only see _Rebase and merge_ (not squash and merge) as an option for this PR when using browser to github. 

Screenshot 2026-04-09 at 9 09 59 AM

I cannot see the Settings for the repo, so cannot check there.

hgohel:
@GaryGriffin No worries - I made the change locally and forced push it up so you can just merge as usual. Thanks.

@Nick-Hall Gary was on the right track that squash-and-merge needs to be enabled for a repo via settings. It's a useful option in many cases (e.g. when there are extra commits for typos, lint checks, minor commits like in this case which don't add value) if you could consider enabling it: 
* [Configuring commit squashing for pull requests](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests)
* [Squashing your merge commits](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/about-merge-methods-on-github#squashing-your-merge-commits)

GaryGriffin:
`Gary was on the right track that squash-and-merge needs to be enabled for a repo via settings. It's a useful option in many cases (e.g. when there are extra commits for typos, lint checks, minor commits like in this case which don't add value) if you could consider enabling it:`

I believe the issue is that the Weblate PRs require Rebase+Merge and will really mess things up if squashed. Safer so that I dont break things by mistake.

hgohel:
> I believe the issue is that the Weblate PRs require Rebase+Merge and will really mess things up if squashed. Safer so that I dont break things by mistake.

Enabling squash-and-merge doesn't take away other options. It will be an additional choice for the merge. You have the choice of not using squash-and-merge even if it's been requested. And given your track record, I have no doubt you'll make the right choice!

Nick-Hall:
I have enabled squash merging, but as @GaryGriffin correctly points out, this should never be used to merge Weblate pull requests.

PR #818, opened 2026-04-17
Fix remaining CI failures: lint, compile, addon structure, Windows

eduralph:
## Summary

Turns the CI pipeline from 6 failing jobs to green by addressing each
failure at its root. All changes are needed together — the CI won't
pass without every piece.

- **Compile Check**: add `shell: bash` so the bash process substitution
  in the find loop parses (default shell is `sh -e`)
- **Addon Structure**: add `po/template.pot` stubs for the 4 addons
  that had no `po/` directory (AnniversariesGramplet, ArchiveAssist,
  GrampsChat, GrampyScript)
- **Unit Tests (Windows)**: switch to conda `environment-file:` so deps
  actually install into `addons-ci` — the prior `mamba install` step
  silently no-op'd and pytest was never available
- **Lint**: 79 → 0 ruff errors across 29 addon files. Mostly missing
  imports (WindowActiveError, ErrorDialog, DbTxn, EditDate, ReportError,
  reduce, time, sys, etc.), plus Python 2 leftovers (`unicode`,
  `basestring`, `reload`, `except E(msg)` syntax) and a handful of real
  bugs (dead code after `return`, wrong variable names, misnamed
  `.grp.py` file never loaded by Gramps)

Full list of per-file changes is in the commit message.

## Test plan

- [ ] Lint job passes (ruff clean)
- [ ] Compile Check job passes (all .py files compile)
- [ ] Addon Structure job passes (every addon has po/template.pot)
- [ ] Unit Tests (Linux) pass
- [ ] Unit Tests (Windows) install the env and run pytest
- [ ] Integration Tests pass
- [ ] Build job passes


kulath:
CLAUDE.md should use Agents.md from Gramps, not a separate file, otherwise there is unnecessary duplication and the need to keep this up to date with gramps Agents.md. Remove it because I don’t think you can have a link to the gramps file.

eduralph:
Hi Tim,

Thanks for pointing this out; the PR was a mistake. I fed it with Agents.md
from Gramps, but it's supposed to stay in my fork. I still have some work
to do before I submit it.

Regards,
Eduard

PR #819, opened 2026-04-17
Wrap label vaue to ensure it is a string

hgohel:
The argument `label` to Gtk.Label must be of type string. In recent versions of Python and/or PyGObject, this check has become stricter and as a result is throwing an exception instead of silently converting to a string. So now we explicitly convert to string.

Fixes #[14181](https://gramps-project.org/bugs/view.php?id=14181)


DaveSch-gramps:
Just tested and no errors given!!

PR #820, opened 2026-04-17
Add CI/CD pipeline with container-based testing (Feature Request ID 9393 in Mantis)

eduralph:
This is part of the answer to [issue 9393](https://gramps-project.org/bugs/view.php?id=9393), using the "X framebuffer" approach (Option 1 of the feature request).

## Scope

This PR is **strictly CI infrastructure**. After review feedback from @kulath (2026-04-18) — "too many completely different changes in a single commit" — all per-addon code/lint/structure work that was originally bundled here has been split into one-PR-per-addon submissions; see the [Companion PRs](#companion-prs) section at the bottom. This PR adds only the workflow files, the container image, the shared CI scripts, and a small shared test harness.

Thanks also to @GaryGriffin for the early testing feedback on bug 9393 and for asking the questions that led to the standalone per-job reproduction commands below.

### CI infrastructure
- `.github/docker/gramps-ci/Dockerfile` — Python 3.12 + Gramps 6.0 + PyGObject + GTK typelibs (incl. `gir1.2-gexiv2-0.10` for EditExifMetadata / PhotoTaggingGramplet) + xvfb/xauth + ruff, intltool, gettext, git. GTK lives in the base image so addon modules that do `from gi.repository import Gtk` at module load are importable; xvfb/xauth are bundled for tests that actually render. Gramps is installed from PyPI for released series, or from a SHA-pinned `gramps-project/gramps@maintenance/grampsNN` snapshot for an unreleased branch (the series is parameterised via the `GRAMPS_SERIES` build arg).
- `.github/workflows/docker-build.yml` — rebuilds the image on `.github/docker/**` changes or via `workflow_dispatch`, deriving the image tag and Gramps series from the branch ref. Default-branch-only; gates on GHCR write access.
- `.github/workflows/ci.yml` — eight jobs:
  - **Setup** — derives the CI image tag and the `make.py` series suffix from the branch ref (`maintenance/grampsNN`), so the rest of the matrix is branch-neutral.
  - **Lint** — `ruff --select=E9,F63,F7,F82` + trailing-whitespace check (uses `grep -P` so `\t` matches a real tab, not a literal `t` — see commit `205b21c` for the regex bug fix).
  - **Addon Structure** — every addon has `po/template.pot`.
  - **Compile Check** — `python3 -m py_compile` on every `.py`.
  - **Unit Tests (Linux)** — `<addon>/tests/test_*.py` (skips `test_windows_*` / `test_integration_*`) via `run_addon_tests.py` (GI bootstrap + per-module timeout + honest skip accounting).
  - **Unit Tests (Windows)** — same matrix on the native Windows runner with a conda+pip env.
  - **Integration Tests (Gramps)** — `tests/test_plugin_registration.py` + `<addon>/tests/test_integration_*.py`, Linux-only, container `--init` so `xvfb-run` doesn't hang.
  - **Build** — `make.py gramps60 build all`.
- `.github/environment.yml` — hybrid conda+pip environment for Windows. Gramps isn't on conda-forge, so `pygobject`/`gtk3` come from conda; only the stable base (`gramps` + `orjson`) comes from pip. Addon runtime deps (`dbf`, `networkx`, `lxml`, …) are no longer pinned here — `ci.yml` auto-derives and installs them at runtime from each `.gpr.py`'s `requires_mod` (single source of truth; `dbf` dropped in commit `8d2654a`).

### Shared CI scripts
- `.github/scripts/addon_system_deps.py` — single source of truth for addon **system** dependencies. Maps each `requires_gi` typelib / `requires_exe` executable declared in a `.gpr.py` to its package on each CI platform (apt vs conda), and scans the tree so `ci.yml` derives the install list from one place instead of a hand-kept list. Records platform asymmetry (the GTK 3 libs — goocanvas, osm-gps-map, gexiv2 — exist on apt but not conda-forge). Pure stdlib.
- `.github/scripts/run_addon_tests.py` — per-addon unit/integration runner that replaces a bare `python -m unittest`: it (1) bootstraps the GI versions the Gramps GUI launcher pins (Pango/PangoCairo/Gtk) before any `gramps.gui` import, (2) runs each module in a subprocess with a wall-clock timeout so a hung test can't stall the job, and (3) FAILs a wholly-skipped module *unless* the addon's declared system deps are genuinely unavailable on the platform — so a silent skip can't read as a pass.
- `.github/scripts/gi_bootstrap/sitecustomize.py` — the same GI version pin as a `sitecustomize` shim, placed on `PYTHONPATH` for the discover-/subprocess-loading steps (e.g. plugin registration) so every spawned interpreter inherits the bootstrap.

### Shared test harness
- `tests/gramps_test_env.py` — `GrampsTestCase` / `GrampsDbTestCase` base classes (stdlib `unittest`, mirroring upstream Gramps' own test style).
- `tests/test_plugin_registration.py` — registers every addon in a subprocess (crash-safe), verifies `gramps_target_version=6.0` plus valid id/name/version, smoke-tests import/export entry functions.
- `tests/__init__.py`.

### Docs
- `CONTRIBUTING.md` — a short note that PRs now run automated CI checks, and that a green check on an unreleased `maintenance/grampsNN` branch means the addon works against that *branch tip* (a SHA-pinned snapshot), not against a tagged release.
- `.github/CI-MAINTAINER.md` — operational runbook for a `gramps-project/addons-source` maintainer: one-time setup when this PR merges, and how to stand up a new maintenance branch. The pipeline is otherwise self-driving.

## Gate policy

PR-blocking checks should only fire on issues the PR's own diff can cause. Every test here runs gramps or addon code, so the policy is:

- **Blocking gates** (job-level `continue-on-error: false`): Lint, Compile Check, Unit Tests (Linux), Unit Tests (Windows), Integration Tests, Build.
- **Advisory gate** (job-level `continue-on-error: true`): Addon Structure — the only advisory job.

Lint and both Unit-test jobs were advisory while their pre-existing backlog was being cleared by the companion PRs; that backlog is now cleared on `maintenance/gramps60`, so they were flipped to blocking (lint in `d265612`, both unit-test jobs in `0dd3f1b`). The single remaining advisory is **Addon Structure**, which stays non-blocking until the four addons missing `po/template.pot` are fixed in a follow-up PR — flip it off in that PR.

## Commits

| SHA       | Date       | Summary |
|-----------|------------|---------|
| `774a9ac` | 2026-04-19 | Add CI/CD pipeline with container-based testing (issue 9393). Originally larger; force-pushed to this minimal scope after @kulath's review. |
| `c6aa10e` | 2026-04-20 | Auto-derive addon pip deps from `requires_mod` in `.gpr.py`. |
| `8d2654a` | 2026-04-20 | Remove `dbf` from image/env — installed via auto-derive now. |
| `28febdc` | 2026-04-20 | `shell: bash` on unit-test-linux + integration-test steps (fixes Dash `Bad substitution`). |
| `715e71d` | 2026-04-20 | OS-split addon tests by filename convention (`test_linux_*` / `test_windows_*` / `test_integration_*`). |
| `dd0fd38` | 2026-05-11 | Add `gir1.2-gexiv2-0.10` typelib to image (companion to #878, #880). |
| `205b21c` | 2026-05-11 | Lint trailing-whitespace check — switch BRE `[ \t]` to PCRE. The previous BRE pattern matched any line ending in `t`/`\`/`[`/`]`/space (~3 000 false positives across ~430 files); PCRE (`-P`) recognises `\t` as a tab — real count 597 lines across 21 files. |
| `f5907af` | 2026-05-15 | Skip addons whose `.gpr.py` declares `include_in_listing=False`. |
| `ff215ac` | 2026-05-16 | Validate `requires_mod` names against `find_spec` (Pillow/PIL trap). |
| `d265612` | 2026-05-19 | Make the **lint** job blocking — ruff backlog cleared on `maintenance/gramps60`. |
| `0dd3f1b` | 2026-05-19 | Make **unit-test-linux** and **unit-test-windows** blocking — backlog cleared. |
| `9a91d89` | 2026-05-18 | Derive image tag and `make.py` argument from the branch ref. |
| `abefdbe` | 2026-05-18 | docker-build: derive image tag and Gramps series from the branch ref. |
| `9927626` | 2026-05-18 | Dockerfile: parameterise Gramps series via `GRAMPS_SERIES` build arg. |
| `894e484` | 2026-05-18 | CI image: hybrid PyPI / git-clone install for unreleased maintenance branches. |
| `73d75ed` | 2026-05-18 | docs: note the first-push race and unreleased-branch CI semantics. |
| `3b2a947` | 2026-05-19 | docs: add CI maintainer runbook (`.github/CI-MAINTAINER.md`). |
| `ad2815c` | 2026-05-24 | tests: detect addons that import a sibling without `depends_on`. |
| `bc8df21` | 2026-05-24 | tests: derive expected addon target version from the Gramps install. |
| `7f8a839` | 2026-05-29 | ci: install addon system deps, run under xvfb, pin GI, fail on silent skips. |
| `1466491` | 2026-05-30 | ci(windows): document the gramps-vs-branch series caveat. |
| `06b95bc` | 2026-06-13 | Strip the dependency detector and TMG test split out of the CI PR — moved to their own submissions so this PR stays strictly CI infrastructure (per @kulath's one-concern-per-PR feedback). |

## Local reproduction

This PR can't be exercised against `gramps-project/addons-source` directly (workflows live in the PR), but the per-job commands are reproducible standalone:

```bash
# Lint
ruff check --select=E9,F63,F7,F82 --no-fix --exclude='*.gpr.py' .
git --no-pager grep -P -n --full-name '[ \t]+$' -- '*.py' && echo FAIL || echo OK

# Addon Structure
for gpr in */*.gpr.py; do
  d="$(dirname "$gpr")"
  [ -f "$d/po/template.pot" ] || echo "MISSING: $d"
done

# Compile Check
find . -name '*.py' ! -path './.git/*' ! -path '*/__pycache__/*' \
  -exec python3 -m py_compile {} +
```

A complete end-to-end test is also available by forking to a personal repo and pushing a branch — the workflows fire because the workflow files come along with the fork. See the comment thread for the step-by-step walkthrough.

## Companion PRs

Code/lint/structure work that was originally bundled here is now submitted as one-PR-per-addon. Tracker:

- **Compile check (4)** — #834 Query, #835 HouseTimelineGramplet, #836 Themes, #837 lxml.
- **Lint backlog (24)** — #843, #847-#867 (excl. #868), #869 SurnameMappingGramplet (rename + migration).
- **Real bugs (5)** — #870 QuiltView `displayer`, #871 JSON `data`, #872 LifeLineChartView tuple-in-if, #873 libaccess lambda (thanks to @GaryGriffin for catching that the first version of #873 was missing the actual code edit), #874 lxml module-level `self`.
- **Plugin-registration migrations (3)** — #875 Wordle, #876 SourceReferences, #877 RebuildTypes.
- **Dependency declarations (3)** — #878 EditExifMetadata (Pillow + GExiv2), #879 MongoDB (pymongo), #880 PhotoTaggingGramplet (GExiv2).
- **Trailing whitespace (6)** — #881 GeoTimeLines, #882 DNA, #883 GraphView, #884 FamilyTreeDNA, #885 GrampyScript, #886 ToDoReport.
- **Gramps core companions (2)** — gramps-project/gramps#2299 (ClipboardGramplet headless-icon fallback), gramps-project/gramps#2302 (`webreport/common.py` no-ICU fallback) — both required to unblock plugin-registration smoke tests in this PR.

As of HEAD the lint and unit-test backlogs are cleared and those gates are blocking; once the remaining `po/template.pot` gaps land, Addon Structure can be promoted to blocking too.


GaryGriffin:
Added Note to https://gramps-project.org/bugs/view.php?id=9393 requesting testing and feedback of this PR.

eduralph:
Corrected to conform to the agents.md guidelines, including using unittest instead of pytest

GaryGriffin:
Can you provide the commands to test the CI workflow. And a usage of the test harness.

kulath:
How did the code work without the imports you have added

kulath:
As per comment on another PR, the CLAUDE.md file should not be here as it just duplicates Agents.md in the main Gramps repository meaning any changes etc. will have to be made in two place and it will be hard to keep the files in step.

kulath:
Thanks for fixing so many bugs and problems, but there are too many completely different changes in a single commit. Should they not be made in several commits(or even separate PRs).

eduralph:
Yeah, I wanted more at one go, I can see that it might not be the right approach. 

@kulath , there are many Lint issues and I chose to fix them them instead of dropping the Lint. I thought this might get some pushback. We can do it the other way around - I'll remove the Linting and we can make issues out of them as preperation for activating them.

eduralph:
I've significantly reduced the scope of this PR based on @kulath's feedback about "too many completely different changes." It's now strictly CI infrastructure — 7 files, one commit.

**What's in the PR**

- .github/docker/gramps-ci/Dockerfile — container image for Linux CI jobs
- .github/environment.yml — conda environment for Windows CI
- .github/workflows/ci.yml — the CI workflow itself (lint, structure, compile, unit tests on Linux+Windows, integration, build)
- .github/workflows/docker-build.yml — publishes the CI image to GHCR
- tests/__init__.py
- tests/gramps_test_env.py — shared GrampsTestCase / GrampsDbTestCase base classes
- tests/test_plugin_registration.py — sanity checks that every registered plugin imports

Everything else from the earlier revision (lint fixes across ~29 addons, po/template.pot stubs for the four addons missing them, TMGimporter tests, the SurnameMappingGramplet.grp.py → .gpr.py rename, CLAUDE.md) has been removed and will be submitted as separate PRs.

**Gate Policy**

All seven jobs run on every push and PR. To avoid showering innocent contributors with red CI for pre-existing tree state, four jobs are marked continue-on-error: true so they surface issues without blocking merges:

lint — non-blocking (~79 pre-existing ruff E9/F63/F7/F82 errors)
addon-structure — non-blocking (4 addons missing po/template.pot)
unit-test-linux — non-blocking (some addon test modules fail to import today)
unit-test-windows — non-blocking (same)
compile-check — blocking (green today)
integration-test — blocking (green today)
build — blocking (green today)

In order to activate all gates, there will be some rework of existing addons needed

eduralph:
@GaryGriffin — here are the commands for each CI job, lifted straight from ci.yml. Run from a checkout of addons-source with Gramps importable (either PYTHONPATH / GRAMPS_RESOURCES pointing at a Gramps checkout, or inside the CI container image).

**Lint — syntax/import errors plus trailing whitespace**

ruff check --select=E9,F63,F7,F82 --no-fix --exclude='*.gpr.py' .
git --no-pager grep -n --full-name '[ \t]$' -- '*.py' && echo FAIL || echo OK

**Addon structure — every addon must have po/template.pot**

for gpr in */*.gpr.py; do
  d="$(dirname "$gpr")"
  [ -f "$d/po/template.pot" ] || echo "MISSING: $d"
done

**Compile check — py_compile on every .py**

find . -name '*.py' ! -name '*.gpr.py' ! -path './.git/*' ! -path '*/__pycache__/*' \
  -exec python3 -m py_compile {} +

**Per-addon unit tests — matches both the Linux and Windows jobs**

export PYTHONPATH=.
modules=""
for f in */tests/test_*.py; do
  [ -f "$f" ] || continue
  case "$(basename "$f")" in test_integration*) continue ;; esac
  case "$f" in Sqlite/tests/test_sqlite.py) continue ;; esac
  mod="${f%.py}"; modules="$modules ${mod//\//.}"
done
python3 -m unittest -v $modules

**Integration tests — plugin registration plus per-addon integration**

export PYTHONPATH=.
python3 -m unittest discover -s tests -p 'test_*.py' -t . -v

*per-addon integration uses the same loop as the unit block above*
*but matching test_integration*.py instead*

GaryGriffin:
> here are the commands for each CI job, lifted straight from ci.yml. 

Sorry, what I meant was, is there a way to invoke the CI process (not the commands that are invoked by the CI process) manually to test the github actions. 

I did try the invoked commands:

**Lint:** I dont have ruff on my Mac. How would I invoke the CI lint process within github as an action.

**Addon Structure:**  if I manually invoke the commands on my Mac that is in the comment, it fails (due to escaping special chars needed for comments). I need to use:

> for gpr in \*/\*\.gpr.py; do
> d="$(dirname "$gpr")"
> [ -f "$d/po/template.pot" ] || echo "MISSING: $d"
> done

Results: 4 issues, as you stated. I do not know if it is a requirement that the template.pot exists. If there are no translated strings, then there shouldn't need to be one, I think. Maybe I am wrong.

**Compile Check:** same issue with escaping special chars. I think the gpr.py should actually be included. I need to use:

> find . -name \*.py -exec python3 -m py_compile {} \\\;

Results: 4 issues - Themes, Query, HouseTimelineGramplet, and lxml

**Integration Tests:** failed with error `ImportError: Start directory is not importable: 'tests'` .

eduralph:
Can you tell me what you are trying to do or understand? I'm not sure if I'm giving you the right answers. 

The tests run in containers stored on ghcr.io containing all the necessary tools. It's optimized for steady-state but is kind of a pain for this specific PR, because you can't really test it in this environment. The upside is that the image is stable across hundreds of normal PRs. It only churns when someone explicitly wants the CI environment to change. 

- The Dockerfile changes rarely. Maybe once per Gramps minor, or when a system dependency is added. Most weeks, nobody touches .github/docker/**.
- The path filter ensures docker-build doesn't fire on every push — only when the Dockerfile actually changes. So in steady state, ci.yml pulls a stable, cached image and runs fast.
- The "default branch only" scope ensures one canonical image per supported branch (:gramps60). PR-built images would proliferate (:gramps60-pr123, :gramps60-pr124...) and clutter GHCR with throwaway tags that need lifecycle management.
- GHCR write access is gated to default-branch context. Fork PRs have read-only GITHUB_TOKEN, so a pull_request: trigger on docker-build would silently no-op for external contributors anyway. Even internal PRs publishing to GHCR would mean unmerged code can write to the registry — a small but real trust-boundary violation. 

You will have to make the images publically available for the PRs to work, but I don't think that's an issue. There will be some maintainance work needed to be done once you flip over to branch61 as default, I'd see what can be done to make it a bit easier then, assuming you like the current process. 

In order for you to verify how it works, take a poke at my forked repo where I've merged the PR. This is how you can do it

1. Fork eduralph/addons-source to your account using GitHub's "Fork" button: https://github.com/eduralph/addons-source/fork. 
2. Create a branch off maintenance/gramps60 on your new fork — e.g., test/ci-tryout. The CI workflow files come along automatically. 
3. Make any change you want to exercise the pipeline, or none at all if you just want to see a clean run. Some ideas:
    - Add a deliberate import nonexistent_module to any .py file → trips the Lint job.
    - Add a stray syntax error → trips the Compile Check job..
    - Modify a real addon you care about → exercises the Build job and unit/integration tests.
4. Commit, push, and open a PR from your branch targeting eduralph/addons-source:maintenance/gramps60. (Not gramps-project/addons-source — point the PR base at my fork.)
5. Wait for approve the first workflow run. Because you'll be a first-time contributor to my fork, GitHub will hold the run for manual approval — click the "Approve and run workflows" button on the PR. Subsequent runs go automatically.
6. Watch the Checks tab on your PR. You'll see 7 jobs fire:
    - 5 Linux jobs (Lint, Addon Structure, Compile Check, Unit Tests Linux, Integration Tests, Build) running in the public container.
    -  1 Windows job (Unit Tests Windows) on a GitHub-hosted Windows runner with a conda environment.
    -  The advisory ones (Lint, Addon Structure, Unit Tests Linux/Windows) will report red without blocking your PR; the blocking ones (Compile Check, Integration Tests, Build) gate the merge.

GaryGriffin:
> Can you tell me what you are trying to do or understand? I'm not sure if I'm giving you the right answers.

I should have been more explicit. I am trying to test the implementation before merging/publishing this PR. Understanding the resources needed and the frequency of activity is part of that. And making sure that it functions completely and as expected. 

I was hoping for something like a github command to activate the action so I can see it work. Given your complete description (much appreciated), I think I am going to have to yield to @Nick-Hall to review/merge/publish this one since it impacts the repo actions in ways far above my knowledge of github. For instance, I dont know if these are the right blocking/non-blocking decisions. Or if the scope should just be the Addon impacted by the PR rather than all Addons.

eduralph:
@GaryGriffin - I thought of adding a manual button for everything, but that only works if it's merged with the default branch, lol

eduralph:
I just reworked the CI approach a bit to be able to differentiate between Base, Linux & Windows test specifically. Also added automated dependency loading for future unit tests. The Unit Tests currently fail because of an issue in Websearch, but that needs to be adressed seperately.

GaryGriffin:
It might be useful to skip an Addon directory if the gpr has `include_in_listing=False` since that Addon does not get built or released. There are 13 Addons that are of this type (HTMLView, TMGimporter, RebuildTypes, PhpGedView, FaceDetection, CheckPlaceTitles, Wordle, SourceIndex, SourceReferences, DetID, Query, MongoDB, SurnameMappingGramplet) and most are no longer functional.

eduralph:
Makes sense, that way the addons being included can be controlled and intermediate development is still under source control. Expanding on that idea, a life cycle might make sense, i.e. a method when then addons get decommissioned and ultimately no longer promoted to the next branch. Some hard and fast criteria might be good. I'll share that idea on the discussion board.

In the mean-time I'd add the ignore in listing functionality as you suggest. 

One minor item: I had reworked the TMGimporter, so I wouldn't necessarily include that in the list.

eduralph:
@GaryGriffin - Done — implemented in f5907afc9. The check is inlined per job step (no separate helper script) so each step stays self-contained, mirroring the pattern the workflow already uses for the requires_mod auto-derive.

Across the current addon tree, 11 directories are now skipped from lint / compile / unit / integration / plugin-registration: CheckPlaceTitles, DetId, FaceDetection, HtmlView, MongoDB, PhpGedView, Query, RebuildTypes, SourceIndex, SourceReferences, WordleGramplet. 

One nuance: TMGimporter stays gated. Its main register() is include_in_listing=True; only the three conditional registers for the import of individual files .PJC, .TMG, .VER use False. The rule is "skip an addon iff every register() in its .gpr.py(s) sets include_in_listing=False" — any True (or no include_in_listing at all, which defaults to True) keeps the addon active. To gate a currently-skipped addon, flip include_in_listing to True on at least one register call; to un-gate a currently-active one, flip them all to False.

Let me know if that approach is ok with you. Otherwise, I'd change the logic in TMGimporter.

GaryGriffin:
> One nuance: TMGimporter stays gated. Its main register() is include_in_listing=True; only the three conditional registers for the import of individual files .PJC, .TMG, .VER use False. The rule is "skip an addon iff every register() in its .gpr.py(s) sets include_in_listing=False" — any True (or no include_in_listing at all, which defaults to True) keeps the addon active.

I agree with your logic.

eduralph:
@GaryGriffin  - based on the issues in PR #890 I added functionality to catch such issues during building

eduralph:
@GaryGriffin - I tested the CI in my local repo and here's the latest status

Job	                                                         Status	        PR 820 gate
- Compile Check	                                ✓ success	blocking
- Build	                                                ✓ success	blocking
- Integration Tests (Gramps)	        ✓ success	blocking
- Unit Tests (Linux)	                        ✓ success	non-blocking
- Unit Tests (Windows)	                ✓ success	non-blocking
- Lint	                                                ✗ failure	non-blocking (advisory)
- Addon Structure	                        ✗ failure	non-blocking (advisory)

Linting will flip to success if you merge #837 , #856 and #874

The Addon Structure item seems unlikely it will turn green in gramps60 as the discussion around the weblate commit on that branch seems to be it won't happen. This is a requirement I introduced in order to ensure that add-ons are properly translated, but strictly speaking it's not a requirement (yet). It should move to green in gramps61 one though.

eduralph:
@GaryGriffin , I've been working on testing the CI on the 6.1 Branch exapnding the CI coverage for all maintainance branches from 6.0 forward. 

The workflows now derive suffix/series/image-tag from github.ref_name, so pushing maintenance/gramps61 builds gramps-ci:gramps61 with gramps==6.1.* without per-branch edits — single setup job in ci.yml, one params step in docker-build.yml, one ARG GRAMPS_SERIES in the Dockerfile.

There are a couple of implications for the gramps images stemming from the fact that 6.0 is a stable release where we just need to build the image once to test addon changes and 6.1 where you have essentially a "moving target".

pip install gramps==6.1.* fails because 6.1 isn't on PyPI yet. PR 820 as-is works on every released maintenance branch, but unreleased ones are blocked until they ship.

The solution (1) is to build a hybrid dockerfile which does following: pip-install first, fall back to a SHA-pinned git clone gramps-project/gramps@maintenance/grampsNN on "no matching version" (not on network errors). One image build path covers every branch — trade-off is that fallback-built images test against the upstream branch's tip, so green CI on an unreleased branch means "addons work with current tip" rather than "addons work with 6.1.0".

That's a semantic difference in terms of what the CI workflow tells you: green on 6.0 means addons work against Gramps 6.0.8 - tagged reproducible artifact users have installed while on gramps 6.1 it means "addons work against whatever was at the tip of maintenance/gramps61 when the image was built" which is a moving target.  PR reviewers reading a green check for an addon-PR-to-gramps61 can't tell which world they're in without opening the build log. We can mitigate with a loud ::notice:: (released path) vs ::warning:: (fallback path) in the docker-build log, but you need to understand that distinction as maintainer.

The alternative (2) is to restrict the CI against releases only until the branch lands on PyPI, but that's not the delivery strategy right now, i.e. ship the addons to a later date to gramps - correct?

I'd suggest going with solution 1, you just need to be aware of the differences. 

PS - this would obviously switch over to  API Versioning, i.e. if API Version 1.0.0 is stable, the CI would be solid no matter what happens to gramps until we bump it to 1.1.0 or 2.0.0 or such.

eduralph:
With the lint backlog merged and the plugin-registration / dep-declaration round in, the gate structure is now:

| Job | Blocking | Notes |
|---|---|---|
| Lint | yes | ruff E9/F63/F7/F82 + trailing-whitespace; flipped to blocking once #843 + #847-#869 landed |
| Compile Check | yes | py_compile across all `include_in_listing=True` addons |
| Unit Tests (Linux) | yes | flipped to blocking after #875-#880, #869, gramps#2299 fixed the import-failure backlog |
| Unit Tests (Windows) | yes | same |
| Integration Tests (Gramps) | yes | plugin registration + per-addon integration |
| Build | yes | `make.py` build |
| Addon Structure | **no** (`continue-on-error: true`) | po/template.pot existence check; #838-#841 closed unmerged, gate stays advisory |

Six of seven jobs are blocking. The Addon Structure check is the only remaining advisory gate.

Once I have your input Gary (Hybrid dockerfile - solution 1) or wait for gramps branch (solution 2)  I can merge in the last piece for gramps61 branch.

GaryGriffin:
I think option 1 is the only one viable. Normally there are a couple beta releases in a cycle that can easily be 2 months. I am expecting 6.2 to be in the Q4 timeframe, so beta could be Q3 timeframe. 6.1 is a special FamilySearch Integration release. 

Just so you know - I actually release 6.x Addons BEFORE 6.x gramps beta is available. For 6.1, I think the source release was the same day but could easily be a few days. The Linux (snap, I think) release is next. Then the Mac release typically. Last is the Win release. It could be a week before all are built, depending on bundler availability.

eduralph:
@GaryGriffin - I added a quality gate that checks for internal dependencies to avoid errors like [13707](https://gramps-project.org/bugs/view.php?id=13707); the good news is that isn't a often recurring issue.

eduralph:
This push adds four CI fixes so the per-addon test pipeline actually *exercises* addons that need a display or system packages, instead of silently skipping them and staying green. Each change is small and independent — here is what each solves and why it is needed.

### 1. Install addon system dependencies (`requires_gi` / `requires_exe`)
Addons declare GI typelibs (e.g. `GooCanvas`) and executables (e.g. `dot` from graphviz) in their `.gpr.py`. These are not pip-installable, and Gramps' own `Requirements` only *checks* them — so the CI never installed them. Any addon needing one could not be loaded, its tests skipped, and the job stayed green, hiding the gap.

`.github/scripts/addon_system_deps.py` is now the single source mapping each declared dependency to its per-platform package. The Linux jobs derive the apt set from the `.gpr.py` files and install it at runtime (the container runs as root), mirroring the existing `requires_mod` derivation; a drift-guard step fails if an addon declares a dep with no map entry. Without this, the pipeline's headline feature — per-addon testing — quietly no-ops for every GUI/graph addon.

*Platform note:* the GTK-3 libs `goocanvas` / `osm-gps-map` / `gexiv2` are not on conda-forge, so on the Windows lane those addons cannot be installed and skip by necessity. The map records that explicitly, and the runner tolerates it (see #4).

### 2. Run the per-addon tests under `xvfb`
Several addons build a `Gtk` style context at import time, which needs a display connection; with none, the interpreter hard-aborts with a `Gtk-ERROR` rather than running or skipping cleanly. The unit and integration test runs are now wrapped in `xvfb-run`.

(The plugin-registration step is deliberately *not* under xvfb: it only imports addon modules in subprocesses and tolerates load failures. Giving it a display made an addon load hang on the absent AT-SPI bus until the per-load timeout, so it runs headless as before.)

### 3. Pin the GI versions before tests import `gramps.gui`
Gramps pins its GI versions in the GUI launcher (`gramps/gui/grampsgui.py` calls `require_version` for `Pango`/`PangoCairo`/`Gtk` at import). A unit test that imports a `gramps.gui.*` module directly never runs that launcher, so `Gtk`/`Pango` get imported with no version pinned first — emitting a `PyGIWarning` and, on a host where GTK 4 is the default, risking the wrong stack (and a silent all-skip).

`.github/scripts/run_addon_tests.py` (and a `gi_bootstrap` sitecustomize for the subprocess-loading plugin-registration test) perform the same `require_version` bootstrap, so addon tests run under the supported GTK 3 stack exactly as a real Gramps session does.

### 4. Fail on degraded coverage (a wholly-skipped module)
`unittest` / `xmlrunner` exit 0 when every test skips, so an addon whose tests all skip — e.g. an import guard tripping on a missing dependency — read as PASS, indistinguishable from one that actually ran. `run_addon_tests.py` now fails a module whose tests all skipped, **unless** the addon's declared system deps are unavailable on that platform (e.g. `goocanvas` on conda), in which case the skip is expected and tolerated. This is the control that turns "tested nothing" into a visible red instead of a false green.

---

The pipeline stays branch-neutral (the series is derived from the ref). I validated these changes on my fork's CI against both `maintenance/gramps60` and `maintenance/gramps61`: the Linux unit-test and integration jobs pass on both, and the Windows (conda) lane passes on gramps60. The `Lint` and `po/template.pot` reds seen on those runs are pre-existing addon-content issues, unrelated to these changes.

eduralph:
**Follow-up: per-module timeout (run_addon_tests).** Validating the above against `maintenance/gramps61` surfaced a hang on the Windows lane — `TMGimporter.tests.test_libtmg.TestImportCitations.test_citation_attached_to_event` (a gramps61-only TMG→Gramps citation import) blocks on Windows. Neither plain `unittest` nor `run_addon_tests` had a timeout, so the job hung indefinitely.

`run_addon_tests.py` now runs each module in its own subprocess with a per-module wall clock (`MODULE_TIMEOUT_S`, default 300s, overridable via env). A module that exceeds it is killed and reported as a **FAILURE** that names the culprit, so the job stays bounded instead of hanging. The GI bootstrap and skip-accounting are unchanged (the bootstrap now runs in the worker subprocess).

Re-validated on the fork CI: `maintenance/gramps60` is green; `maintenance/gramps61` now **completes** the Windows job (~9 min) and reports the TMGimporter test as a bounded timeout-failure rather than hanging — Linux unit + integration stay green. That TMGimporter hang is a separate gramps61/Windows addon issue (it needs a Windows skip or a fix in the addon); the pipeline now surfaces it honestly instead of stalling.

dsblank:
Note that if/when the mini-installer from https://github.com/gramps-project/gramps/pull/2308 is merged, that it uses slightly different names (pip names rather than import names).

Nick-Hall:
I plan to review #2308 next and it will probably be merged.

eduralph:
@dsblank - I've added some stuff on that account to PR 820 now
@Nick-Hall  - I'll update the rest of the dependency once PR 2308 lands

dsblank:
How far away are we from having even a basic CI test framework ready?

dsblank:
Noting that https://github.com/gramps-project/gramps/pull/2308 has been merged.

eduralph:
@dsblank — PR #2308 is now accounted for (commits 427f8c3a9 + ad98e1d30):

- **Name mapping single-sourced.** `addon_python_deps.py` now reads `_IMPORT_TO_PYPI` from the installed gramps' `gen/utils/pypi.py` at lookup time, so 6.1+ lanes install exactly what Gramps' own installer would — including entries added upstream later. The local table remains only as a fallback mirror for lanes where gramps predates 6.1 (the gramps60 image, the conda-forge Windows lane, bare runners), with a sync-guard test (`tests/test_addon_python_deps.py::GrampsTableSync`) that fails on any drift wherever gramps ≥ 6.1 is importable.
- **The `--check-resolves` gate delegates to `Requirements().check_mod`**, so it matches whichever series the lane ships: `find_spec`-only on 6.0, `find_spec` plus a real import on 6.1+.
- Fixed along the way: the pip-installed-ness probe used the raw import name (`pip show PIL` fails even with Pillow installed), which would have silently skipped exactly the declarations the mapping exists for; it now probes by the mapped distribution name while still judging the raw import name.

One small upstream observation from wiring this up: `resolve_pypi_name()`'s warning advises addon authors to update `requires_mod` to the PyPI name (e.g. `"Pillow"`), but `check_mod` verifies the *import* name — `find_spec("Pillow")` fails, so an addon following that advice would report its dependency as unsatisfied. The import name plus the `_IMPORT_TO_PYPI` translation seems to be the consistent contract; the warning text might be worth revisiting in gramps.

eduralph:
Ran an adversarial review of this PR (four independent passes: workflow logic, helper scripts, test adequacy via mutation testing, and CI security) and pushed fixes for everything that survived verification. Summary of what changed:

**Blocking (CI couldn't pass its own first run):**
- The lint gate was red on pre-existing base files — 27 trailing-whitespace lines and a real `UnboundLocalError` in `AssociationsTool` (`for _ in range(...)` shadowing the gettext `_`). Split into a separate PR against `maintenance/gramps60` (#996) since it's not pipeline code; this PR's lint lane goes green once that merges and is merged back.
- Fixed a false-red I'd introduced in `tests/test_addon_python_deps.py`: the `{"gramps": None}` overlay didn't force `ImportError` when a `gramps.*` submodule was already cached (which the integration lane does via `test_plugin_registration`), so `test_stdlib_fallback` took the delegated path and failed. The overlay now None-outs every cached `gramps.*` key.

**Test blind spots (mutation-tested):**
- `test_no_requires_mod_heredoc_remains` was a tautology (its regex could never match the old heredoc's literal `\s*` text) — now matches the heredoc's real fragments.
- The Dockerfile toolchain assertions were line-based and missed a multi-line purge, `apt-get -y purge` ordering, and dropping `python3-dev` — now scans logical lines and asserts all three tools.
- `CheckResolvesGate` stubs were constant lambdas that let a "judge by distribution name" mutant survive — now assert the gate judges the raw import name.

**Security:**
- Added `permissions: contents: read` to `ci.yml` (it runs addon-influenced pip/import code and had no permissions block).
- The `--unmapped` drift gate ran only on the Linux lane; `unit-test-windows` pip-installed derived `requires_mod` unguarded. Added the gate there too.
- Pinned `ruff` in the image so the lint verdict can't drift across rebuilds.

**Correctness / fail-open:**
- `check_resolves` now fails (not skips) when a wheel-only `requires_mod` never installed — a pure wheel that doesn't install is a real regression, not an environment gap. Source-built deps stay advisory.
- `run_addon_tests.py`: the per-module timeout was defeated by a grandchild holding the stdout pipe (fixed with process-group kill + bounded wait); a module-level `ImportError` was misclassified (it becomes a `_FailedTest`, not a raise) while a `SyntaxError` on an unsatisfiable platform was excused — now the load-failure taxonomy is honest (dep-shaped → skippable, code bug → always fails); a zero-test module (read as green by unittest) now fails; a non-integer timeout env no longer crashes the runner.
- Both dep scanners now skip a tuple-shaped `requires_mod` entry with a note instead of crashing with a `TypeError`.
- `is_active` is now ast-based/per-register/comment-proof (matching make.py's per-registration rule), replacing the file-granular grep — verified behaviour-identical across all 146 current addon dirs, with the whole-tree oracle embedded as a test.

**Lifecycle / docs:**
- `docker-build.yml`: PR builds validate the image without pushing, a `no-cache` dispatch input, and a weekly refresh; documented the one-push image lag.
- Corrected the environment.yml/ci.yml comments that claimed the Windows lane's gramps comes from conda-forge (it's PyPI) and "self-heals" to 6.1 (the pin needs a manual per-branch bump); lowercased the GHCR image ref so mixed-case fork owners can pull.

13 commits total. All five CI-owned test modules pass locally (49 tests, ruff clean). Next I'm validating the full container pipeline on both the 6.0 and 6.1 lanes on my fork before considering this ready.

dsblank:
> One small upstream observation from wiring this up: resolve_pypi_name()'s warning advises addon authors to update requires_mod to the PyPI name (e.g. "Pillow"), but check_mod verifies the import name — find_spec("Pillow") fails, so an addon following that advice would report its dependency as unsatisfied. The import name plus the _IMPORT_TO_PYPI translation seems to be the consistent contract; the warning text might be worth revisiting in gramps.

What should we do here? Is just changing the warning text enough? Or does something need to change for a proper check?

Also noting relationship to: https://github.com/gramps-project/gramps/pull/2250

dsblank:
CC: @DavidMStraub

PR #821, opened 2026-04-17
Form: fix crash and surface clear errors for malformed XML (bug 11707)

eduralph:
## Summary

- Fixes the `ValueError: not enough values to unpack (expected 2, got 1)` crash in the Forms addon's `FamilySection` when a `<section type='family'>` title omits the `X/Y` separator ([gramps-project/gramps#11707](https://gramps-project.org/bugs/view.php?id=11707)).
- Adds a pure-Python `form_validator` module so the loader can reject malformed XML up front, with user-facing `ErrorDialog`s naming the file, form, and rule that failed.
- Partially addresses [#11010](https://gramps-project.org/bugs/view.php?id=11010) (request for a user error dialog for unsupported elements) by covering its core ask: clear errors for invalid section types, missing/empty `role`, missing/empty `type`, and XML syntax errors.

## What changed

**Loader (`Form/form.py`)**

- Pre-validates each definition file; syntax errors, missing required attributes, and invalid section types all surface as `ErrorDialog`s with the file path and a precise message.
- Partially-broken files still load their well-formed `<form>` entries — one bad section no longer nukes the whole file.
- Added diagnostic logging: `INFO` count of forms loaded per file, `DEBUG` per file/form trace.

**Validator (`Form/form_validator.py`, new)**

- `split_family_title()` — belt-and-braces helper so `FamilySection` never raises on a malformed title.
- `validate_form_element()` / `validate_form_dom()` — returns a list of human-readable errors; free of GTK/Gramps imports so it runs in any CI job.
- `parse_and_validate()` — file-level entry point that catches both parse and schema errors cleanly.

**Editor (`Form/editform.py`)**

- Uses `split_family_title()` instead of the raw `title.split('/')` that blew up in #11707.
- Added a module logger with `DEBUG` on `EditForm` open and a `WARNING` in `FamilySection` if the title lacks the separator (so future regressions are visible in logs, not just as a traceback).

**Translations (`Form/po/template.pot`)** — three new translatable dialog titles.

## Tests

All green locally (38 passed, 7 subtests):

- `Form/tests/test_form_validator.py` — 32 pure-Python unit tests: every branch of `split_family_title`, every validation rule, `parse_and_validate` file handling, plus a sanity test that every shipped `form_*.xml` passes validation.
[...]

PR #822, opened 2026-04-17
Form: detect empty definition files and warn on column-size mismatches (bug 11010)

eduralph:
## Summary

Closes [Gramps bug 11010](https://gramps-project.org/bugs/view.php?id=11010).

The Form loader previously accepted form definition files that had a `<forms>` root element with no `<form>` children, silently producing zero usable forms and leaving the user with no feedback. It also had no diagnostic for a common authoring mistake the bug reporter called out — `<column>` `<size>` values are expected to sum to 100.

This change:

- Extends `validate_form_dom` to report `<forms>` roots containing no `<form>` definitions as an error, surfacing them through the same `ErrorDialog` path introduced by bug 11707.
- Adds a non-fatal `get_form_warnings` check that flags sections whose explicitly-sized columns do not sum to 100, and logs the result via `LOG.warning`.
- Leaves the form loading — 78 shipped definition files (Belgian, Canadian, Danish, etc. censuses) currently violate the sum-to-100 rule without breaking rendering (the `size` field is parsed but never read by the layout code), so escalating this to an error would reject working forms. A separate issue will track re-balancing the existing forms; this PR just makes the problem diagnosable for user-authored `custom.xml` files.

## Dependency

Depends on #821 (bug 11707 fix) — introduces `form_validator.py`. GitHub will display both commits until #821 is merged; this PR will auto-rebase to a single commit afterwards.

## Test plan

- [x] `python3 -m pytest Form/tests/ -v` — 48 pass (7 new unit tests for `get_form_warnings`, 1 new unit test for empty `<forms>`, 1 new unit test for comments-only `<forms>`, 2 new integration tests).
- [x] `TestShippedDefinitionFilesValidate` still green — no shipped file is flagged as an error by the new empty-forms rule.
- [x] Manual: shipped-file load produces no `ErrorDialog`, only WARNING-level log entries for the 78 known offenders.

### DIFF
--- a/Form/form.py
+++ b/Form/form.py
@@ -47,7 +47,11 @@ from gramps.gui.dialog import ErrorDialog
 # Gramps specific
 #
 # ---------------------------------------------------------------
-from form_validator import validate_form_dom, validate_form_element
+from form_validator import (
+    get_form_warnings,
+    validate_form_dom,
+    validate_form_element,
+)
 
 LOG = logging.getLogger(".FormGramplet")
 
@@ -194,6 +198,9 @@ class Form:
                 full_path,
                 "\n".join(errors),
             )
+
+        for warning in get_form_warnings(dom):
+            LOG.warning("In %s: %s", full_path, warning)
 
         try:
             self.__load_definitions(dom)
--- a/Form/form_validator.py
+++ b/Form/form_validator.py
@@ -127,7 +127,8 @@ def validate_form_dom(dom: xml.dom.minidom.Document) -> list[str]:
 
     Checks that:
 
-    * a ``<forms>`` root element exists;
+    * a ``<forms>`` root element exists and contains at least one
+      ``<form>`` definition;
     * each ``<form>`` element has ``id``, ``title`` and ``type`` attributes;
     * each ``<section>`` element has non-empty ``role`` and ``type``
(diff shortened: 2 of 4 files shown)


eduralph:
Converted the tests in this PR from pytest back to stdlib unittest so they match Gramps' own test conventions [AGENTS.md] and can be run upstream without adding pytest as a dependency.

Changes in [6126174e](https://github.com/eduralph/addons-source/commit/6126174e):

- Restored FormLoaderTestCase base class from the bug 11707 PR (#821) and ported the two bug 11010 additions (TestEmptyFormsValidation, TestColumnSizeWarnings) onto it
- pytest fixture + monkeypatch → setUp/addCleanup with unittest.mock.patch.object
- tmp_path → tempfile.mkdtemp + shutil.rmtree cleanup
- caplog → self.assertLogs(".FormGramplet", level=logging.WARNING)
- pytest.importorskip → raise unittest.SkipTest guard
- Updated the invocation hint in test_form_validator.py docstring from python3 -m pytest … to python3 -m unittest …

Verified on Ubuntu 24.04 (host + Docker)

GaryGriffin:
I merged 821, but 822 now has conflicts. Please resolve.

PR #825, opened 2026-04-18
CalculateEstimatedDates: handle ancestry-loop DatabaseError (bug 0007898)

eduralph:
## Summary
Fixes [bug 0007898](https://gramps-project.org/bugs/view.php?id=7898) — the *Calculate Estimated Dates* tool aborts the entire run when `probably_alive_range` raises `DatabaseError` on a family-tree loop (ancestor or descendant cycle), leaving signals disabled and the progress dialog stuck open.

## Changes
- Wrap each per-person iteration across the three loops (removal, selection, apply) with `try/except` so a single bad record is logged via `LOG.warning` and skipped instead of tearing down the tool.
- Wrap each loop with an outer `try/finally` so `db.enable_signals()` and `progress.close()` always run, even on unexpected errors.
- Surface a `Skipped N people due to errors (see log)` message to the user when any rows were skipped.
- Add unit tests covering `get_modifier` branches, `calc_estimates` happy path and `DatabaseError` propagation, and `.gpr.py` registration metadata. The addon module is loaded lazily inside a fixture so pytest collection succeeds even when the GUI stack cannot import.

No behavioural changes for trees without loops — happy-path output is unchanged.

## Test plan
- [x] `python3 -m pytest CalculateEstimatedDates/tests/test_calculate_estimated_dates.py::test_gpr_registration_metadata -v` passes locally
- [ ] Full test file passes in CI
- [ ] Manual verification on a tree with a known ancestry loop: tool completes with a *Skipped* message rather than crashing

### DIFF
--- a/CalculateEstimatedDates/CalculateEstimatedDates.py
+++ b/CalculateEstimatedDates/CalculateEstimatedDates.py
@@ -29,6 +29,7 @@
 # python modules
 #
 #------------------------------------------------------------------------
+import logging
 import time
 
 #------------------------------------------------------------------------
@@ -43,6 +44,7 @@ from gramps.gen.plug.menu import BooleanOption, NumberOption, StringOption, \
 from gramps.gen.lib import (Date, Event, EventType, EventRef, Source,
                             Citation, Note, NoteType)
 from gramps.gen.db import DbTxn
+from gramps.gen.errors import DatabaseError
 from gramps.gen.config import config
 from gramps.gen.display.name import displayer as name_displayer
 from gramps.gen.plug.report.utils import get_person_filters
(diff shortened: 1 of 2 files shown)


GaryGriffin:
Needs rebased as 55 files are changed.

eduralph:
Rebased and removed the versioning bumping

eduralph:
Converted the tests in this PR from pytest to stdlib unittest so they match Gramps' own test conventions [AGENTS.md] and can be run upstream without adding pytest as a dependency.

Changes in [6e6a35fd](https://github.com/eduralph/addons-source/commit/6e6a35fd):

pytest fixtures → unittest.TestCase classes with _make_tool helper
monkeypatch.setattr → mock.patch.object (context-manager form)
pytest.raises(..., match=...) → self.assertRaisesRegex
pytest.importorskip/module-level skips → raise unittest.SkipTest guards
plain assert → self.assertEqual
Pin Gtk/Gdk to 3.0 before gramps imports so systems with GTK4 present don't crash on Gtk.IconSize.MENU during collection

Verified on Ubuntu 24.04 (host + Docker)

PR #826, opened 2026-04-18
ImportMerge: fix AttributeError when adding/merging Tag objects (bug 0014056)

eduralph:
## Summary
Fixes [bug 0014056](https://gramps-project.org/bugs/view.php?id=14056) — *Import and Merge tool* crashes with `AttributeError: 'SQLite' object has no attribute 'has_tag_gramps_id'` when the user selects *Add* (or hits a GID conflict) on a `Tag` row.

`Tag` is a table object and, unlike primary objects, does not have a `gramps_id` field — there is no `has_tag_gramps_id` / `find_next_tag_gramps_id` on the database. The generic `getattr(self.db1, 'has_' + obj_type.lower() + '_gramps_id')` lookup in `do_commits` therefore raised before the commit could complete.

## Changes
- Guard both GID-conflict branches in [`ImportMerge/importmerge.py`](https://github.com/gramps-project/addons-source/blob/maintenance/gramps60/ImportMerge/importmerge.py) (`S_ADD` and `S_DIFFERS`) so they skip `Tag`.
- Bump `version` in `importmerge.gpr.py` (`0.0.25` → `0.0.26`).
- Add an integration test exercising both branches via the real Gramps SQLite backend; the test fails without the guard and passes with it.

Confirmed by upstream maintainers (Nick_H, dstraub) in the bug thread: tags are table objects without a Gramps ID, so the generic lookup is invalid for them.

## Test plan
- [x] `python3 -m pytest ImportMerge/tests/test_integration_importmerge.py -v` passes
- [x] Reverting the guard makes the new tests fail with the original `AttributeError`
- [ ] Manual verification with the steps from the bug report (Gramps Sample tree, sort by Object descending, *Add* a Tag row, *Done*, *Save*) — completes without crash

### DIFF
--- a/ImportMerge/importmerge.py
+++ b/ImportMerge/importmerge.py
@@ -803,8 +803,8 @@ class ImportMerge(tool.BatchTool, ManagedWindow):
                     continue
                 # now check the missing list for a match
                 changed += self.check_miss(r_objtype, r_hndl, item)
-            # check for GID conflict
-            if item.gramps_id != item1.gramps_id:
+            # check for GID conflict (Tag is a table object without gramps_id)
+            if obj_type != 'Tag' and item.gramps_id != item1.gramps_id:
                 if getattr(self.db1, 'has_' + obj_type.lower() +
                            '_gramps_id')(item.gramps_id):
                     item.gramps_id = getattr(self.db1, 'find_next_' +
@@ -823,9 +823,10 @@ class ImportMerge(tool.BatchTool, ManagedWindow):
                     continue
                 # now check the differences list for a match
                 self.check_diffs(r_objtype, r_hndl, obj_type, item)
-            # check for GID conflict
-            if getattr(self.db1, 'has_' + obj_type.lower() +
-                       '_gramps_id')(item.gramps_id):
+            # check for GID conflict (Tag is a table object without gramps_id)
+            if obj_type != 'Tag' and getattr(
+                    self.db1, 'has_' + obj_type.lower() +
+                    '_gramps_id')(item.gramps_id):
                 item.gramps_id = getattr(self.db1, 'find_next_' +
                                          obj_type.lower() +
                                          '_gramps_id')()
--- a/ImportMerge/tests/test_integration_importmerge.py
+++ b/ImportMerge/tests/test_integration_importmerge.py
@@ -38,10 +38,8 @@ Regression coverage:
 import os
 import shutil
 import tempfile
-import types
+import unittest
 from types import SimpleNamespace
-
-import pytest
 
 # The ImportMerge module imports Gtk at module load — skip the whole file if
 # gi/Gtk aren't available (headless-without-GTK environments). On systems
@@ -49,15 +47,14 @@ import pytest
 # import (mirrors what gramps.grampsapp does at startup); otherwise
 # PyGObject loads GTK4 and the gramps.gui import chain crashes on
 # Gtk.IconSize.MENU (a GTK3-only enum).
-pytest.importorskip("gi")
-import gi  # noqa: E402
-
 try:
+    import gi
+
     gi.require_version("Gtk", "3.0")
     gi.require_version("Gdk", "3.0")
-except (ValueError, AttributeError):
-    pytest.skip(
-        "GTK 3.0 typelib not available", allow_module_level=True
+except (ImportError, ValueError, AttributeError) as err:
+    raise unittest.SkipTest(
+        "GTK 3.0 / PyGObject not available: %s" % err
     )
 
 # ------------------------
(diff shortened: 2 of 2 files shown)


GaryGriffin:
no need to bump version.

eduralph:
Undid the versioning

eduralph:
The tests require following

1. Python 3 + pytest
2. Gramps importable as a Python package (python3 -c "import gramps" must work)
3. PyGObject (python3 -c "import gi")
4. GTK 3.0 typelib (gir1.2-gtk-3.0 / gir1.2-gdk-3.0)
5. PYTHONPATH=. from the addons-source repo root (so ImportMerge is importable)

then you can runs the tests like so

PYTHONPATH=. GRAMPS_RESOURCES=$(python3 -c "import os, gramps; print(os.path.dirname(os.path.dirname(gramps.__file__)))") python3 -m pytest ImportMerge/tests/test_integration_importmerge.py -v

PR 820 fulfills all the requirements, so you don't have to take care of that normally. That's why the tests are classified as integration tests you can find at test_integration_importmerge.py. I ran into a gap while testing on my dev machine and will correct that in the tests.

Nick-Hall:
Tests must use the _unittest_ framework — do not use _pytest_.

See the [Testing](https://github.com/gramps-project/gramps/blob/master/AGENTS.md#testing) section in the AGENTS.md file.

eduralph:
Sorry, the agent caches the content, so it doesn't pick up on changes. changed that, should now be conform. I also adjusted it to using unittest.

Pre-reqs are
------------------
- Python 3.10+
- Gramps — installed and importable as a Python package (python3 -c "import gramps" must succeed)
- PyGObject (the Python gi bindings)
- GTK 3.0 typelib — gir1.2-gtk-3.0 and gir1.2-gdk-3.0 on Debian/Ubuntu (ImportMerge is a GUI tool whose module-level from gi.repository import Gtk runs at import time)
- The addons-source repo checked out, with ImportMerge/ on disk

Command is 
-------------------
PYTHONPATH=. GRAMPS_RESOURCES=$(python3 -c "import os, gramps; print(os.path.dirname(os.path.dirname(gramps.__file__)))") python3 -m unittest ImportMerge.tests.test_integration_importmerge -v

I'll now double-check the other PRs

GaryGriffin:
Unittest fails on Mac. Traceback is below. It appears to still want to use pytest, among other issues.

```
[jhbuild]:  % PYTHONPATH=. GRAMPS_RESOURCES=$(python3 -c "import os, gramps; print(os.path.dirname(os.path.dirname(gramps.file)))") python3 -m unittest ImportMerge.tests.test_integration_importmerge -v
  File "<string>", line 1, in <module>
    import os, gramps; print(os.path.dirname(os.path.dirname(gramps.file)))
                                                             ^^^^^^^^^^^
AttributeError: module 'gramps' has no attribute 'file'
test_integration_importmerge (unittest.loader._FailedTest.test_integration_importmerge) ... ERROR

ERROR: test_integration_importmerge (unittest.loader._FailedTest.test_integration_importmerge)
ImportError: Failed to import test module: test_integration_importmerge
  File "/Users/GrampsTest/gtk/inst/lib/python3.13/unittest/loader.py", line 137, in loadTestsFromName
  File "/Volumes/Storage/Development/addons-source/ImportMerge/tests/test_integration_importmerge.py", line 44, in <module>
    import pytest
ModuleNotFoundError: No module named 'pytest'


Ran 1 test in 0.000s


```

eduralph:
Yes, you're right - I'm guessing it also has something to do with the environment I've been testing it in. Let me get a bit more structured and define a proper baseline and run the tests against that.

eduralph:
@Nick-Hall , part of the reason why the system wasn't honoring AGENTS.md was because it's not in the maintainance/gramps60 line. It is in master and maintainance/gramps61 line of gramps though.

eduralph:
Sorry for the churn earlier, @GaryGriffin — the traceback you got was hitting a pre-conversion snapshot where import pytest was still at the top of the file. That was fixed a few commits later in [a37f52ed](https://github.com/gramps-project/addons-source/pull/826/commits/a37f52ed) (ImportMerge: rewrite integration test with unittest framework), and the branch has since also picked up Black formatting (ac81d5f4) and type hints + class header (72281fb9) per the gramps [AGENTS.md](https://github.com/gramps-project/gramps/blob/master/AGENTS.md) conventions. The current tip of the branch has no pytest reference anywhere.

To avoid any messes with any kind of mix-up, I put together a reproducible Docker-based harness at [eduralph/gramps-testbed](https://github.com/eduralph/gramps-testbed). Clean Ubuntu 24.04 run against this PR branch.

Any comments on that are welcomed of course.

PR #832, opened 2026-05-06
Update unittests for 6.1

GaryGriffin:
Update to fix gramps_version (from 6.0 to 6.1) and add Gtk 3.0 check

### DIFF
--- a/CalculateEstimatedDates/tests/test_calculate_estimated_dates.py
+++ b/CalculateEstimatedDates/tests/test_calculate_estimated_dates.py
@@ -250,7 +250,7 @@ class TestGprRegistration(unittest.TestCase):
         self.assertEqual(args, ("TOOL",))
         self.assertEqual(kwargs["id"], "calculateestimateddates")
         self.assertEqual(kwargs["fname"], "CalculateEstimatedDates.py")
-        self.assertEqual(kwargs["gramps_target_version"], "6.0")
+        self.assertEqual(kwargs["gramps_target_version"], "6.1")
         self.assertEqual(kwargs["status"], "STABLE")
         self.assertEqual(kwargs["toolclass"], "CalcToolManagedWindow")
         self.assertEqual(kwargs["optionclass"], "CalcEstDateOptions")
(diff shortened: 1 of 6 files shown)


eduralph (commented):
Thanks @GaryGriffin — picking up the import-order fix and adding the GTK-3 pin makes sense. One issue and one nit, both isolated to `WebSearch/tests/test_filetable.py`; the other five files look good.

### `WebSearch/tests/test_filetable.py` — `unittest` referenced before import

In this file (only) the existing `import unittest` is removed from the top of the file and re-added several lines below the new `gi.require_version` block. Since the except branch raises `unittest.SkipTest`, this means a missing/incompatible `gi` makes the except branch crash with `NameError: name 'unittest' is not defined` instead of cleanly skipping the module:

```python
"""docstring"""

# Comment...
try:
    import gi
    gi.require_version("Gdk", "3.0")
except (ImportError, ValueError, AttributeError) as err:
    raise unittest.SkipTest("GTK 3.0 / PyGObject not available: %s" % err)
    #     ^^^^^^^^ NameError on the first failure path

import os

import sys
import unittest      # ← imported only here
```

Suggested fix — hoist `import unittest` above the `try` block (and keep it grouped with the other stdlib imports):

```python
"""docstring"""

import os
import sys
import unittest

# Comment...
try:
    import gi

    gi.require_version("Gdk", "3.0")
except (ImportError, ValueError, AttributeError) as err:
    raise unittest.SkipTest("GTK 3.0 / PyGObject not available: %s" % err)

sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))

from models import DBFileTableConfig
…
```

The other five files (`Sqlite/tests/test_sqlite.py`, both `TMGimporter/tests/*.py`, `CalculateEstimatedDates/...`, `DataEntryGramplet/...`) all retain their original top-of-file `import unittest`, so they're not affected by this issue — only WebSearch needs the change.

### Nit — comment misattribution

The comment block above the `try` reads:

> The TMGimporter module imports Gtk at module load — skip the whole file if gi/Gtk aren't available …

This is `WebSearch/tests/test_filetable.py`, not TMGimporter. Looks like a copy-paste from one of the TMGimporter test files in the same PR. Either drop the "TMGimporter module" reference or rephrase to something like "WebSearch's test imports `models`/`db_file_table` which transitively …" if there's actually a GTK import path here (I couldn't find one — `models.py`/`db_file_table.py`/`constants.py` look gi-free, so the GTK-pin block may not even be needed for this file).

If the GTK pin block isn't needed in WebSearch, an alternative is to drop it entirely from this file and keep just the `sys.path.insert` prologue. Happy to leave that call to you.

GaryGriffin:
Thanks for the feedback. Fixed as you suggested.
The Gtk pin is needed.

PR #842, opened 2026-05-09
Initial 1926 IE Census and Birth Marriage Death Form

GaryGriffin:
Add IE 1926 Form and reference. Feedback requested.

### DIFF
--- a/Form/form.py
+++ b/Form/form.py
@@ -94,6 +94,7 @@ definition_files = [
     "form_dk.xml",
     "form_fr.xml",
     "form_gb.xml",
+    "form_ie.xml",
     "form_pl.xml",
     "form_us.xml",
     "test.xml",
--- a/Form/form_ie.xml
+++ b/Form/form_ie.xml
@@ -103,4 +103,269 @@
             </column>
         </section>
     </form>
+        <form id='IEBirth' type='Birth' title='Irish Birth Certificate'>
+		<heading>
+            <_attribute>(1.) Entry no.</_attribute>
+        </heading>
+		<heading>
+            <_attribute>Superintendent Registrar's District</_attribute>
+        </heading>
+		<heading>
+            <_attribute>Registrar's District</_attribute>
+        </heading>
+		<heading>
+            <_attribute>District</_attribute>
+        </heading>
+		<heading>
+            <_attribute>Union</_attribute>
+        </heading>
+		<heading>
+            <_attribute>County</_attribute>
+        </heading>
+		<heading>
+            <_attribute> (10.) Signature of Registrar</_attribute>
+        </heading>
+		<heading>
+            <_attribute> (11.) Baptismal Name if added after Registration of Birth, and Date</_attribute>
+        </heading>
+        <section role='Primary' type='person' title='Child'>
+			<column>
+                <_attribute>Name</_attribute>
+				<_longname>(3.) Name (if any)</_longname>
+            </column>
+			<column>
+				<_attribute>Place of Birth</_attribute>
+				<_longname>(2.) Date and Place of Birth</_longname>
+            </column>
+			<column>
+                <_attribute>Date of Birth</_attribute>
+				<_longname>(2.) Date and Place of Birth</_longname>
+			</column>
+            <column>
+                <_attribute>Sex</_attribute>
+				<_longname>(4.) M or F for male or female</_longname>
+				<size>18</size>
+            </column>
+        </section>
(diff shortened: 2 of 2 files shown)


OscarBrownbread:
1. Change "Total Children Under 18 Living" to "Total Children Under 16 Living" (16 years not 18)
2. Change "Street etc with No of House" to "Street etc with No. of House" (add . for consistency)

I submitted a similar form on the bug tracker: 
https://gramps-project.org/bugs/view.php?id=13977

I like how this form separates the age by year and month.

My version tried to include the column labels, (a), (b), (c), etc. Admittedly, this was not implemented well by me and is probably confusing.

GaryGriffin:
> Change "Total Children Under 18 Living" to "Total Children Under 16 Living" (16 years not 18)

I read the form wrong. Fixed

> Change "Street etc with No of House" to "Street etc with No. of House" (add . for consistency)

Good catch.

Updated ie_form with these changes.

GaryGriffin:
It would be good to add the IE Birth/Marriage/Death Forms .

Let me review the B/M/D forms in your bug report and add them here for review. Some of the fields seem mixed between header and per-person section, but that may just be my initial reaction.

GaryGriffin:
Added IE Birth/Marriage/Death forms from [Bug13977](https://gramps-project.org/bugs/view.php?id=13977) . 

Removed the Signed field from Marriage Form. Not sure what its purpose.
Changed the Type to Birth/Marriage/Death (instead of Birth Certificate ..) to align with other form uses.

Did not change any other fields as they made sense in usage. Personally I prefer not to have numbers in _attribute (only in _longname) but I left as is for comments.

Nick-Hall:
> Personally I prefer not to have numbers in _attribute (only in _longname) but I left as is for comments.

I originally used `_attribute` as a standardised descrioption in my original implementation of English census forms.  Since then, people have used the `_attribute` and `_longname` fields differently.  This is not wrong - just how things have developed.

I'm not sure that it is even helpful to translate these any more.  We have had complaints from the translators, and may need to discuss this again.

OscarBrownbread:
Thanks for the improvements. I agree that the "Signed" column can be removed. I added it to record the name as it is on the form as well as the default display name of the person (after choosing a person). For example, my default display name shows the maiden name after choosing a person, whereas married names would be signed on some forms. Similarly, "Patrick Joseph (PJ) Murphy" might be signed "Paddy Murphy". However, now I realise that users can overwrite the normal name field with the signed name on the form. 

My reasoning behind "type=Birth cert." was to prevent overwriting data in the Birth event. I'm unsure about this and created a new type to avoid any possible data loss. Is it possible to overwrite previous Birth info, e.g. the date? Does the form overwrite a previous Birth event? Anyway, in principle I'm fine with this change to type=Birth.

Yes, the labels are still in the Headings as Headings do not allow  _longname as far as I am aware. Perhaps it is better to move those labels to the end instead of the start, e.g. change "(1.)  Entry no." to "Entry no. (1.)". I'm also OK with removing them from Headings and leaving them in the _longname.

Thanks for your feedback and work.

OscarBrownbread:
I propose to add a few more Headings for 1926 census:

`
```
<heading>
    <_attribute>Name of Public Institution</_attribute>
</heading>
<heading>
    <_attribute>Total Area in Statute Acres</_attribute>
</heading>
<heading>
    <_attribute>Signature of Head of Household</_attribute>
</heading>
<heading>
    <_attribute>Signature of Enumerator</_attribute>
</heading>
```
`

The Headings order could be adjusted, depending on how users read the document. I suggest the last three above should be the first Headings. Then "Persons, Males, Females, Rooms", "County", etc., and finally the last Heading should be the first one mentioned above ("Name of Public Institution").

OscarBrownbread:
I created a PR. Let me know if that is helpful or not helpful. 

https://github.com/gramps-project/addons-source/pull/889

GaryGriffin:
Any more issues with this or is it ready to be published?

OscarBrownbread:
1. Minor spacing issue before "        <form id='IEBirth'". It should be just 1 tab in to match the others.
2. Also preceding spaces (twice); remove space within "> (".
I linked to this post from my related issue on Mantis. 

I don't see any other issue and think it's ready to publish if no one else contributes within a day. Thanks! :)

GaryGriffin:
Fixed the issues you mentioned. Please test and let me know if this is final.

OscarBrownbread:
tested and ready for publication.

PR #846, opened 2026-05-10
PDF forms

dsblank:
This addon (PDFForms) provides two plugins:

## Generate PDF Forms (Tool)

A tool under **Tools → Utilities → Generate PDF Forms** that lets you create blank fillable PDF forms from within Gramps. Supports two modes:

- **Census / Event Form** — select any form defined in the Form addon (US, GB, CA, BE, DK, FR, PL, or custom XML), choose the number of data rows, and generate a fillable PDF ready to distribute.
- **Ahnentafel Pedigree Chart** — choose 1–5 generations to produce a pedigree chart PDF with fields for name, birth, marriage, and death for each ancestor.

Settings (last-used form, row count, generations, output path, open-after-generate) are persisted between sessions. The generated PDF can be opened automatically in the system PDF viewer after creation.

## Import Fillable PDF Forms (Import)

Imports a filled-in PDF (generated by this addon) via **File → Import**. Supports both census/event form PDFs (routes data to the Form addon) and pedigree chart PDFs (converts Ahnentafel fields to Gramps persons, families, events, and citations using CSVParser).

The imported PDF is copied to the Gramps media directory and a MediaRef is added to all created objects.

## Sample PDFs

image

image

Example PDFs:

1. [Pedigree5.pdf](https://github.com/user-attachments/files/27816417/Pedigree5.pdf)
2. [USWWIIDRC.pdf](https://github.com/user-attachments/files/27816423/USWWIIDRC.pdf)
3. [USNY1915.pdf](https://github.com/user-attachments/files/27816439/USNY1915.pdf)


GaryGriffin:
Very Interesting
Couple question
1) Mac (and I assume Win) bundles do not include pypdf. 
2) The naming really excited me as I thought it would be a generic PDF importer. Might I suggest you rename the addon ImportAhnentafelPDF or something more specific. This could open the door for many more PDF import addons (descendent tree, for instance). OR Is your intention to add additional template PDFs that this can read? 
3) If pypdf were in all releases, is there more places to leverage (PDF Media thumbnails, for instance)

dsblank:
> 1. Mac (and I assume Win) bundles do not include pypdf.

Will the `requires_mod=["pypdf"]` help in installing? How does it work?

> 2. The naming really excited me as I thought it would be a generic PDF importer. Might I suggest you rename the addon ImportAhnentafelPDF or something more specific. This could open the door for many more PDF import addons (descendent tree, for instance). OR Is your intention to add additional template PDFs that this can read?

I plan on adding more PDF templates (blank PDF files). Pedigree is just the first. I don't know how we could import **any** PDF file.

> 4. If pypdf were in all releases, is there more places to leverage (PDF Media thumbnails, for instance)

image

dsblank:
How about `ImportGrampsPDFForms` ?

GaryGriffin:
> Mac (and I assume Win) bundles do not include pypdf.
> Will the requires_mod=["pypdf"] help in installing? How does it work?
> 
For Mac, the user has to install any additional modules into a specific folder (gramps/gramps60/plugins/lib). This is new for 6.0 . For 5.x, the user had to update into the bundle - not for the weak of heart. 

For Win, there is no solution that I am aware due to signing of bundle. 

Using the requires_mod is the right answer. For both, the Addon Manager will not allow the user to upload the addon without the module. So an alternate is to convince the bundle-creators to add the module into the bundle. If there are more uses for the module, it may become an easier sell.

> The naming really excited me as I thought it would be a generic PDF importer. Might I suggest you rename the addon ImportAhnentafelPDF or something more specific. This could open the door for many more PDF import addons (descendent tree, for instance). OR Is your intention to add additional template PDFs that this can read?
> I plan on adding more PDF templates (blank PDF files). Pedigree is just the first. I don't know how we could import any PDF file.

Agreed that a generic is not feasible. If you are planning on more templates, great. Just trying to avoid giving the user the wrong impression when they see the addon name in the Addon Manager. 

> 
> If pypdf were in all releases, is there more places to leverage (PDF Media thumbnails, for instance)

One of the current complaints is that the preview (thumbnail) of a PDF Media is OS-dependent. On Mac, it doesnt work AFAIK. So when you load a PDF as media, you dont get a thumbnail or preview. Double clicking the preview brings up the native Mac app, Preview. This shows the image. 

Screenshot 2026-05-10 at 10 23 33 AM

dsblank:
> For Mac, the user has to install any additional modules into a specific folder (gramps/gramps60/plugins/lib). This is new for 6.0 . For 5.x, the user had to update into the bundle - not for the weak of heart.

> For Win, there is no solution that I am aware due to signing of bundle.

> Using the requires_mod is the right answer. For both, the Addon Manager will not allow the user to upload the addon without the module. So an alternate is to convince the bundle-creators to add the module into the bundle. If there are more uses for the module, it may become an easier sell.

Sorry, I'm not following. Are you saying that `requires_mod` should be used, but it doesn't automatically install the modules? If so, can we fix this?

GaryGriffin:
> Sorry, I'm not following. Are you saying that requires_mod should be used, but it doesn't automatically install the modules? 

That is correct - requires_mod should be used and does not install the modules (it blocks install if the module is not present).

Neither Win nor Mac include pip in the gramps bundle, and even if it did, I don't know that security issues would allow users to add to this installed bundle. 

@jralls added a folder into the gramps Python path so that on a Mac, users can manually run pip and gramps can leverage this path for additional modules. There is no similar solution for Win AFAIK.

dsblank:
Attempting to fix the **requires_mod** issue: <s>https://github.com/gramps-project/gramps/pull/2303</s> https://github.com/gramps-project/gramps/pull/2308

GaryGriffin:
There were 3 issues in the install using the custom mini-install 6.1 Mac version. 

**Install of 2nd Addon generates error**
in Addon Manager:
Click on Generate PDF Forms --> Install
Then click on Import Fillable PDF Forms and I get an Error popup 
Screenshot 2026-05-17 at 8 28 42 PM

If I install Import Fillable PDF Forms first, it loads. I then restart gramps and try to install Generate PDF Forms. It generated the same error. 

Could it be that both are installing to the same GPR file and the Addon Manager doesn't want to overwrite? 

**Import Fillable PDF Forms**
Addon Manager shows both installed, but only Generate PDF Forms is actually available in Tools --> Utilities. I cannot find Import Fillable PDF Forms. I do see `pypdf` in the newly loaded modules. 

**GPR depends_on**
The GPR has `depends_on=["Form Gramplet"],` but it installed without the Form gramplet being installed. When I ran the Generate PDF Forms the option menu had no forms. Once I loaded the Forms Gramplet and restarted gramps, the option menu had all of the available census and event forms available.

dsblank:
Thanks @GaryGriffin , I will take a look.

dsblank:
This was a good test as it exposed a few things that needed to change:

On Gramps core:

1. depends_on was not added to the install (as you mentioned). The mini-installer fixes that.
2. As multiple addons can be in the same tgz, and are all installed when you install one in the UI, the UI now updates all installed addons

In addons-source make.py (PR on master currently):

1. If an addon has depends_on, it should be included in the requirements.

So these were pre-existing conditions, but glad to clean them up too!

GaryGriffin:
Sorry - I found the Import Fillable PDF in the Import section, where it should be.

GaryGriffin:
Testing Import Fillable PDF. Created an IE1911 form. Tried to Import it. Got the error

Screenshot 2026-05-18 at 9 45 29 AM

dsblank:
Nice catch; thanks for letting me know.

Should be fixed.

GaryGriffin:
This is getting interesting.... 

- I re-tried without the Form gramplet installed and got a good error to install Form gramplet
- I loaded Form gramplet and restarted. Got error below. The Form gramplet was in the list of gramplets, but was not installed in any of the views. 
- Screenshot 2026-05-18 at 10 38 24 AM

- Now I added the Gramplet to the bottom section of the Person view. Got the following error:

> 44433: ERROR: dbloader.py: line 547: Failed to import database.
>   File "/Users/gary/Desktop/Gramps6.1 mini-inst.app/Contents/Resources/lib/python3.14/site-packages/gramps/gui/dbloader.py", line 524, in do_import
>     self.import_info = importer(
>                        ~~~~~~~~^
>         self.dbstate.db,
>         ^^^^^^^^^^^^^^^^
>     ...<5 lines>...
>         ),
>         ^^
>     )
>     ^
>   File "/Users/gary/Library/Application Support/gramps/gramps61/plugins/PDFForms/importpdf.py", line 507, in importData
>     _import_form_data(dbase, fields, user)
>     ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/gary/Library/Application Support/gramps/gramps61/plugins/PDFForms/importformpdf.py", line 226, in _import_form_data
>     _media.set_path(_stored_path)
>                     ^^^^^^^^^^^^
> UnboundLocalError: cannot access local variable '_stored_path' where it is not associated with a value

dsblank:
My bad! Please try again.

dsblank:
> I loaded Form gramplet and restarted. Got error below. The Form gramplet was in the list of gramplets, but was not installed in any of the views.

The updated mini-installer should install it now.

GaryGriffin:
Used the #902 make.py to generate download and listing of PDFForms.
Installed Import Fillable gramplet. It did NOT install Form Gramplet when using custom jralls Mac mini-install version of gramps. It did NOT have the Form Gramplet in the Requires popup. Do I need a more recent (still not yet built) gramps?

After installing Forms Gramplet and restarting gramps, I tried to import PDF. Error
Screenshot 2026-05-18 at 10 38 24 AM

Next I went to the bottom bar and Add Gramplet to add Forms Gramplet. Tried import - Still Error
Restarted gramps and now the import works. 

Had to restart gramps after Addon Manager install of Form Gramplet.
Had to restart gramps after Add Gramplet of Forms Gramplet. 

This needs a Wiki Help Page that is well documented.

dsblank:
> Do I need a more recent (still not yet built) gramps?

Probably as I just updated that a few hours ago.

> Had to restart gramps after Addon Manager install of Form Gramplet.
> Had to restart gramps after Add Gramplet of Forms Gramplet.

I wonder if the restart could be avoided, or if we can add a dialog "Restart now?"

If not, yes, wiki docs!

Nick-Hall:
> I wonder if the restart could be avoided

A restart should not be necessary for Gramplets.

GaryGriffin:
I went step by step to reproduce this, but I was not required to restart after Add Gramplet Forms. I guess I was wrong.

New issue: If I run the Import twice, I get 2 Sources, 2 Events, and 2 copies of each person. I would expect the Events and Persons to be duplicates. I was surprised to see the 2 Sources. The Source was correctly created with the form attribute, so I thought it would leverage and not duplicate.

dsblank:
Good to hear that things are working!

> New issue: If I run the Import twice, I get 2 Sources, 2 Events, and 2 copies of each person. I would expect the Events and Persons to be duplicates. I was surprised to see the 2 Sources. The Source was correctly created with the form attribute, so I thought it would leverage and not duplicate.

I'm not sure about handling that. I think if one were to load a (regular) Gramps Form twice, it would just duplicate the data.

On the other hand, the Pedigree uses the CSV infrastructure, so it is partially a merge.

GaryGriffin:
For the specific case of Sources, it seems that the gramplet should not create a duplicate. Imagine running 5 imports and getting 5 identical Sources, each with an Attribute=Form for the same form. Now the user decides to manually add a Form to a Person. They get a Source Menu that lists all 5 Sources for the specific Form. 

Since the Imported PDF was created against a specific Form that had an ID, I think it is safe to assume there should be only a single Source for the specific Form. Otherwise the user will need to merge 2-at-a-time for each instance.

dsblank:
I'm not following. Do you have example forms that you can attach here?

dsblank:
> Since the Imported PDF was created against a specific Form that had an ID, I think it is safe to assume there should be only a single Source for the specific Form. Otherwise the user will need to merge 2-at-a-time for each instance.

I think I am understanding. Let's consider that on the items to fix. I think this addon is ready for public use.

PR #852, opened 2026-05-10
LinesOfDescendency: import ReportError

eduralph:
## Root cause

`LinesOfDescendency.__init__` looks up two people by Gramps ID and was mishandling the not-found case three ways:

1. `ReportError` was used but never imported — Ruff F821; the `raise` would itself have hit `NameError`.
2. The `if self.descendent == None` guard sat **after** `self.descendent.get_handle()`, so a missing descendent raised `AttributeError` before the guard could run — the guard was unreachable.
3. The `ancestor` lookup had no guard at all — a missing ancestor surfaced later as an unrelated error.

(2) and (3) were raised by @GaryGriffin in review; the first revision of this PR only added the import.

## Fix

```diff
 from gramps.gen.const import GRAMPS_LOCALE as glocale
+from gramps.gen.errors import ReportError
```

```diff
         self.descendent = database.get_person_from_gramps_id(pid)
+        if (self.descendent == None) :
+            raise ReportError(_("Person %s is not in the Database") % pid )
         self.descendent_handle = self.descendent.get_handle()
         ancestor = menu.get_option_by_name('ancestor').get_value()
         self.ancestor = database.get_person_from_gramps_id(ancestor)
-        if (self.descendent == None) :
-            raise ReportError(_("Person %s is not in the Database") % pid )
+        if (self.ancestor == None) :
+            raise ReportError(_("Person %s is not in the Database") % ancestor )
```

The descendent guard moves above `get_handle()`; a symmetric guard is added for `ancestor`. `ReportError` is imported for both raises.

## Verified against

- `gramps-project/gramps` `gramps/gen/errors.py` — `ReportError` is defined there.

## Test

Adds `LinesOfDescendency/tests/test_linesofdescendency_guards.py` (3 tests): a missing descendent or a missing ancestor must raise `ReportError` naming the offending ID; `__init__` completes cleanly when both are present. It drives `__init__` with a mock options/database and stubs `Report.__init__`. The addon module's filename has a hyphen, so the test loads it by file path via `importlib`.

Run with the dotted-path form CI uses:

```
python3 -m unittest LinesOfDescendency.tests.test_linesofdescendency_guards
```

- **Fixed tree**: 3 tests, OK.
- **Pre-fix tree**: fails — `ReportError` was not importable, and with the guard after `get_handle()` a missing descendent raised `AttributeError` rather than `ReportError`.

Also: `ruff check --select=E9,F63,F7,F82 --no-fix --exclude="*.gpr.py" LinesOfDescendency/` passes; `python3 -m py_compile` exits 0.

### DIFF
--- a/LinesOfDescendency/lines-of-descendency.py
+++ b/LinesOfDescendency/lines-of-descendency.py
@@ -97,11 +97,13 @@ class LinesOfDescendency(Report):
         menu = options.menu
         pid  = menu.get_option_by_name('pid').get_value()
         self.descendent = database.get_person_from_gramps_id(pid)
+        if (self.descendent == None) :
+            raise ReportError(_("Person %s is not in the Database") % pid )
         self.descendent_handle = self.descendent.get_handle()
         ancestor = menu.get_option_by_name('ancestor').get_value()
         self.ancestor = database.get_person_from_gramps_id(ancestor)
-        if (self.descendent == None) :
-            raise ReportError(_("Person %s is not in the Database") % pid )
+        if (self.ancestor == None) :
+            raise ReportError(_("Person %s is not in the Database") % ancestor )
 
     def write_path(self, path):
         gen = 1
--- a/LinesOfDescendency/lines-of-descendency.py
+++ b/LinesOfDescendency/lines-of-descendency.py
@@ -27,6 +27,7 @@
 #
 
 from gramps.gen.const import GRAMPS_LOCALE as glocale
+from gramps.gen.errors import ReportError
 try:
     _trans = glocale.get_addon_translator(__file__)
 except ValueError:
--- a/LinesOfDescendency/tests/test_linesofdescendency_guards.py
+++ b/LinesOfDescendency/tests/test_linesofdescendency_guards.py
@@ -1,0 +1,131 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2026  Gramps Development Team
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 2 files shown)


GaryGriffin:
I think the problem is larger. The code is:
```
        menu = options.menu
        pid  = menu.get_option_by_name('pid').get_value()
        self.descendent = database.get_person_from_gramps_id(pid)
        self.descendent_handle = self.descendent.get_handle()
        ancestor = menu.get_option_by_name('ancestor').get_value()
        self.ancestor = database.get_person_from_gramps_id(ancestor)
        if (self.descendent == None) :
            raise ReportError(_("Person %s is not in the Database") % pid )
```
If `self.descendant` were None, an error would be generated 3 lines above it on the `self.descendant.get_handle() `call. Doesn't this check need to be moved up as shown? And a second check for the ancestor?

```
        menu = options.menu
        pid  = menu.get_option_by_name('pid').get_value()
        self.descendent = database.get_person_from_gramps_id(pid)
        if (self.descendent == None) :
            raise ReportError(_("Person %s is not in the Database") % pid )
        self.descendent_handle = self.descendent.get_handle()
        ancestor = menu.get_option_by_name('ancestor').get_value()
        self.ancestor = database.get_person_from_gramps_id(ancestor)
        if (self.ancestor == None) :
            raise ReportError(_("Person %s is not in the Database") % ancestor )
```

eduralph:
Good catch — you're right that the guard was unreachable: `self.descendent.get_handle()` two lines above would raise `AttributeError` first, so the `ReportError` path never fired.

Pushed a follow-up that moves the descendent check above `get_handle()` and adds the symmetric `ancestor` guard you suggested. The `ReportError` import this PR originally added is still needed — both raises use it.

PR #781, opened 2025-12-19
Update PostgreSQLEnhanced to v1.7.3

glamberson:
# Update PostgreSQLEnhanced to v1.7.3

## Summary

Major update to PostgreSQLEnhanced addon with dual backend registration, ConfigManager integration, and Gramps 6.0.6 API compatibility.

### Dual Backend Registration
- **PostgreSQL Enhanced (Monolithic)** - All trees in shared database with table prefixes (recommended)
- **PostgreSQL Enhanced (Separate)** - Individual database per tree (advanced users)

Mode selection now integrated into backend choice - users select the appropriate backend when creating a tree.

### ConfigManager Integration
- Per-tree `settings.ini` files following standard Gramps pattern
- Inherits host/port from Edit → Preferences → Database
- Password handled securely (prompted, not stored)
- Backward compatible with connection_info.txt

### Gramps 6.0.6 API Compatibility
- Implemented `json_extract_expression()` method (required by DBAPI changes from Aug 2025)
- Added `person_name_composite` index for improved performance
- Compatible with Gramps commit ce25f3c01

### Bug Fixes
- Import guard prevents registration when psycopg unavailable
- Fixed config file being ignored due to environment variable logic
- Fixed database auto-creation in separate mode
- Proper initialization of force_mode attribute

## Migration for Existing Users

Two migration scripts provided in scripts/ directory:

1. **migrate_to_dual_backends.py** - Updates backend IDs in database.txt
2. **migrate_to_settings_ini.py** - Migrates connection_info.txt to settings.ini

Users upgrading from v1.5.2 should run both scripts.

## Testing

- Tested on Gramps 6.0.1 and 6.0.6
- Both modes (Monolithic and Separate) verified working
- Migration scripts tested on existing databases
- ConfigManager integration validated

## Documentation

- Updated README.md with dual backend explanation
- Added Windows AIO installation instructions
- Added Flatpak installation instructions
- Comprehensive troubleshooting section
- CHANGELOG.md with complete version history

## Breaking Changes

- Backend IDs changed from `postgresqlenhanced` to `postgresqlenhanced-monolithic` and `postgresqlenhanced-separate`
- Configuration migrated from connection_info.txt to per-tree settings.ini
- Migration scripts handle transition automatically

## Compatibility

- Python 3.9+
- Gramps 6.0+
- PostgreSQL 15+
- psycopg 3.1+

## Related

- Fixes glamberson/gramps-postgresql-enhanced#1
[...]

PR #856, opened 2026-05-10
PostgreSQLEnhanced: import time

eduralph:
## Summary

`PostgreSQLEnhanced/postgresqlenhanced.py:2102` calls `time.time()` when stamping new tree metadata:

```python
db.set_metadata("created", str(time.time()))
```

…but the file never imports `time`. Ruff (F821) flags it:

```
F821 Undefined name `time`
   --> PostgreSQLEnhanced/postgresqlenhanced.py:2102:44
```

Without the import, the tree-creation path raises `NameError` instead of recording the timestamp.

## Fix

```diff
 import logging
 import os
 import re
 import pickle
 import sys
+import time
 from urllib.parse import urlparse, parse_qs
```

## Verification

- `ruff check --select=E9,F63,F7,F82 --no-fix --exclude="*.gpr.py" PostgreSQLEnhanced/` → All checks passed.
- `python3 -m py_compile PostgreSQLEnhanced/postgresqlenhanced.py` exits 0.

### DIFF
--- a/PostgreSQLEnhanced/postgresqlenhanced.py
+++ b/PostgreSQLEnhanced/postgresqlenhanced.py
@@ -40,6 +40,7 @@ import os
 import re
 import pickle
 import sys
+import time
 from urllib.parse import urlparse, parse_qs
 
 # -------------------------------------------------------------------------
@@ -1958,7 +1959,6 @@ class PostgreSQLEnhanced(DBAPI):
                         pass
 
                 if attempt < max_retries - 1:
-                    import time
                     time.sleep(0.01 * (2 ** attempt))  # Exponential backoff
                     continue
                 else:


GaryGriffin:
Don't you also have to remove the `import time` on L1961 (inside a block)

Is this compatible with the large rewrite in [PR781](https://github.com/gramps-project/addons-source/pull/781)

eduralph:
L1961 redundant local import time: removed in the latest push (3a9d05cac); the top-level import this PR adds makes it redundant. Thanks for the catch.

Compatibility with #781: I checked. The version of postgresqlenhanced.py in #781 doesn't have an import time anywhere — not module-level, not local — while still calling time.sleep (its L2020) and time.time() (its L2160). So this PR is complementary to #781, not superseded: if #781 lands first, the same bug is still present and this fix still needs to land; if this lands first, #781 will need to absorb the import when it rebases. Worth flagging on #781 separately so glamberson sees it.

PR #863, opened 2026-05-10
QuiltView: import Surname for child-add path

eduralph:
## Summary

`QuiltView/QuiltView.py:1266` constructs a fresh `Surname()` to attach to a new `Name` when adding a child via the editor, but `Surname` is never imported. Ruff (F821) flags it:

```
F821 Undefined name `Surname`
   --> QuiltView/QuiltView.py:1266:26
```

The file already imports `Person, Family, ChildRef, Name` from `gramps.gen.lib` — extend that import to include `Surname`. Without it, the add-child path raises `NameError` instead of opening the editor.

## Fix

```diff
-from gramps.gen.lib import Person, Family, ChildRef, Name
+from gramps.gen.lib import Person, Family, ChildRef, Name, Surname
```

## Scope

This PR addresses only the `Surname` F821. Two other F821 sites in the same file (`displayer` on lines 1239 and 1241, where the call should likely be `name_displayer.display(...)` rather than a top-level `displayer` reference) are real-bug fixes and are deliberately left out of this surgical lint PR — they need a small behavioural review of their own.

## Verification

- `ruff check --select=E9,F63,F7,F82 --no-fix --exclude="*.gpr.py" QuiltView/` → `Surname` F821 removed; the two unrelated `displayer` F821s remain (out of scope).
- `python3 -m py_compile QuiltView/QuiltView.py` exits 0.

### DIFF
--- a/QuiltView/QuiltView.py
+++ b/QuiltView/QuiltView.py
@@ -56,7 +56,7 @@ from gi.repository import Pango, PangoCairo
 # Gramps Modules
 #
 #-------------------------------------------------------------------------
-from gramps.gen.lib import Person, Family, ChildRef, Name
+from gramps.gen.lib import Person, Family, ChildRef, Name, Surname
 from gramps.gui.views.navigationview import NavigationView
 from gramps.gen.display.name import displayer as name_displayer
 from gramps.gen.errors import WindowActiveError


GaryGriffin:
I do not see how to add a Child. When I use this updated code, the sub-menu for child does not have Add option. The submenu for Parent and Spouse has an Add option. How can I reproduce?

Screenshot 2026-05-15 at 6 01 32 PM

eduralph:
@GaryGriffin  — your menu screenshot is the person menu (Parents/Spouse/Children, plural), which goes through a different add_children_submenu branch than the one that touches Surname. The reference at QuiltView.py:1266 is a real bug, not lint paint.

Where it lives. add_child_to_family at QuiltView.py:1258 — line 1266 is name.add_surname(Surname()). AST-verified: Surname has no binding anywhere in the module (no import, no assignment), single read at 1266:25. Hitting that line raises NameError; the surrounding try only catches WindowActiveError, so the exception propagates and EditPerson never opens.

Where it's wired. The live add_children_submenu is the second definition at L1528 (the first at L1203 is shadowed — that's the dead duplicate #870 removes). The live copy only renders the "Add child to family" item inside the if family: branch. Two call sites:

L1193 in family_menu: passes a non-None family → "Add" item rendered → activates add_child_to_family.
L1427 in person_menu: passes family=None → else branch, no Add item.
Dispatcher at L1018-L1021: PersonNode → person_menu, FamilyNode → family_menu, both on left-click of the canvas.

Why you can't see it. Your screenshot is the person popup (left-clicked a person portrait) — that's L1427, by design no Add in Children. The bug only fires from the family popup. To reproduce on unpatched code: in the quilt canvas, left-click on the small 18×18 square sitting between the parents row and the children row (the FamilyNode, not a parent or child). Open Children → click "Add child to family". Without #863, nothing useful happens — Gramps surfaces its uncaught-exception "Error Report" dialog because NameError: name 'Surname' is not defined fires inside the GLib callback. With #863, EditPerson opens normally.

Verified end-to-end. I built a dogtail/AT-SPI test in my testbed that drives exactly that click path on a minimal 2-parent / 0-child tree. On the unpatched branch: no EditPerson, Error Report dialog appears. Apply this PR's one-line import: EditPerson opens within ~1s. Test discriminates cleanly between the two.

Relationship to #870. Complementary, not overlapping — #870 deletes the dead first add_children_submenu (which carried the displayer F821s); #863 fixes the live path the second copy reaches via family_menu. Either merge order works.

PR #866, opened 2026-05-10
CombinedView: add missing gramps.gen imports

eduralph:
## Summary

`CombinedView/combinedview.py` references four `gramps.gen.*` names in its event-summary block (lines 552-572) without importing any of them. Ruff (F821) flags 8 sites:

```
Undefined name `get_birth_or_fallback`  L552
Undefined name `EventType`              L553, L565
Undefined name `get_date`               L554, L560, L566, L572
Undefined name `get_death_or_fallback`  L564
```

The names live in:

| Module | Name(s) |
|---|---|
| `gramps.gen.utils.db` | `get_birth_or_fallback`, `get_death_or_fallback` |
| `gramps.gen.lib` | `EventType` (extend existing import) |
| `gramps.gen.datehandler` | `get_date` |

Without these, the event-summary path raises `NameError` instead of rendering birth/death rows.

## Fix

```diff
 from gramps.gen.config import config
+from gramps.gen.datehandler import get_date
-from gramps.gen.lib import Family, ChildRef, Person
+from gramps.gen.lib import Family, ChildRef, EventType, Person
+from gramps.gen.utils.db import get_birth_or_fallback, get_death_or_fallback
 from gramps.gui.uimanager import ActionGroup
```

## Verification

- `ruff check --select=E9,F63,F7,F82 --no-fix --exclude="*.gpr.py" CombinedView/` → All checks passed.
- `python3 -m py_compile CombinedView/combinedview.py` exits 0.

### DIFF
--- a/CombinedView/combinedview.py
+++ b/CombinedView/combinedview.py
@@ -28,7 +28,6 @@ Combined View
 # Python modules
 #
 #-------------------------------------------------------------------------
-from html import escape
 from operator import itemgetter
 import pickle
 import logging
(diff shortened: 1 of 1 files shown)


GaryGriffin:
But this is in info_string() (L547-L589) which is never called. 

Isn't a better solution to remove the info_string() routine?

eduralph:
Agreed — `info_string()` is never called anywhere in CombinedView (dead since the addon's initial 2020 commit). Adding imports to satisfy the linter on unreachable code is the wrong fix.

Replaced the import-adding commit with one that removes `info_string()` outright. Its only consumer of `from html import escape` goes away too, so I dropped that import as well. Verified: ruff E9/F63/F7/F82 passes and the F401 unused-import count is unchanged from base.

PR #873, opened 2026-05-10
libaccess: add missing value parameter to Person.gramps_id setter lambda

eduralph:
## Summary

`libaccess/libaccess.py:256` reads:

```python
setters = {
    "handle":    lambda self, value: self.setit("handle", value),
    "gramps_id": lambda self:        self.setit("gramps_id", value),  # ← BUG
}
```

The `gramps_id` lambda accepts only `self` but its body references `value`, which is undefined. Ruff (F821) flags it.

`Object.__setattr__` calls `self.setters[attr](self, value)` with two positional arguments. Setting `person.gramps_id = "I0001"` via the libaccess alias interface therefore raises:

```
TypeError: Person.<lambda>() takes 1 positional argument but 2 were given
```

## Fix

```diff
     setters = {
         "handle": lambda self, value: self.setit("handle", value),
-        "gramps_id": lambda self: self.setit("gramps_id", value),
+        "gramps_id": lambda self, value: self.setit("gramps_id", value),
         }
```

Mirrors the `handle` entry directly above.

## Regression test

Added `libaccess/tests/test_libaccess_setattr.py` with two cases:

- `test_set_gramps_id_via_setattr_does_not_raise` — the buggy site. **Fails on the unfixed tree** with `TypeError: Person.<lambda>() takes 1 positional argument but 2 were given`; passes after the fix.
- `test_handle_setter_remains_functional` — pins down that the fix didn’t accidentally regress the symmetric (already-correct) entry.

The test imports the implementation module via the explicit `from libaccess import libaccess` form to avoid the same package-shadowing trap as gramps bug 0012691 (the addon directory is named `libaccess` and contains both `libaccess.py` and `tests/`, so under dotted-path loading a bare `import libaccess` binds the namespace package, not the implementation module). It also stubs `libaccess._` (the gettext shortcut, addressed by a separate PR) and patches `DbTxn`/`database` to avoid touching a real Gramps backend.

Verified locally via the testbed’s `run-addon-unit.sh libaccess`:
- **Before fix**: TypeError on the lambda invocation in `__setattr__` (FAIL)
- **After fix**: 2 tests, OK

## Verification

- `ruff check --select=E9,F63,F7,F82 --no-fix --exclude="*.gpr.py" libaccess/` → F821 on `value` removed (the other libaccess F821s are addressed in #865, kept out of scope here).
- `./scripts/ubuntu/run-addon-unit.sh libaccess` → 2 tests, OK.

### DIFF
--- a/libaccess/libaccess.py
+++ b/libaccess/libaccess.py
@@ -249,7 +249,7 @@ class Person(Object):
 
     setters = {
         "handle": lambda self, value: self.setit("handle", value),
-        "gramps_id": lambda self: self.setit("gramps_id", value),
+        "gramps_id": lambda self, value: self.setit("gramps_id", value),
         }
 
     def setit(self, attr, value):
--- a/libaccess/tests/test_libaccess_setattr.py
+++ b/libaccess/tests/test_libaccess_setattr.py
@@ -1,0 +1,143 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2026  Gramps Development Team
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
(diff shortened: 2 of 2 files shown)


GaryGriffin:
I dont see the lambda fix in the PR. The diff currently has white space in the __init__ and the added unittest file. No change to libaccess.py

eduralph:
> I dont see the lambda fix in the PR. The diff currently has white space in the **init** and the added unittest file. No change to libaccess.py

Sorry, you are right. Corrected now

I'm rechecking all PRs for any issues

Update: the others should be clean.

PR #877, opened 2026-05-10
RebuildTypes: migrate Py2 / Gramps-3 era imports to Gramps 5+

eduralph:
## Summary

`RebuildTypes/RebuildTypes.py` used Gramps-3 era pre-namespace imports:

```python
from gui.plug import tool
from QuestionDialog import OkDialog
```

Neither resolves on Gramps 5+ (modules now under `gramps.gui.*`). The addon failed plugin registration with `ModuleNotFoundError: No module named "gui"` and was completely unusable.

## Note on the April 19 error message

The April 19 plugin-registration smoke test in `eduralph/addons-source` CI reported a different error for this addon: `RuntimeError: could not create new GType: UndoableEntry (subclass of GtkEntry)`. On a current Py3 / Gramps 5 environment, the actual first error is the `ModuleNotFoundError` above, and **after this import migration the addon imports cleanly with no GType conflict**. The earlier RuntimeError appears to have been a downstream symptom or an environment-specific artifact — likely from a sequencing issue in how the April 19 test environment loaded modules.

If a GType conflict does surface in some other environment, that is a separate gramps-core issue (cf. [gramps-project/gramps#2299](https://github.com/gramps-project/gramps/pull/2299) for the analogous `gui.clipboard` headless-import fix).

## Fix

```diff
-from gui.plug import tool
-from QuestionDialog import OkDialog
+from gramps.gui.plug import tool
+from gramps.gui.dialog import OkDialog
```

## Regression test

Added `RebuildTypes/tests/test_rebuildtypes_imports.py` that imports the module via the explicit submodule path (addon dir and impl module share the name — namespace-package shadowing, same trap as libaccess; see gramps bug 0012691 family). Asserts the `RebuildTypes` class is a `Tool` subclass.

Verified via `./scripts/ubuntu/run-addon-unit.sh RebuildTypes`:
- **Before fix**: `ModuleNotFoundError: No module named "gui"` at line 31 (FAIL)
- **After fix**: 1 test, OK — no GType conflict surfaces.

### DIFF
--- a/RebuildTypes/RebuildTypes.py
+++ b/RebuildTypes/RebuildTypes.py
@@ -28,8 +28,8 @@
 # GRAMPS modules
 #
 #-------------------------------------------------------------------------
-from gui.plug import tool
-from QuestionDialog import OkDialog
+from gramps.gui.plug import tool
+from gramps.gui.dialog import OkDialog
 
 from gramps.gen.const import GRAMPS_LOCALE as glocale
 try:
--- a/RebuildTypes/tests/test_rebuildtypes_imports.py
+++ b/RebuildTypes/tests/test_rebuildtypes_imports.py
@@ -1,0 +1,90 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2026  Gramps Development Team
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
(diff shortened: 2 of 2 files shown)


GaryGriffin:
This addon gpr has the following:
```
include_in_listing=False,
status=UNSTABLE,
authors=["Nick Hall"],
category=TOOL_DBFIX,
```
so it is not set to be published for users.
@Nick-Hall do you want this module updated and listed? 

Seeing TOOL_DBFIX and UNSTABLE together is scary for a released addon. 
But merging this PR will make progress on the passing-lint step.

eduralph:
If you let me know what the reason for the "unstable" flag, I'm happy to fix it. Or remove it, as you like.

GaryGriffin:
I am guessing that @Nick-Hall  set to UNSTABLE as it has not been ported to gramps5/python3. We should leave alone at this point. Need guidance from him on what to do with the multiple addons that seem to need porting to current gramps/python version. Then we can work thru the go-forward strategy. I can imagine some will be dropped and some will be updated. 

Reports that are UNSTABLE don't really worry me much (worst case = crash gramps). But tools that can destroy a db are another matter.

Nick-Hall:
If `include_in_listing=False` is specified then the addon should not be easily made available to the public.  If an addon has the status of "Unstable" then it may be untested and if it writes data, then it could potentially cause data corruption.

The _RebuildTypes_ addon was either just an example or written for a particular person to fix a bug.  An export followed by an import will achieve the same result.  The _Type Cleanup_ tool may also be a better option.  I am happy for this addon to be deleted.

PR #878, opened 2026-05-11
EditExifMetadata: declare Pillow and GExiv2 dependencies

eduralph:
## Summary

`EditExifMetadata/editexifmetadata.py` imports two unrelated external dependencies at module load:

```python
L31: from PIL import Image
L44: gi.require_version('GExiv2', '0.10')
L45: from gi.repository import GExiv2
```

…but `editexifmetadata.gpr.py` declared neither `requires_mod` nor `requires_gi`. Result: neither Gramps' Addon Manager (when installing on behalf of a user) nor the auto-derive CI step in addons-source's PR 820 installs Pillow, and the typelib search can't communicate the GExiv2 requirement either. Plugin registration fails with:

```
Plugin error (from 'editexifmetadata'): No module named 'PIL'
```

(The PIL error fires first because line 31 precedes the GExiv2 imports; once Pillow is installed the GExiv2 typelib error would follow next.)

The April 19 plugin-registration smoke test in `eduralph/addons-source` CI flagged this as one of ten addon load failures.

## Fix

```diff
     authors_email=["robhealey1@gmail.com", "paulr2787@gmail.com"],
     navtypes=["Media"],
+    requires_mod=["Pillow"],
+    requires_gi=[("GExiv2", "0.10")],
 )
```

`Pillow` is the PyPI package name; the importable name is `PIL`. The `requires_gi` format matches the convention used by sibling addons (`GeoAncestor`, `GeoTimeLines`, `GraphView`, `PlaceCoordinateGramplet`).

## Note on the CI image

`requires_gi` declares the **GObject Introspection** typelib requirement, but unlike `requires_mod` (pip packages) typelibs come from **system apt packages** — `gir1.2-gexiv2-0.10` on Debian/Ubuntu. The addons-source CI image (Dockerfile in PR 820) needs that apt package added in parallel; tracked separately as a follow-up commit to PR 820.

## Note on PR scope

An earlier revision of this PR only declared `Pillow`. The follow-up sweep for the GExiv2 typelib (which fires next once `Pillow` is installed) was missed. Force-pushed to declare both deps in one coherent change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #880, opened 2026-05-11
PhotoTaggingGramplet: declare GExiv2 in requires_gi

eduralph:
## Summary

`PhotoTaggingGramplet/PhotoTaggingGramplet.py:49` does `from gi.repository import GExiv2`, but `PhotoTaggingGramplet.gpr.py` declared no `requires_gi`. Gramps' Addon Manager and the addons-source CI auto-derive step therefore had no way to surface the missing typelib; on systems without GExiv2 the addon fails plugin registration with the unhelpful:

```
Plugin error (from "PhotoTaggingGramplet"): cannot import name GExiv2, introspection typelib not found
```

Unlike `EditExifMetadata` on gramps60 (which hard-pins `gi.require_version('GExiv2', '0.10')` in code), `PhotoTaggingGramplet.py` does a plain unpinned `from gi.repository import GExiv2`. Any installed GExiv2 typelib works for the addon's own import path — so the declaration only has to express "this addon needs *some* GExiv2 typelib."

`gen/utils/requirements.py:check_gi` accepts a comma-separated version list and matches the dep if any listed version satisfies `gi.require_version`. That mechanism is what makes a single declaration cover the GExiv2 versions in circulation across distros / Flatpak / Snap / Mac AIO / Windows AIO.

## Fix

```diff
     navtypes=["Media"],
     help_url="Addon:Photo_Tagging_Gramplet",
     include_in_listing=True,
+    requires_gi=[("GExiv2", "0.10,0.12,0.14,0.16")],
 )
```

Version list compiled from currently-shipping GExiv2 typelibs:
- `0.10` — gramps60 Linux distributions
- `0.12`, `0.14` — newer Linux distributions
- `0.16` — Mac AIO bundle (confirmed by @GaryGriffin on Mac 6.1 in this PR)

## History on this PR

The original revision declared `requires_gi=[("GExiv2", "0.10")]`, mirroring EditExifMetadata #878. That was too strict — PhotoTaggingGramplet doesn't pin a version in code, so a hard `0.10` declaration rejected every system carrying only a newer typelib. Force-pushed to the multi-version form. Gary's review then surfaced that GExiv2 0.16 was missing from the list — added in `2590e94a7`.

## Note on the CI image

`requires_gi` declares the **GObject Introspection** typelib requirement; typelibs come from system apt packages, not pip. The addons-source CI image (Dockerfile in PR 820) carries `gir1.2-gexiv2-0.10` already; that's sufficient for the auto-derive check to pass on this addon.

## Note on the gramps61 follow-up

This declaration covers the addon's needs across the GExiv2 versions in circulation today, but it requires another list entry whenever a new minor GExiv2 ships. The structurally correct fix is to port @prculley's PR #829 pattern (dynamic version detection via `Repository.enumerate_versions("GExiv2")` then `gi.require_version` of the latest) into `PhotoTaggingGramplet.py` itself — but that's a code change in the addon, not a declaration change, and matches the gramps61 idiom where the addon code adapts to the installed typelib and the `.gpr.py` declares no `requires_gi` at all (cf. EditExifMetadata #890). That follow-up is being raised separately on the gramps61 branch.

## Verification

- `python3 -c "import ast; ast.parse(open('PhotoTaggingGramplet/PhotoTaggingGramplet.gpr.py').read())"` — parses OK.
- `gen/utils/requirements.py:check_gi` (lines 62-80) splits the version field on `,` and accepts if any listed version satisfies `gi.require_version` — verified against the current upstream source.
- End-to-end: once Gramps loads with any of the four declared GExiv2 typelib versions installed, `from gi.repository import GExiv2` resolves and PhotoTaggingGramplet registers cleanly. The plugin-registration smoke test in addons-source CI is the natural gate; this PR doesn't change CI image apt packages, only the declared version list.

### DIFF
--- a/PhotoTaggingGramplet/PhotoTaggingGramplet.gpr.py
+++ b/PhotoTaggingGramplet/PhotoTaggingGramplet.gpr.py
@@ -39,5 +39,5 @@ register(
     navtypes=["Media"],
     help_url="Addon:Photo_Tagging_Gramplet",
     include_in_listing=True,
-    requires_gi=[("GExiv2", "0.10,0.12,0.14")],
+    requires_gi=[("GExiv2", "0.10,0.12,0.14,0.16")],
 )
--- a/PhotoTaggingGramplet/PhotoTaggingGramplet.gpr.py
+++ b/PhotoTaggingGramplet/PhotoTaggingGramplet.gpr.py
@@ -39,4 +39,5 @@ register(
     navtypes=["Media"],
     help_url="Addon:Photo_Tagging_Gramplet",
     include_in_listing=True,
+    requires_gi=[("GExiv2", "0.10,0.12,0.14")],
 )


GaryGriffin:
Mac 6.1 version prereq lists:

> Python Image Library (PIL) is needed for cropping images and LaTeX output.
>  • (PILLOW 12.2.0)
>  • GExiv2 : 0.16 (Exiv2 library : exiv2 executable not installed, can't retrieve libexiv2 version.)

The current PR change is:
`requires_gi=[("GExiv2", "0.10,0.12,0.14")],`
which will not work.

GaryGriffin:
Just to be sure I understand, this fix should be cherry picked to gramps61, right. And there is additional logic added to gramps61 in PR891.

And at some future time, we will probably have to add more strings to the GPR for later versions of GExiv2. Is there a strategy for future version-proofing changes?

eduralph:
> Just to be sure I understand, this fix should be cherry picked to gramps61, right. And there is additional logic added to gramps61 in PR891.
> 
> And at some future time, we will probably have to add more strings to the GPR for later versions of GExiv2. Is there a strategy for future version-proofing changes?

This PR adresses gramps 6.0 and fixes the immidiate issue you raised with using GExiv2 0.16 by adding it to the list. You are right in the sense that it isn't future proof.

PR 891 fixes it more permanent in gramps 6.1 following the same approach used in PR #829 with  EditExifMetadata by jrails;

GaryGriffin:
Sorry, Still unclear to me. Do you want this PR cherry picked to 6.1?

PR880 adds v 0.16 in the GPR. That value is used in Mac 6.1 gramps. So this leads me to think that you want this PR cherry picked to 6.1

PR891 inserts logic in the addon (not the GPR) that checks on GExiv2 installed. With this, you dont need PR880. Anyone can install with the Addon Manager. The addon will not have a requires_gi but will instead check at runtime. 

My sense is below. Please confirm.

- PR880 is for 6.0 only (and you don't need to add 0.16 as allowable)
- PR891 is for 6.1 only (and the GPR will not test existence of GExiv2)

eduralph:
> Sorry, Still unclear to me. Do you want this PR cherry picked to 6.1?

Sorry for not being clearer - no, don't cherry pick it for 6.1 

> * PR880 is for 6.0 only (and you don't need to add 0.16 as allowable)
I was under the assumption that you needed 0.16 for the Mac in 6.0; if not, I'll remove that one. 

> * PR891 is for 6.1 only (and the GPR will not test existence of GExiv2)
Correct, it dynamically detects it just like #829 and loads it that way resolving the forward looking versioning

PR #882, opened 2026-05-11
DNA: remove trailing whitespace

eduralph:
## Summary

Pure whitespace cleanup; no semantic change. Removes **79** trailing-whitespace line(s) across **1** file(s) inside `DNA/` flagged by addons-source CI's Lint job:

```
git --no-pager grep -n --full-name '[ \t]$' -- '*.py'
```

Generated by `sed -i 's/[ \t]\+$//' $(find DNA -name '*.py')`. After this PR, the addon's directory contributes 0 trailing-whitespace lines to the Lint job's scan.

### DIFF
--- a/DNA/dnasegmentmap.py
+++ b/DNA/dnasegmentmap.py
@@ -84,11 +84,11 @@ CONFIG.init()
 draw_single_chromosome = False
 current_chromosome = '1'
 #
-# draw_grandparent_color is a quick and dirty pass at visual phasing to the grandparent. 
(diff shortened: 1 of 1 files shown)


eduralph:
Heads-up that the commit message here is also affected by the same buggy BRE regex as #883/#885 — it claims "79 instances" but the real PCRE count is 55 (which matches the diff). Same root cause; PR #820 commit 205b21c fixes the regex. Diff is otherwise the correct minimal cleanup.

PR #890, opened 2026-05-14
EditExifMetadata: declare Pillow dependency (gramps61)

eduralph:
## Root cause

`EditExifMetadata/editexifmetadata.py` imports `from PIL import Image` at module load (line 31), but `editexifmetadata.gpr.py` declares no `requires_mod`. So neither Gramps' Addon Manager nor the addons-source CI auto-derive step installs Pillow, and plugin registration fails with `No module named 'PIL'`.

## Fix

```diff
     navtypes=["Media"],
+    requires_mod=["Pillow"],
 )
```

## Relationship to #878 — and why this is a separate PR, not a cherry-pick

#878 declares the same dependency on `maintenance/gramps60`, where it *also* adds `requires_gi=[("GExiv2", "0.10")]`. That GExiv2 pin is correct on gramps60 (the code there hard-pins `gi.require_version('GExiv2', '0.10')`) but **wrong on gramps61**: PR #829 reworked `editexifmetadata.py` on gramps61 to load whatever GExiv2 typelib is installed —

```python
v_array = repository.enumerate_versions("GExiv2")
...
gi.require_version("GExiv2", v_array[-1])
```

`gen/utils/requirements.py` checks `requires_gi` with an exact `gi.require_version` match, so a static `("GExiv2", "0.10")` declaration would report the dependency as unmet on any system with a newer GExiv2 — the opposite of #829's intent. GExiv2 is therefore deliberately **left out** of this PR; the addon's own code already raises a clear `ValueError("GExiv2 is not installed")` when the typelib is absent.

So #878 cannot be cherry-picked here as-is — `requires_mod=["Pillow"]` is the portable part, and that is all this PR adds.

## Verified against

- gramps-project/addons-source@`maintenance/gramps61`:`EditExifMetadata/editexifmetadata.py:31` (`from PIL import Image`) and `:45-49` (PR #829 dynamic GExiv2 version handling)
- gramps-project/gramps:`gramps/gen/utils/requirements.py` — `check_gi` / `_test_gi` (exact `gi.require_version` match)

## Test

No automated test: `.gpr.py` files are plugin-registration metadata, exercised only by Gramps' loader. The change is verifiable by inspection against the import in `editexifmetadata.py` cited above.

### DIFF
--- a/EditExifMetadata/editexifmetadata.gpr.py
+++ b/EditExifMetadata/editexifmetadata.gpr.py
@@ -44,5 +44,5 @@ register(
     authors=["Rob G. Healey", "Paul Culley"],
     authors_email=["robhealey1@gmail.com", "paulr2787@gmail.com"],
     navtypes=["Media"],
-    requires_mod=["Pillow"],
+    requires_mod=["PIL"],
 )
--- a/EditExifMetadata/editexifmetadata.gpr.py
+++ b/EditExifMetadata/editexifmetadata.gpr.py
@@ -44,4 +44,5 @@ register(
     authors=["Rob G. Healey", "Paul Culley"],
     authors_email=["robhealey1@gmail.com", "paulr2787@gmail.com"],
     navtypes=["Media"],
+    requires_mod=["Pillow"],
 )


GaryGriffin:
I thought Pillow was part of all grtamps installs. What does the PrerequisitesChecker (in the Optional section) state? On a Mac it is installed in the bundle. (as PIL, not Pillow) On Win is should be installed in the bundle as it is needed for multiple addons. I dont know about Linux versions.

**MAC:**
If you add the mod_requires(Pillow) to the gpr, then Mac cannot install it as the Addon Manager does not find it. Changing to mod_requires(PIL) and it meets requirements and can install. 

Travis:
A comment in DynamicWeb test states below. Is this the reason that CI fails but gramps works?
` # (commented out because PIL not installed in Travis environment)`

Nick-Hall:
> I dont know about Linux versions.

It is installed with the Flatpak and Snap, but optional for distribution packages.

jralls:
> On a Mac it is installed in the bundle. (as PIL, not Pillow)

[Nope](https://github.com/gramps-project/gramps/blob/7dc332d304a879bb1c93cafcea86805cb5e85459/mac/gramps.modules#L168).

You're confusing the package name, Pillow, with the import symbol, `PIL`. They're different because there is a [PIL package in PyPI](https://pypi.org/project/PIL/) that's no longer maintained. Pillow exports a compatible API so it reuses the import symbol.

GaryGriffin:
I am confused. 
In the 6.1 Mac bundle, there is a file `lib/python3.14/site-packages/PIL`
In the 6.0 Mac bundle, there is a file `lib/python3.13/site-packages/PIL`
In both PrerequisitesChecker and EditExifMetadata code, there is : `import PIL`

Should the EditExifMetadata GPR file have a `mod_requires(PIL)`? 
PrerequisitesChecker has the import in a try..except so should not require it.

eduralph:
The disagreement

GaryGriffin: empirically requires_mod=["Pillow"] fails Mac's Addon Manager dep check; ["PIL"] passes.
jralls: "Pillow" is the PyPI package name; "PIL" is the import symbol they reuse for compat.
Nick-Hall: Pillow is bundled with Flatpak/Snap, optional in distros.

What the field actually does — two consumers, two contracts:

- Gramps Addon Manager check	requirements.py:50-60 — find_spec(module)	importable name
- Gramps Addon Manager install	_windows.py:294-302 — subprocess(['pip', 'install', package])	pip-install name
- (PR 820) addons-source CI auto-derive	ci.yml:178-191 — pip install "$mod"	pip-install name 

For Pillow these names differ (importable PIL, pip-install Pillow); for every other addon currently using requires_mod in the repo (litellm, svgwrite, psycopg2, boto3, dbf, networkx, pygraphviz, life_line_chart) they happen to coincide, which is why this is the first PR to hit the conflict.

So Gary is right about the observable behaviour: gramps' gramps/gen/utils/image.py:126 does import PIL.Image, and find_spec("Pillow") returns None even on a system with Pillow installed. PrerequisitesCheckerGramplet at line 1225 also does import PIL. The PR as written declares the dep in a form that the runtime gate will always reject — i.e. it breaks every Pillow-installed system in exchange for telling pip the right name on the (opt-in, default-off, behavior.addons-allow-install-gated) auto-install path.

jralls is right that "Pillow" is the canonical install name, but his comment didn't engage with Gary's empirical test result — and on a find_spec-based check, his "Pillow exports a compatible API so it reuses the import symbol" is exactly the reason ["Pillow"] is the wrong string for that gate.

Options

1. Change PR #890 to requires_mod=["PIL"] — fixes the check for the common case (Pillow already installed). Auto-install path becomes wrong (pip install PIL → the legacy stub), but auto-install is opt-in and gated, and matches the existing pattern of import-names elsewhere in the repo. Smallest defensible diff.
2. Close PR #890 as not-fixable-locally — argue this is a gramps core bug: requires_mod overloads two names that aren't always equal, and the fix belongs in requirements.py (accept tuples, or maintain a known-alias table for at least Pillow→PIL). The addon stays broken on barebones distros until that lands.
3. Hybrid: open a tiny core PR adding a {"Pillow": "PIL"}-style alias in check_mod, then declare ["Pillow"] here. Cleanest, but two PRs.

My suggestion would be to do Option 1 now and open another PR for gramps to fix that bug.

Nick-Hall:
I agree with Option 1.

eduralph:
@Nick-Hall - implemented Option 1

PR #892, opened 2026-05-16
PostgreSQLEnhanced: remove trailing whitespace

eduralph:
## Root cause
294 lines across 5 files in `PostgreSQLEnhanced/` end with trailing
whitespace, which the testbed CI Lint job flags via
`git --no-pager grep '[ \t]$' -- '*.py'`.

## Fix
Strip the offending trailing whitespace from `concurrency.py`,
`postgresqlenhanced.py`, `schema.py`, `schema_migrations.py`, and
`undo_postgresql.py`. Pure formatting change.

## Verified
- `git diff -w upstream/maintenance/gramps60..HEAD` = 0 lines (proves
  the diff is whitespace-only)
- per-hunk audit: every `-` line equals the `+` line plus `[ \t]+$`
- `ast.parse()` clean on every touched file
- post-strip grep `[ \t]+$` on `PostgreSQLEnhanced/*.py` = 0 matches

### Multi-line string content audit
Tokenized every changed file before and after, compared every `STRING`
token. 38 multi-line string literals had content modified:
- **36 docstrings** (function/class/method `"""..."""`) — these are
  consumed only by `help()` and Sphinx autodoc, both of which ignore
  trailing whitespace inside RST.
- **2 SQL queries** passed to `cur.execute("""...""")`:
  - `postgresqlenhanced.py:2138-2143` — `SELECT DISTINCT ... FROM
    pg_tables WHERE tablename LIKE 'tree_%_metadata'`
  - `undo_postgresql.py:247-252` — `INSERT INTO {} ... RETURNING
    change_id` (wrapped in `sql.SQL(...).format(...)`)

  PostgreSQL is whitespace-insensitive between SQL tokens, so trailing
  spaces inside these heredocs have no semantic effect.

Grep over `PostgreSQLEnhanced/` for `assertEqual`/`assertIn`/`assertRegex`
returned 0 hits — no test references any of these strings as data.

## Test
No regression test added — whitespace removal at end-of-line has no
functional behaviour to assert against, and the audits above prove the
patch is content-neutral.

Manual repro:
```
git fetch origin && git checkout PostgreSQLEnhanced-cleanup
git --no-pager grep '[ \t]$' -- 'PostgreSQLEnhanced/*.py'  # prints nothing
git diff -w upstream/maintenance/gramps60..HEAD             # empty
```

### DIFF
--- a/PostgreSQLEnhanced/concurrency.py
+++ b/PostgreSQLEnhanced/concurrency.py
@@ -65,7 +65,7 @@ class ConcurrencyError(Exception):
 class PostgreSQLConcurrency:
     """
     PostgreSQL concurrency control features with graceful fallbacks.
-    
+
     Provides advanced concurrency features that are safe to use
     regardless of PostgreSQL version or capabilities.
     """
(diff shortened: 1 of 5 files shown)


eduralph:
@GaryGriffin - 1/10 whitespace clean up PRs popping up in linting. After this, there should be no issues appearing if #820 gets implemented.

PR #902, opened 2026-05-18
Emit depends_on as "do" field in addon catalog listings

dsblank:
When a gpr.py declares `depends_on`, the build system now emits a `"do"` key in the catalog JSON entry for that plugin. The Gramps Addon Manager (in `gramps/gui/plug/_windows.py`) reads this field when the user clicks Install and automatically installs any unregistered dependencies first.

Both the `as-needed` and `listing` make.py commands are updated.

## Example

```python
# PDFForms/PDFForms.gpr.py
register(TOOL, id="generate_pdf_forms", depends_on=["Form Gramplet"], ...)
```

Produces a catalog entry:
```json
{"i": "generate_pdf_forms", "do": ["Form Gramplet"], ...}
```

Installing `generate_pdf_forms` will then automatically install the Form Gramplet if it is not already present.

### DIFF
--- a/make.py
+++ b/make.py
@@ -721,6 +721,8 @@ elif command == "as-needed":
                             plugin["h"] = p["help_url"]
                         if "audience" in p:
                             plugin["a"] = p["audience"]
+                        if "depends_on" in p:
+                            plugin["do"] = p["depends_on"]
                         listings[lang].append(plugin)
                         if lang == "en":
                             print("Listed:          %s" % p["name"])
@@ -968,6 +970,8 @@ elif command == "listing":
                             plugin["h"] = p["help_url"]
                         if "audience" in p:
                             plugin["a"] = p["audience"]
+                        if "depends_on" in p:
+                            plugin["do"] = p["depends_on"]
                         listings.append(plugin)
                     else:
                         print("   ignoring '%s'" % (p["name"]))


dsblank:
@GaryGriffin let me know if this should target a different branch.

GaryGriffin:
I never touch `master` branch. This would need to be in gramps61.

Should the Addon Manager also list the Form Gramplet in the Requires button so that the user knows that it will also be silently installed as a dependency (like it lists the other requires_mod and requires_gi)?

And this is even more complicated as the user could have multiple Projects with the same named Addon. Which would it use?

dsblank:
> I never touch master branch. This would need to be in gramps61.

Ok, will do.

> Should the Addon Manager also list the Form Gramplet in the Requires button so that the user knows that it will also be silently installed as a dependency (like it lists the other requires_mod and requires_gi)?

Probably. But I think that is out-of-scope for the bug fix.

> And this is even more complicated as the user could have multiple Projects with the same named Addon. Which would it use?

This shouldn't be a problem because the require_mod uses IDs, not names. (This one is a bit confusing because the name and ID are the same).

dsblank:
> Should the Addon Manager also list the Form Gramplet in the Requires button so that the user knows that it will also be silently installed as a dependency (like it lists the other requires_mod and requires_gi)?

I went ahead and fixed that.

PR #904, opened 2026-05-19
PDFForms: fix oversized fields and hidden headings for wide census forms

dsblank:
## Summary

Forms with many narrow columns (e.g. CA1851EW-A has 42 columns, 22 of them `size=1`) caused the page to expand to ~5700 pt (≈ 2 m wide) via the old formula `MIN_COL_W * sum(sizes) / min(size)`.  This made every large column enormous and pushed all but the first heading column per row off-screen, so only 3 of the 7 CA1851EW-A headings were visible.

- Add `MAX_COL_W = 100 pt` (≈ 20 chars at 8 pt Helvetica) as an upper bound on any single field.
- Replace `_required_avail_w` with a sum of per-column clamped widths so the page only expands enough to give every column between `MIN_COL_W` and `MAX_COL_W`.
- Replace `_col_widths` with an iterative clamp that redistributes space from fixed (clamped) columns to the remaining free ones, keeping all widths in `[MIN_COL_W, MAX_COL_W]`.
- For heading fields, use `base_avail` (A4/landscape width) instead of the expanded `avail_w`, and cap each heading field at `MAX_COL_W`, so all heading fields stay on-screen regardless of the data section width.

**Result for CA1851EW-A:** page shrinks from ~5719 pt to ~1651 pt; all 7 headings visible; no field wider than 100 pt.  Simple forms (e.g. US1790) remain standard A4 landscape.

## Test plan

- [ ] Generate CA1851EW-A PDF — page ≈ 1651 pt wide, all 7 heading fields visible (Schedule, EnumDistrict, Township, County, Comprising, OfTheSaid, Enumerator), no data column wider than 100 pt
- [ ] Generate a simple form (e.g. US1790) — page stays standard A4 landscape (842 pt), no regression
- [ ] Import a filled CA1851EW-A PDF back into Gramps — heading values populate the Headings tab and Reference field correctly

### DIFF
--- a/PDFForms/generate_pdf.py
+++ b/PDFForms/generate_pdf.py
@@ -199,28 +199,65 @@ def _sanitize(text):
 
 
 MIN_COL_W = 20   # pt — minimum usable column width
+MAX_COL_W = 100  # pt — maximum column width (≈ 20 chars at 8 pt Helvetica)
 
 
 def _col_widths(columns, available_w):
-    """Return a list of point widths proportional to each column's <size>."""
+    """
+    Return column widths fitting *available_w*, clamped to [MIN_COL_W, MAX_COL_W].
+    Excess/deficit from clamped columns is redistributed iteratively among the rest.
+    """
     sizes = [c["size"] for c in columns]
     total = sum(sizes)
     if total > 0 and all(s > 0 for s in sizes):
-        return [available_w * s / total for s in sizes]
+        widths = [0.0] * len(sizes)
+        fixed = set()
+        for _ in range(len(sizes)):
+            free = [i for i in range(len(sizes)) if i not in fixed]
+            if not free:
+                break
+            fixed_sum = sum(widths[i] for i in fixed)
+            free_avail = available_w - fixed_sum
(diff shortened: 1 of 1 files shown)


GaryGriffin:
I created an SC1891 form and filled it out. In the PDF viewer, the last field heading was clipped. I see WindowR but it should be WindowRooms

Image before adding Person

Screenshot 2026-05-20 at 12 45 49 PM

Image after adding Person
Screenshot 2026-05-20 at 12 46 15 PM

Image from Mac PDF viewer (Preview). There is about 1 inch border on the page. The relative spacing of the columns does not match the gramps view (not sure that it should, just noticed):
Screenshot 2026-05-20 at 12 46 44 PM

dsblank:
I'll see if I can get the entire label in :eyes:

dsblank:
@GaryGriffin Thanks for catching that. Looks better I think:

image

dsblank:
> The relative spacing of the columns does not match the gramps view (not sure that it should, just noticed)

No, it doesn't need to match.

PR #908, opened 2026-05-22
Add HasTagSubstr filter addon for all object namespaces

dsblank:
## Summary

Adds a new third-party filter addon `HasTagSubstr` with nine filter rules — one per Gramps object namespace — that match objects whose tag names contain a given substring (case-insensitive).

**Motivation**: Users with many tags often use a naming convention (e.g. all indirect-relative tags end in 'x'). This addon lets them filter by partial tag name rather than requiring an exact match.

## Implementation

- `hastagsubstr.py` — `HasTagSubstrBase` base class + 9 per-namespace subclasses (Person, Family, Event, Place, Source, Citation, Repository, Media, Note)
- `hastagsubstr.gpr.py` — plugin registration for all 9 rules

The base class uses the `selected_handles` optimizer pattern: `prepare()` scans all tags for names containing the substring, then collects matching object handles via `find_backlink_handles(tag_handle, include_classes=[self.namespace])`. The filter engine detects `selected_handles` and skips `apply_to_one` for non-matching objects, making the filter O(matching tags × matching objects) rather than O(all objects × all tags).

## Test plan

- [ ] Install addon and apply "People with a tag containing \<substring\>" filter in the People view
- [ ] Verify case-insensitive matching (e.g. "x" matches tags "indirect-x", "X-marker")
- [ ] Verify empty substring matches all tagged objects
- [ ] Verify no results when substring matches no tag names

### DIFF
--- a/HasTagSubstr/hastagsubstr.gpr.py
+++ b/HasTagSubstr/hastagsubstr.gpr.py
@@ -1,0 +1,171 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2025  Doug Blank <doug.blank@gmail.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


dsblank:
We can make this available for 6.0 too.

GaryGriffin:
Published for 6.0 and 6.1

PR #912, opened 2026-05-23
GenealogyTree: compact-template option for the Sandclock report (bug 10512)

eduralph:
## Root cause

The Sandclock Genealogy Tree report rendered only one page of an
arbitrarily-large tree. The LaTeX `genealogytree` package's default
`database` template lays the tree out so widely that big sandclocks
clip off the page; the package itself does not support multi-page
output for that layout. SNoiraud confirmed this on the Mantis ticket
in note ~0046612 and offered a workaround template
(`template=database pole reduced`) that gives ~4x more space per
page at the cost of denser per-node formatting.

## Fix

Expose SNoiraud's workaround as a user-facing report option. A new
"Compact tree layout" boolean on the Sandclock report appends
`template=database pole reduced` to the `\genealogytree[…]`
parameter list when enabled.

The directive is appended **last** so it overrides any node-spacing
defaults that `gramps/gen/plug/docgen/treedoc.py`'s built-in keys
(`level distance`, `node size`, `level size`) set earlier in the
parameter list — `pgfkeys` is order-sensitive, so a later
`template=` re-sets the keys the template controls.

Option-list assembly was extracted into a `_build_tree_options()`
method so the construction logic can be unit-tested without driving
a full report run (no LaTeX toolchain dependency).

## Verified against

- `GenealogyTree/gt_sandclock.py:148` — `_build_tree_options` body
- `GenealogyTree/gt_sandclock.py:96-128` — `write_report` calls into it
- `GenealogyTree/treeplugins.gpr.py` — **unchanged**; per the
  addons-source convention the maintainer manages the `.gpr.py`
  version, not the contributor

## Test

`GenealogyTree/tests/test_sandclock_compact_template.py` — five focused
unit tests exercising the new option-list builder via
`SandclockTree.__new__(...)` skeleton (no Gramps GUI, no LaTeX, no
database):

- Default layout omits any `template=` directive
- `compact=True` appends exactly `template=database pole reduced`
- Directive lands **last** in the option list (pgfkeys order)
- Other default options (`pref code=`, `list separators hang`,
  `place text=`, `box=`) remain intact when compact is on
- `include_images` and `compact` coexist without either suppressing
  the other

All 5 pass via `gramps-testbed/scripts/ubuntu/run-addon-unit.sh GenealogyTree`.

Fixes #10512

### DIFF
--- a/GenealogyTree/gt_sandclock.py
+++ b/GenealogyTree/gt_sandclock.py
@@ -90,6 +90,7 @@ class SandclockTree(Report):
         self.max_down = menu.get_option_by_name('gendown').get_value()
         self.include_siblings = menu.get_option_by_name('siblings').get_value()
         self.include_images = menu.get_option_by_name('images').get_value()
+        self.compact = menu.get_option_by_name('compact').get_value()
         self.set_locale(menu.get_option_by_name('trans').get_value())
 
     def write_report(self):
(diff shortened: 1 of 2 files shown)


GaryGriffin:
Should this be on gramps60 instead? And then also cherry picked to gramps61?

eduralph:
Rebased as asked, ready for cherry pick

PR #915, opened 2026-05-24
Fix make.py listing wiping addons-<lang>.json for excluded addon (bug 13694)

eduralph:
## Root cause
`make.py <ver> listing <Addon>` builds a per-language `listings`
buffer and then either merges it with the existing
`addons-<lang>.json` or writes a fresh file. When the targeted addon
yields no eligible plugin — `include_in_listing=False`, or the
`.addon.tgz` isn't built yet — `listings` is empty, the merge loop
doesn't iterate, `output` stays `[]`, and the next statement
overwrites the file with `[]`. Every previously listed addon for
that language is wiped.

## Fix
In the per-language loop, after `listings` is built, gate the write:
when `cmd_arg != "all"` and `listings` is empty and the listings
file already exists, skip the write and print a clear pointer to
`unlist` for the user. The existing entries are preserved.

The same path string —
`f"../addons/{gramps_version}/listings/" + ("addons-%s.json" % lang)`
— was duplicated at three sites: the existence check in the merge
branch, the `open(..., "r")` that reads the existing file inside the
merge loop, and the final `open(..., "w")` that writes the output.
Adding a fourth use for the new guard would have made it four
copies in a 50-line block, and the bug is precisely the kind that
hides when one of those copies drifts (the guard checking one path
while the writer overwrites another). Hoisting into a single
`listings_path` local at the top of the loop makes the invariant
"all references read and write the same file" syntactically
enforced and keeps the new guard, the existence check, the read,
and the write trivially aligned.

## Verified against
- `make.py:947-977` — the `for p in plugins` block that filters by
  `include_in_listing` and prints `"   ignoring '%s'"` without
  contributing to `listings`. This is where `listings` ends up
  empty for an excluded addon.
- `make.py:978-1033` (pre-fix) — the write path that consumed an
  empty `output` and produced `[]` on disk. The `else:` branch
  (`# just update the lines from these addons:`) is where the
  `for plugin in sorted(listings, …)` loop fails to iterate.
- prculley's note 1 on the tracker — `listing` is not the unlist
  command, so silently corrupting the file is doubly wrong; the
  user message in the guard tells them which command to use.

## Test
`tests/test_make_listing.py` builds a temp addons-source/addons
pair, creates a synthetic addon with `include_in_listing=False`, a
placeholder `.addon.tgz`, and a seeded `addons-en.json` containing
one entry; then runs `python3 make.py <ver> listing ExcludedAddon`
as a subprocess and asserts the seed survives. Pre-fix the
assertion fails (`[] != [seed]`); post-fix it passes.

```
python3 -m unittest tests.test_make_listing -v
```

Resolves #13694

### DIFF
--- a/make.py
+++ b/make.py
@@ -973,13 +973,26 @@ elif command == "listing":
                         print("   ignoring '%s'" % (p["name"]))
         # Write out new listing:
         output = []
+        listings_path = (
+            f"../addons/{gramps_version}/listings/" + ("addons-%s.json" % lang)
+        )
+        if cmd_arg != "all" and not listings and os.path.isfile(listings_path):
+            # Bug 13694: a single-addon `listing <Addon>` whose source yields
+            # no eligible plugin (include_in_listing=False, or no .tgz built
+            # yet) used to overwrite the listings file with [], wiping every
+            # previously listed addon. Leave the file untouched and tell the
+            # user how to remove an existing entry on purpose.
+            print(
+                "   '%s' produced no listing entry; leaving %s untouched. "
+                "Use 'make.py %s unlist %s' to remove an existing entry."
+                % (cmd_arg, os.path.basename(listings_path), gramps_version, cmd_arg)
+            )
+            continue
         if cmd_arg == "all":
             # Replace it!
             for plugin in sorted(listings, key=lambda p: (p["t"], p["i"])):
                 output.append(plugin)
-        elif not os.path.isfile(
-            f"../addons/{gramps_version}/listings/" + ("addons-%s.json" % lang)
-        ):
+        elif not os.path.isfile(listings_path):
             for plugin in sorted(listings, key=lambda p: (p["t"], p["i"])):
                 output.append(plugin)
         else:
@@ -987,7 +1000,7 @@ elif command == "listing":
             for plugin in sorted(listings, key=lambda p: (p["t"], p["i"])):
                 already_added = []
                 fp_in = open(
-                    f"../addons/{gramps_version}/listings/" + ("addons-%s.json" % lang),
+                    listings_path,
                     "r",
                     encoding="utf-8",
                 )
@@ -1021,7 +1034,7 @@ elif command == "listing":
                         output.append(plugin)
         mkdir(f"../addons/{gramps_version}/listings")
         fp_out = open(
-            f"../addons/{gramps_version}/listings/" + ("addons-%s.json" % lang),
+            listings_path,
             "w",
             encoding="utf-8",
             newline="",
--- a/make.py
+++ b/make.py
@@ -996,42 +996,43 @@ elif command == "listing":
             for plugin in sorted(listings, key=lambda p: (p["t"], p["i"])):
                 output.append(plugin)
         else:
-            # just update the lines from these addons:
-            for plugin in sorted(listings, key=lambda p: (p["t"], p["i"])):
-                already_added = []
-                fp_in = open(
-                    listings_path,
-                    "r",
-                    encoding="utf-8",
-                )
-                added = False
-                for line in json.load(fp_in):
-                    if line["i"] in already_added:
-                        continue
-                    if (
-                        cmd_arg + ".addon.tgz" == line["z"]
-                        and plugin["t"] == line["t"]
-                        and not added
-                    ):
-                        # print("UPDATED")
-                        output.append(plugin)
-                        added = True
-                        already_added.append(line["i"])
-                    elif (
-                        (plugin["t"], plugin["i"]) < (line["t"], line["i"])
-                    ) and not added:
-                        # print("ADDED in middle")
-                        output.append(plugin)
-                        added = True
-                        output.append(line)
-                        already_added.append(line["i"])
-                    else:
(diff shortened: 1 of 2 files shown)


GaryGriffin:
There is another case where this fails the same way. If there are multiple gpr files in a folder, the listing file gets corrupt. For example, Form. 

Checked and this still fails in this case. Can you address that case too?
Do you want this on gramps60 and then cherry picked to gramps61?

eduralph:
I'll look at the issue you mentioned and sort it out.

Concerning the gramps60 branch, I'm a bit confused tbh. gramps core has switched over to gramps61 so I've been running my fixes against that despite there being a 6.0.9 in mantis. For addons-source we seem to be doing something slightly different. Is there a reference page or something that I can refer to? Or maybe I don't quite get the strategy here? That would be a great help.

Anyways, the last 3 PRs you flagged as gramps60, I'm happy to do it that way and rebase it so you can cherry pick it

eduralph:
Quick update on the issue you raised — reproduced it cleanly. Pushed a follow-up commit (`abbeb0433`) that extends the fix. The root cause is in the same merge path but a separate symptom: the outer `for plugin in sorted(listings, ...)` loop reopened and re-read the existing listings file on each iteration while accumulating into a shared `output`. With three new plugins for Form (two .gpr.py files, three `register()` calls total), every existing entry ended up appended three times — 170 → 510 entries in `addons-en.json`, every other addon appearing thrice. Same shape
would corrupt for any addon with N>1 registers.

The follow-up commit replaces the per-plugin re-read with a single-pass merge: read the existing file once, drop every row that belongs to `cmd_arg` (matched by `.z`) or whose `(t, i)` collides with one of the fresh plugins, then merge the kept rows with the new plugins in sorted `(t, i)` order. Two side-effects worth flagging, both intentional:

1. Stale entries for `cmd_arg` (e.g. a `register()` removed from a  `.gpr.py` since the last build) are now dropped, not preserved.  The old code couldn't distinguish "this entry's plugin still  exists" from "this entry is stale" because it matched lines by  `(z, t)` — multiple Form lines all matched `Form.addon.tgz` so  "first match wins" was effectively undefined.
2. The "first match wins" ambiguity of the old `z + t` matcher is  gone; replacement matches per-plugin by `(t, i)`. For Form the  stale `formgramplet` entry gets replaced with the fresh  `myform_gramplet` (in the test fixture) without touching the two  quickreports.

Test extended in `tests/test_make_listing.py` — the multi-gpr case fails on the pre-fix code with the exact "14 != 6" duplication symptom, passes after. Verified end-to-end against the real `../addons/gramps61/listings/addons-en.json`: `listing Form` keeps the file at 170 entries with 3 Form rows; `listing CheckPlaceTitles` (`include_in_listing=False`) and `listing QuiltView` (single-plugin) still behave correctly.

GaryGriffin:
There will NOT be another core release of 6.0.x gramps. The next release for core will be 6.1 . 

For Addons, I can be convinced to do 5.2 PRs if really justified. The 6.1 addons branch needed to be available before the 6.1 core first-available - in this case beta1. But most users are still on 6.0, given 6.1 is beta. 

So most addon PRs should still be on 6.0 and I cherry pick to 6.1 . There are some addons that are only on 6.1 (PDFForms, for instance). That is the authors decision.

Once 6.1 is released, I will stop cherry picking PRs from 6.0 and will ask for a second PR for the other branch. Maybe this is wrong, but is how I am currently thinking. For example. Websearch has 2 PRs open currently - one for 6.0 and one for 5.2 . I will cherry pick 6.0 --> 6.1 when merged. 

Does that help?

eduralph:
> There will NOT be another core release of 6.0.x gramps. The next release for core will be 6.1 .
> 
> For Addons, I can be convinced to do 5.2 PRs if really justified. The 6.1 addons branch needed to be available before the 6.1 core first-available - in this case beta1. But most users are still on 6.0, given 6.1 is beta.
> 
> So most addon PRs should still be on 6.0 and I cherry pick to 6.1 . There are some addons that are only on 6.1 (PDFForms, for instance). That is the authors decision.
> 
> Once 6.1 is released, I will stop cherry picking PRs from 6.0 and will ask for a second PR for the other branch. Maybe this is wrong, but is how I am currently thinking. For example. Websearch has 2 PRs open currently - one for 6.0 and one for 5.2 . I will cherry pick 6.0 --> 6.1 when merged.
> 
> Does that help?

Thanks, that helps - I'm not 100% familiar with the release cycle so I guess I'm still swimming a bit. Hopefully my GEPS lands and those worries are gone, lol.

eduralph:

dsblank:
> So most addon PRs should still be on 6.0 and I cherry pick to 6.1 . There are some addons that are only on 6.1 (PDFForms, for instance). That is the authors decision.

I didn't know that. What should an author do if they want it for 6.0 and 6.1 after it has been merged to 6.1?

GaryGriffin:
Just ask and I will cherry pick backwards.

GaryGriffin:
@dsblank Do you approve this change?

dsblank (approved):
Looks good to me!

PR #921, opened 2026-05-25
ImportGramplet: pin death-fallback contract for clean-XML import (bug 13420 — cannot-reproduce evidence)

eduralph:
## Root cause
Mantis 13420 reports that a person imported via the Text Import gramplet
does NOT show their Death event as the death fallback in list views until
the event ref is re-validated with OK in the editor. snoiraud established
on the tracker (note 8) that the reporter's XML was generated by a no-code
Android tool with UUID handles and longer-than-normal change-times; no
developer ever reproduced the symptom on conformant Gramps XML.

The fallback-setting logic lives in gramps **CORE** at
`gramps/plugins/importer/importxml.py:1434-1473` (`start_eventref`) and
is inherited unchanged by the addon's `AtomicGrampsParser`. The core
parser itself spells out its precondition at line 1451-1452: *"We count
here on events being already parsed prior to parsing people or families.
This code will fail if this is not true."* If `inaugurate` fails to
resolve the event handle (e.g. UUID handles that don't match between
the events block and the eventref `hlink` attributes, or events listed
after people), `start_eventref` silently early-returns and never wires
the fallback.

## Fix
None. This PR ships only a regression test that pins the contract on
clean input.

## Verified against
- `gramps/plugins/importer/importxml.py:1434-1473` — `start_eventref`
  is the only place death_ref is set during XML import; the early
  return at line 1454-1455 fires when the event handle doesn't resolve.
- `ImportGramplet/ImportGramplet.py:102-154` — `AtomicGrampsParser`
  subclasses `GrampsParser` and overrides only `parse()` for the atomic
  DbTxn wrapper; per-element handlers including `start_eventref` are
  inherited verbatim.

## Test
`ImportGramplet/tests/test_xml_death_fallback.py` (new; stdlib
`unittest`, plus `tests/__init__.py`). One case:

- Build CONFORMANT minimal Gramps XML — events listed before people,
  standard Gramps handles (no UUIDs), both events `role="Primary"`.
- Parse via `AtomicGrampsParser` against an in-memory sqlite Gramps
  database (same pattern as `Sqlite/tests/test_sqlite.py`).
- Fetch the single imported person **by handle** — core normalises
  gramps_ids to the database's configured width on import (the XML
  `I00001` becomes `I0001` under the default `I%04d`), so a literal
  gramps_id lookup is unreliable.
- Assert `person.get_death_ref()` is set; control-assert
  `person.get_birth_ref()` too; both refs resolve to events of the
  correct type.

Gated on `gi.require_version("Gtk", "3.0")` **and**
`gi.require_version("Gdk", "3.0")` because the addon's import chain pulls
both at module load (`editform`-style, Gdk before Gtk); pinning both —
mirroring `gramps/gen/constfunc.py` — keeps the imports on the GTK 3 stack
regardless of the host's default GI resolution. Headless; no display
required.

**Result: 1 test, OK** — verified two ways on the PR's base branch:
- direct `python3 -m unittest ImportGramplet.tests.test_xml_death_fallback`
  against `maintenance/gramps60`;
- gramps-testbed addon-unit runner (`run-addon-unit.sh ImportGramplet`).

The test passing IS the empirical evidence that Mantis 13420 does NOT
reproduce on conformant XML — the reporter's malformed UUID-handle XML
was the cause. The test stays as a regression guard so a later refactor
of the atomic parser cannot silently break the fallback wiring.

> **Correction (revised commit).** The first revision looked the person up
> by `get_person_from_gramps_id("I00001")` and reported a green result that
> did not hold — gramps_id normalisation makes that lookup return `None`, so
> the test actually failed on `maintenance/gramps60` (and on macOS, as
> GaryGriffin reported). The handle-based lookup above is the corrected,
> genuinely-green form.

Companion to a `cannot-reproduce / invalid input` close on bug 13420.

Bug #13420

### DIFF
--- a/ImportGramplet/tests/test_xml_death_fallback.py
+++ b/ImportGramplet/tests/test_xml_death_fallback.py
@@ -1,0 +1,187 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2026  Gramps Development Team
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+"""
+Repro-or-close test for bug 13420: Text Import gramplet death event
+is not shown as the person's death fallback until re-validated in the
+editor.
+
+The reporter's input was XML produced by a no-code Android tool with
+UUID handles and longer-than-normal change-times; no developer
+reproduced the symptom on conformant Gramps XML. Per the triage
+verdict in `triage/batches/batch-03-confirmed-velocity/issue_13420.md`
+the decisive test is:
+
+  - Build CONFORMANT minimal Gramps XML (one person, Birth + Death
+    events both role Primary, standard Gramps handles)
+  - Import via the addon's `AtomicGrampsParser` (the wrapper the
+    ImportGramplet uses for its XML branch)
+  - Assert person.get_death_ref() returns the Death event ref
+    WITHOUT manual editor re-save
+
+If clean XML imports correctly → reporter's XML was malformed; close
+13420 as cannot-reproduce. If clean XML still fails → real addon bug,
+fix needed.
+
+The fallback-setting logic lives in gramps CORE at
+`gramps/plugins/importer/importxml.py:1434-1473` (`start_eventref`)
+and is shared between the standard XML import and AtomicGrampsParser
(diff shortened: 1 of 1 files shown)


GaryGriffin:
When I run from your branch, I get a failure.

COMMAND: `python3 -m unittest ImportGramplet.tests.test_xml_death_fallback`

Results:
```

(python3 -m unittest:39577): Gdk-WARNING **: 11:11:00.874: Failed to initialize CVDisplayLink!
UpdateCallback with total == 0 created
3400%F
FAIL: test_clean_xml_imports_set_death_ref_via_atomic_parser (ImportGramplet.tests.test_xml_death_fallback.TestImportGrampletDeathFallback.test_clean_xml_imports_set_death_ref_via_atomic_parser)
Clean XML through AtomicGrampsParser must set the
  File "/Volumes/Storage/Development/addons-source/ImportGramplet/tests/test_xml_death_fallback.py", line 158, in test_clean_xml_imports_set_death_ref_via_atomic_parser
    self.assertIsNotNone(
    ~~~~~~~~~~~~~~~~~~~~^
        person,
        ^^^^^^^
        "Person I00001 must exist after XML import",
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
AssertionError: unexpectedly None : Person I00001 must exist after XML import

Ran 1 test in 0.382s

FAILED (failures=1)

```

PR #926, opened 2026-05-26
Form editor: guard save against mid-edit deletion of referenced objects (bug 11054)

eduralph:
## Root cause
The Form editor has no protection against changes made outside it
while it is open. If a Source / Person / Family the form references is
deleted in another window mid-edit, **OK** writes a citation with a
None-target reference; Check & Repair later flags the dangling handle
as db corruption. nick_h documents this on notes 5/7 of the tracker
issue as a known limitation of the snapshot/write-back design and
calls out an acceptable narrow fix that avoids the wider architectural
change.

## Fix
Implement nick_h's narrow save-time guards verbatim — no behaviour
change when the references are intact:

* `EditForm.save` (source path): if the citation's source handle no
  longer resolves, create a replacement `Source` and repoint the
  citation at it rather than commit a None-target reference.
* `MultiSection.save` (person rows + removed-people loop): skip rows
  whose person handle no longer resolves, both in the row-update loop
  and the detach loop.
* `PersonSection.save`: early-return when the attached person handle
  no longer resolves; skip the detach step when the initial handle is
  gone.
* `FamilySection.save`: same shape as `PersonSection.save`, on family.

Architecture-level fixes (snapshot / per-form locking / JSON-patch
write-back — all discussed on the tracker) are **explicitly out of
scope** per nick_h.

## Verified against
- `Form/editform.py:EditForm.save` — insertion point between
  `add_event` and `commit_citation`; the source handle is read here.
- `Form/editform.py:MultiSection.save` — row-update loop (handle
  comes from `row[0]` in the model) and removed-people detach loop
  (handles come from `set(initial_people) - set(all_people)`).
- `Form/editform.py:PersonSection.save` — `self.handle` is the
  current person, `self.initial_handle` the previously-attached one.
- `Form/editform.py:FamilySection.save` — same shape on family.

## Test
`Form/tests/test_editform_save_guards.py` — eight cases across four
`TestCase` classes:

* `EditFormSourceGuardTest` — drives `EditForm.save` via
  `__new__`-bypass with `DbTxn` patched at the module level;
  asserts `add_source` + `set_reference_handle` fire when the source
  is gone, and don't fire on the happy path.
* `MultiSectionPersonGuardTest` — drives `MultiSection.save` against
  a mock model + a mock db whose `get_person_from_handle` returns
  `None` for dead handles. Asserts the dead row is skipped (commit
  count = live row count) and the dead-removed-person case commits
  nothing.
* `PersonSectionGuardTest` — drives `PersonSection.save`; asserts
  dead `self.handle` short-circuits with no commits, and a live
  attached + dead initial handle commits the attached person and
  skips the detach.
* `FamilySectionGuardTest` — same shape as PersonSection, on family.

Pre-fix: 2 failures + 5 errors among the eight cases (i.e. each guard
is load-bearing). Post-fix: green.

Verified locally via the gramps-testbed addon-unit runner
(`scripts/ubuntu/run-addon-unit.sh Form`): 56 tests OK with the patch
applied.

Fixes #11054

### DIFF
--- a/Form/editform.py
+++ b/Form/editform.py
@@ -63,6 +63,7 @@ from gramps.gen.lib import (
     Person,
     Family,
     Attribute,
+    Source,
 )
 from gramps.gen.db import DbTxn
 from gramps.gen.display.name import displayer as name_displayer
@@ -379,6 +380,20 @@ class EditForm(ManagedWindow):
         with DbTxn(self.get_menu_title(), self.db) as trans:
             if not self.event.get_handle():
                 self.db.add_event(self.event, trans)
+
+            # Mantis 11054 (nick_h's narrow guard): if the source the
+            # citation references was deleted while this form was open,
+            # write a replacement source rather than committing a
+            # dangling-reference citation. No behaviour change when the
+            # reference is intact.
+            source_handle = self.citation.get_reference_handle()
+            if source_handle and self.db.get_source_from_handle(source_handle) is None:
+                replacement = Source()
+                replacement.set_title(
+                    _("[Source recreated after deletion mid-form-edit]")
+                )
+                self.db.add_source(replacement, trans)
+                self.citation.set_reference_handle(replacement.get_handle())
 
             citation_handle = self.citation.get_handle()
             if not self.citation.get_handle():
@@ -891,6 +906,12 @@ class MultiSection(Gtk.Box):
         for order, row in enumerate(self.model):
             all_people.append(row[0])
             person = self.db.get_person_from_handle(row[0])
+            # Mantis 11054 (nick_h's narrow guard): if the person was
+            # deleted while this form was open, drop the row rather than
+            # commit a None-target reference. No behaviour change when
+            # the reference is intact.
+            if person is None:
+                continue
             event_ref = get_event_ref(self.event, person, self.role)
 
             # Write attributes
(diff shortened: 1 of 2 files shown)


GaryGriffin:
Running unittest generates Warning:

> Form/editform.py:30: PyGIWarning: Gdk was imported without specifying a version first. Use gi.require_version('Gdk', '3.0') before import to ensure that the right version gets loaded.
>   from gi.repository import Gdk

eduralph:
Thanks for allin git out. It's aproblem with my Gdk-4 default box which swallows tests silently. I need to update my testbed and sort this. I'll provide an update soon

eduralph:
So this opened a can of worms: the immediate issue is that the addon relies on  gramps to enforce Gtk 3 lamd Gdk 3 instead of 4 oading. In cases of runtime & boxes that default to Gtk 3, the tests run fine. In cases of Gdk 4 default boxes, it loads the wrong Gdk version and throws an error and the tests catching of ImportErrors ends skipping the test. 

What does that mean:  

1. All unit tests in addons-source must pin the Gtk & Gdk version to 3 to work correctly on all machines
2. I had to  rework PR 820 to pin the version of Gtk & Gdk just like gramps does to proivde the correct environment. 
3. I think I'll rewrite the section about Addon development in the wiki, if that is permissable? These kind of insights need documentation.

I have 2,3 more items that needed updating arising from other PR issues you raised. You can find them in the commits and comments in PR 820.

Nick-Hall:
> I think I'll rewrite the section about Addon development in the wiki, if that is permissable?

Yes.  Feel free to improve the documentation.

PR #931, opened 2026-05-29
Add __init__.py to all unittest folders

GaryGriffin:
If a unittest folder is missing the __init__.py, create it.


GaryGriffin:
@eduralph 
Is this PR needed for unittest discover? There are 5 tests directories that do not have the __init__.py file.

eduralph:
@GaryGriffin  - great addition to improving the overall unit testing capabilities. I'll include this aspect in the Addon Development Documentation. I already have a change in mind to move the Gtk 3 and Gdk 3 dependency pinning into a central __init__.py

PS - I verified it against the CI in PR 820 and it runs clean

GaryGriffin:
Do we need also on gramps60? Would need a separate PR since addon list would be different.

PR #933, opened 2026-05-31
New "exclude subtree" filter

azrdev:
This adds a new Person Filter rule, which "walks the tree" of relatives (looking for each person at all their Families and the parents and children in them for new persons), starting at a given person, but *stopping* if any person from a specified filter is encountered.
This allows for splitting one branch from a bigger tree with only two filters, e.g. when a Family Tree contains two married persons and their families, which are not otherwise connected.

I've published my code already years ago <https://gramps.discourse.group/t/how-to-extract-a-subtree-of-my-tree/5142/14> but now I'd finally see it submitted to the addons repo.

### DIFF
--- a/excludesubtree.gpr.py
+++ b/excludesubtree.gpr.py
@@ -11,7 +11,7 @@ register(RULE,
   name  = _("All people, starting at <person>, except those in/behind <filter>"),
   description = _("Matches people who are reachable starting from <person> "
       "except those in <filter>."),
-  version = '0.3',
+  version = '0.4',
   authors = ["Jonathan Biegert"],
   authors_email = ["azrdev@gmail.com"],
   gramps_target_version = '5.1',
--- a/excludesubtree.gpr.py
+++ b/excludesubtree.gpr.py
@@ -1,0 +1,22 @@
+
+"""
+Person filter rule including all persons reachable from the active/selected person, except those in the given filter.
+
+This allows to cut partial trees, e.g. exclude everything learned about my spouse but have all non-relatives in my half of the tree.
+"""
+
+# https://github.com/gramps-project/gramps/blob/master/gramps/gen/plug/_pluginreg.py
+register(RULE,
+  id    = 'ExcludeSubtree',
+  name  = _("All people, starting at <person>, except those in/behind <filter>"),
+  description = _("Matches people who are reachable starting from <person> "
+      "except those in <filter>."),
+  version = '0.2',
+  authors = ["Jonathan Biegert"],
+  authors_email = ["azrdev@gmail.com"],
+  gramps_target_version = '5.1',
+  status = UNSTABLE,
+  fname = "excludesubtree.py",
+  ruleclass = 'ExcludeSubtree',  # must be rule class name
+  namespace = 'Person',  # one of the primary object classes
+  )
--- a/excludesubtree.gpr.py
+++ b/excludesubtree.gpr.py
@@ -11,7 +11,7 @@ register(RULE,
   name  = _("All people, starting at <person>, except those in/behind <filter>"),
   description = _("Matches people who are reachable starting from <person> "
       "except those in <filter>."),
-  version = '0.2',
+  version = '0.3',
   authors = ["Jonathan Biegert"],
   authors_email = ["azrdev@gmail.com"],
   gramps_target_version = '5.1',
--- a/excludesubtree.gpr.py
+++ b/excludesubtree.gpr.py
@@ -27,12 +27,12 @@ register(
         "(walking all parents and children of attached families, "
         "recursively) stopping at persons in <Filter>."
     ),
-    version="0.5",
+    version="0.6",
     authors=["Jonathan Biegert"],
     authors_email=["azrdev@gmail.com"],
     gramps_target_version="6.0",
     status=BETA,
     fname="excludesubtree.py",
-    ruleclass="ExcludeSubtree",  # must be rule class name
-    namespace="Person",  # one of the primary object classes
+    ruleclass="ExcludeSubtree",
+    namespace="Person",
 )
--- a/excludesubtree.gpr.py
+++ b/excludesubtree.gpr.py
@@ -1,16 +1,29 @@
-
-"""
-Person filter rule including all persons reachable from the active/selected person, except those in the given filter.
-
-This allows to cut partial trees, e.g. exclude everything learned about my spouse but have all non-relatives in my half of the tree.
-"""
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2026  Jonathan Biegert
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <https://www.gnu.org/licenses/>.
+#
 
 # https://github.com/gramps-project/gramps/blob/master/gramps/gen/plug/_pluginreg.py
 register(RULE,
   id    = 'ExcludeSubtree',
-  name  = _("All people, starting at <person>, except those in/behind <filter>"),
-  description = _("Matches people who are reachable starting from <person> "
-      "except those in <filter>."),
+  name  = _("People reachable from <Person>, stopping at <Filter> matches"),
+  description = _("Matches people who are reachable starting from <Person> "
+                  "(walking all parents and children of attached families, "
+                  "recursively) stopping at persons in <Filter>."),
   version = '0.5',
   authors = ["Jonathan Biegert"],
   authors_email = ["azrdev@gmail.com"],
--- a/excludesubtree.gpr.py
+++ b/excludesubtree.gpr.py
@@ -18,18 +18,21 @@
 #
 
 # https://github.com/gramps-project/gramps/blob/master/gramps/gen/plug/_pluginreg.py
-register(RULE,
-  id    = 'ExcludeSubtree',
-  name  = _("People reachable from <Person>, stopping at <Filter> matches"),
-  description = _("Matches people who are reachable starting from <Person> "
-                  "(walking all parents and children of attached families, "
-                  "recursively) stopping at persons in <Filter>."),
-  version = '0.5',
-  authors = ["Jonathan Biegert"],
-  authors_email = ["azrdev@gmail.com"],
-  gramps_target_version = '6.0',
-  status = BETA,
-  fname = "excludesubtree.py",
-  ruleclass = 'ExcludeSubtree',  # must be rule class name
-  namespace = 'Person',  # one of the primary object classes
+register(
+    RULE,
+    id="ExcludeSubtree",
+    name=_("People reachable from <Person>, stopping at <Filter> matches"),
+    description=_(
+        "Matches people who are reachable starting from <Person> "
+        "(walking all parents and children of attached families, "
+        "recursively) stopping at persons in <Filter>."
+    ),
+    version="0.5",
+    authors=["Jonathan Biegert"],
+    authors_email=["azrdev@gmail.com"],
+    gramps_target_version="6.0",
+    status=BETA,
+    fname="excludesubtree.py",
+    ruleclass="ExcludeSubtree",  # must be rule class name
+    namespace="Person",  # one of the primary object classes
 )
--- a/excludesubtree.gpr.py
+++ b/excludesubtree.gpr.py
@@ -11,12 +11,12 @@ register(RULE,
   name  = _("All people, starting at <person>, except those in/behind <filter>"),
   description = _("Matches people who are reachable starting from <person> "
       "except those in <filter>."),
-  version = '0.4',
+  version = '0.5',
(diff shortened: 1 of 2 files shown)


dsblank:
@azrdev if you simply rename `matched_relatives` to `selected_handles` it will be optimizable in a variety of ways.

I'll write the details up in a document.

dsblank:
I attempted to describe `selected_handles` and optimizations here: https://github.com/gramps-project/gramps/discussions/2350

azrdev:
@dsblank thanks! After reading it I gather the concept of your optimizer, though not every detail, but I think I don't need to after 0cfcc88 as you suggested.

That said, anything else needed for this PR to be merged & my rule plugin to be published?

dsblank (changes requested):
## Review: New "ExcludeSubtree" filter

The algorithm and concept are correct and the feature is genuinely useful — this fills a real gap not covered by existing filter rules. However there are several bugs that need to be fixed before merging.

---

### Critical bugs

**1. GUI import breaks CLI usage**

```python
from gramps.gui.editors.filtereditor import MyBoolean
```

No existing filter rule in `gramps/gen/` ever imports from `gramps.gui`. This import makes the entire module fail to load in any non-GUI context (CLI, scripts, batch processing). The `GUICheckBox` wrapper only exists to fix `MyBoolean`'s constructor signature — consider whether a boolean parameter is even necessary (accept `"0"`/`"1"` strings with a descriptive label, as other rules do), or find a way to reference the widget class without a top-level GUI import.

**2. Missing type imports — `NameError` at import time**

`Set`, `PrimaryObjectHandle`, `Database`, and `Person` are used in annotations but never imported. The class attribute annotation is evaluated at class-definition time, so this raises `NameError` on import. Compare with `_ischildoffiltermatch.py`:

```python
from typing import Set
from gramps.gen.lib import Person
from gramps.gen.db import Database
```

**3. Mutable class-level attribute**

```python
class ExcludeSubtree(Rule):
    selected_handles: Set[PrimaryObjectHandle] = set([])  # shared across all instances
```

This is the classic Python mutable-default bug — all instances share one set until `reset()` is called. Initialize it as an instance attribute inside `prepare()` instead, matching the pattern in `_ischildoffiltermatch.py`:

```python
def prepare(self, db, user):
    self.selected_handles: set[str] = set()
    ...
```

**4. `reset()` never tears down the sub-filter**

`_ischildoffiltermatch.py` calls `self.filt.requestreset()` in `reset()`. Without it the sub-filter stays prepared (holds a DB reference, progress state, etc.):

```python
def reset(self):
    self.filt.requestreset()   # ← missing
    self.selected_handles.clear()
```

**5. Potential `AttributeError` when Gramps ID is not found**

```python
start = db.get_person_from_gramps_id(self.list[0]).handle
```

`get_person_from_gramps_id` returns `None` for an unknown ID; chaining `.handle` raises `AttributeError`. Guard against it:

```python
person = db.get_person_from_gramps_id(self.list[0])
if person is None:
    return
start = person.handle
```

---

### Significant issues

**6. Progress counter can exceed its declared total**

`user.begin_progress(..., db.get_number_of_people())` declares the total as the number of people in the DB, but `user.step_progress()` is called once per BFS queue pop — which can far exceed that count before the duplicate-handle guard fires. Use `0` as the total for indeterminate progress, or restructure.

**7. Name/description mismatch between `gpr.py` and the class**

- `gpr.py`: `"All people, starting at <person>, except those in/behind <filter>"`
- class `name`: `"People reachable from <Person>, stopping at <Filter> matches"`

These should be the same string — the `gpr.py` value is what appears in the rule selector UI.

---

### Style / conventions

**8.** Both `.py` files are missing the required GPL-2.0-or-later license header block.

**9.** Import sections need the standard comment headers (`# Standard Python modules`, `# Gramps modules`, etc.) used throughout the codebase.

**10.** Each class needs a navigation header comment:
```python
#------------------------------------------------------------
#
# ExcludeSubtree
#
#------------------------------------------------------------
class ExcludeSubtree(Rule):
```

**11.** The `ExcludeSubtree` class has no docstring.

**12.** `set([])` → `set()`.

**13.** Run `black .` before committing — the gpr.py alignment spaces and the `labels` tuples will likely fail CI's `black --check`.

---

### Minor

- `get_relatives` yields `None` handles (missing father/mother); filtering inside the generator rather than at the call site would be cleaner.
- `bool(int(self.list[1]))` works but a short comment explaining the double-cast would help readers.

---

`_ischildoffiltermatch.py` in the main repo is a good model for the overall structure. Happy to help iterate on a revised version.

azrdev:
Thanks a lot for your review, I'll address your remarks (though it'll take some days until I get to it)

azrdev:
> **1. GUI import breaks CLI usage**

> No existing filter rule in `gramps/gen/` ever imports from `gramps.gui`. This import makes the entire module fail to load in any non-GUI context (CLI, scripts, batch processing). 

A few plugins in the addon repository do (e.g. <FilterRules/hasrolerule.py>). And I actually used the code as-is in gramps headless operations (report generation) without crash. You're correct, though, that this breaks encapsulation, I've changed the code to take a stringly-typed enum (filtereditor.py really could use some work :O )

Everything else I've addressed, so I'd be glad for another review!

GaryGriffin:
I ran on Mac Gramps 6.0.6 and it failed on install:

> /Applications/Gramps6.0.6.app/Contents/MacOS/Gramps
>   File "/Applications/Gramps6.0.6.app/Contents/Resources/lib/python3.13/site-packages/gramps/gen/plug/_manager.py", line 270, in load_plugin
>     _module = self.import_plugin(pdata)
>   File "/Applications/Gramps6.0.6.app/Contents/Resources/lib/python3.13/site-packages/gramps/gen/plug/_manager.py", line 304, in import_plugin
>     module = __import__(pdata.mod_name)
>   File "/Users/gary/Library/Application Support/gramps/gramps60/plugins/ExcludeSubtreeFilter/excludesubtree.py", line 85, in <module>
>     class ExcludeSubtree(Rule):
>     ...<77 lines>...
>             return person.handle in self.selected_handles
>   File "/Users/gary/Library/Application Support/gramps/gramps60/plugins/ExcludeSubtreeFilter/excludesubtree.py", line 93, in ExcludeSubtree
>     selected_handles: Set[PrimaryObjectHandle] = None
>                           ^^^^^^^^^^^^^^^^^^^
> NameError: name 'PrimaryObjectHandle' is not defined

azrdev:
@GaryGriffin  thank you, I fixed the type annotations

dsblank (changes requested):
## Review — "New Exclude Subtree Filter"

The core BFS logic is correct and fills a real gap. A few issues need addressing before merge.

---

### Bug: Mutable class-level attributes

```python
# excludesubtree.py:86-87
selected_handles: Set[PersonHandle] = set()
filt = None
```

`selected_handles` as a class attribute means **all instances share the same set** until `reset()` reassigns it as an instance attribute. While `reset()` is called at the top of `prepare()`, any code path that calls `apply_to_one` before `prepare()` (or after `requestreset()`) reads the shared class-level set. The correct pattern (as seen in `isrelatedwithfiltermatch.py`) is to initialize these only in `prepare()`:

```python
def prepare(self, db, user):
    self.selected_handles: Set[PersonHandle] = set()
    self.filt = None
    ...
```

Remove the class-level declarations entirely.

---

### Bad UX: Second parameter is a free-text "include/exclude" toggle

```python
_("Handle filter matches: include (default), exclude"),
```

This renders as a plain text input where the user must literally type `"exclude"`. Any other value silently defaults to include. Consider dropping this parameter and creating two separate rules (one that includes the boundary, one that excludes), which is simpler and avoids the invisible string-matching.

---

### Missing `po/` directory

All other filter addons (`FilterRules/`, `FilterRules2/`) include a `po/` directory. Without it the translatable strings (`_("...")`) will never be extracted for localization.

---

### Moderate Issues

- **Missing `help_url`** in `excludesubtree.gpr.py`. Every other FilterRules addon links to the Gramps wiki via `help_url="Addon:Rule_expansions#..."`.
- **Missing `from __future__ import annotations`** — other Python files in this repo include it.
- **Version `"0.6"`** — conventional starting versions in this repo are `"1.0.0"`. `"0.6"` implies five prior unreleased versions.
- **Verbose debug logging inside the BFS loop** — `LOG.debug("tree walk arrived at id %s", ...)` fires once per person walked. On a large tree this floods the log; guard with `if LOG.isEnabledFor(logging.DEBUG)` or log only on entry/exit.

---

### What Works Well

- BFS with a visited-set correctly avoids infinite loops; stop-filter logic is clean.
- `get_relatives()` is a tidy, reusable helper with proper attribution to `IsRelatedWith`.
- `reset()` properly calls `self.filt.requestreset()`.
- `gramps_target_version="6.0"` and `namespace="Person"` are correct for the base branch.
- GPLv2+ license header and author info are present.

GaryGriffin:
> **Missing po/ directory**
> All other filter addons (FilterRules/, FilterRules2/) include a po/ directory. Without it the translatable strings (_("...")) will never be extracted for localization.

If the po directory is missing when I merge a new addon, I create it, generate the template.pot, and update the strings in Weblate as part of the merge step. 

If the merge contains any string changes for existing addons, I update the template.pot and update strings in Weblate as part of the merge.

azrdev:
Thanks again for your comments!


> `selected_handles` as a class attribute means **all instances share the same set** until `reset()` reassigns it as an instance attribute. While `reset()` is called at the top of `prepare()`, any code path that calls `apply_to_one` before `prepare()` (or after `requestreset()`) reads the shared class-level set. The correct pattern (as seen in `isrelatedwithfiltermatch.py`) is to initialize these only in `prepare()`:

Actually this is no bug but coded by intention: `apply_to_one` implemented as `return person.handle in self.selected_handles`  is not harmed when accessing the class-shared empty `set()`. If I didn't initialize the attribute, a call to `apply_to_one` before `prepare` would raise an `AttributeError` exception.

> parameter is a free-text "include/exclude" toggle
> This renders as a plain text input where the user must literally type `"exclude"`. Any other value silently defaults to include. Consider dropping this parameter and creating two separate rules (one that includes the boundary, one that excludes), which is simpler and avoids the invisible string-matching.

That sounds like a good approach. Since adding the persons from the "stop" list to this filter rule's result is trivial as they already have to be gathered in a filter which can be used by a `MatchesFilter` Rule, I'd refrain from adding a copy of my Rule for that purpose.

> Missing `po/` directory

@GaryGriffin how do you generate the template.pot? I can include at least one translation myself.

> * **Missing `help_url`** in `excludesubtree.gpr.py`

I've not yet setup a wiki account, is this a necessary requisite?

> * **Missing `from __future__ import annotations`**

fixed


> * **Version `"0.6"`** — conventional starting versions in this repo are `"1.0.0"`. `"0.6"` implies five prior unreleased versions.

well, the implication is correct :)  I gather tagging the first release version as 1.0 can be usefully done with the (squash) merge of this PR

> * **Verbose debug logging inside the BFS loop**

fixed. This method needs mentioning in <https://gramps-project.org/wiki/index.php/Logging_system>

dsblank:
> Actually this is no bug but coded by intention: apply_to_one implemented as return person.handle in self.selected_handles is not harmed when accessing the class-shared empty set(). If I didn't initialize the attribute, a call to apply_to_one before prepare would raise an AttributeError exception.

I don't think your analysis is correct here, and even if it were it is confusing. 

1. `apply_to_one` should never be called before `prepare`
2. if you do need to initialize a value before `prepare` (I don't think you need to) you can put it in the constructor

dsblank:
> This method needs mentioning in https://gramps-project.org/wiki/index.php/Logging_system

I don't think we need to mention that; it is just good programming sense, and there are an infinite number of those kinds of issues.

dsblank:
> If the po directory is missing when I merge a new addon, I create it, generate the template.pot, and update the strings in Weblate as part of the merge step.

We could put this in an AGENTS.md file.

GaryGriffin:
> @GaryGriffin how do you generate the template.pot? I can include at least one translation myself.

I run `make.py gramps60 init ExcludeSubtreeFilter` to generate the template.pot file. 

Alternately, you can run 
`xgettext --language=Python --keyword=_ --keyword=:1,2c --keyword=N --from-code=UTF-8 -o po/template.pot`

This does not create any PO files, which are part of the distribution. The POT file goes to Weblate. The translated strings come from Weblate and are processed to generate PO files, which are addon-specific and bundled when the addon is distributed.

azrdev:
@dsblank to get this merged I removed the class-level initialization as you requested. Are there any remaining objections?

dsblank (approved):
## Re-review: all previously requested changes verified fixed

I checked out this PR and actually ran the addon against Gramps' real example family tree (2157 people), not just read the diff. Results:

**Fixed and verified:**
- No more `gramps.gui` import — module imports cleanly in a headless context
- Missing type imports (`Set`, `PersonHandle`, `Database`, `Person`) all present; no `NameError` at import time
- `selected_handles` is now just a type annotation, not a class-level default — confirmed two rule instances don't share state (a fresh instance has no `selected_handles` attribute until `prepare()` runs)
- `reset()` now calls `self.filt.requestreset()`
- Unknown Gramps ID no longer raises `AttributeError` — returns an empty match set gracefully (tested directly with a bogus ID)
- `gpr.py` `name` and class `name` string now match
- GPL header, import-section comments, class docstring, `set()` instead of `set([])` — all present
- `black --check` passes clean

**Functional test:** ran the BFS traversal against the real example tree — with a stop-filter matching nobody, it correctly found 1844/2157 reachable people including the start person, confirming the core algorithm is sound.

**Remaining non-blocking items** (cosmetic, not bugs):
- `help_url` still missing in `excludesubtree.gpr.py`
- `po/` directory missing, but per @GaryGriffin this is generated at merge time
- Progress bar could still slightly overstep its declared total in a narrow edge case at the filter boundary — cosmetic only

This looks ready to merge.

PR #941, opened 2026-06-05
Add NameSuite Addon (tool + gramplet) to fix, infer and audit given and patronymic names

bryndin:
See https://github.com/bryndin/namesuite


GaryGriffin:
Thanks for the submission. Couple comments

- Please review [Contributing to Gramps](https://gramps-project.org/wiki/index.php/Howto:_Contribute_to_Gramps)
- Please include in the GPR a help_url - wiki page that provides help to the addon
- Please use a real email in the GPR file.
- You need a MANIFEST to [include the required files](https://gramps-project.org/wiki/index.php/Addons_development#Package_your_addon)


When I tried to run, I got: 
`Plugin error (from 'names_tool'): No module named 'name_processor' `and 
`Plugin error (from 'patronymics_gramplet'): No module named 'name_processor'`


Running unittests:
```
python3 -m unittest NameSuite.tests.controllers.test_tool
E
ERROR: test_tool (unittest.loader._FailedTest.test_tool)
ImportError: Failed to import test module: test_tool
  File "/GrampsTest/gtk/inst/lib/python3.13/unittest/loader.py", line 137, in loadTestsFromName
  File "/Development/addons-source/NameSuite/tests/controllers/test_tool.py", line 2, in <module>
    from name_processor.controllers.tool import ToolController
ModuleNotFoundError: No module named 'name_processor'


Ran 1 test in 0.000s

FAILED (errors=1)
```

bryndin:
Thanks for the feedback.
It looks like the import structure must be based on the `NameSuite`. Somehow, I got an impression from the docs that addons themselves are brought into the namespace,  and it did work with local Gramps.
Let me look at it.

GaryGriffin:
I ran the unittest from the addons-source directory. 

By default (without a MANIFEST) only [Package your Addon](https://gramps-project.org/wiki/index.php/Addons_development#Package_your_addon) files in the Addon directory are put into the download file.

bryndin:
@GaryGriffin 

A few questions:
  - I use `pytest` for fixtures and running tests, can it be used, or should I convert tests to the vanilla `unittest`? (`reuirements.txt` inside `NameSuite` are for the tests + linting only). 
  -  Imports must be addressed via the package name  `from NameSuite.name_processor ...`, right? I'll add them.
  - Do I have to add the license block to every Python file (including tests?), or adding a package-wide `LICENSE.txt` or something would work?
  - Should I remove the existing `README.md` and screenshots?  I presume I'll be able to create a wiki page later?
  - Should I add `__init__.py` to the addon root dir (`NameSuite`)?

GaryGriffin:
> A few questions:
> 
> I use pytest for fixtures and running tests, can it be used, or should I convert tests to the vanilla unittest? (reuirements.txt inside NameSuite are for the tests + linting only).

You should use unittest instead of pytest for the Addons repo.

> Imports must be addressed via the package name from NameSuite.name_processor ..., right? I'll add them.

imports are relative to the test file location.

> Do I have to add the license block to every Python file (including tests?), or adding a package-wide LICENSE.txt or something would work?

Every python file should have the license info in the header. Be sure to include your copyright info.

> Should I remove the existing README.md and screenshots? I presume I'll be able to create a wiki page later?

If you do not want them in the download, then they are not needed in Github. They can remain if you want - your call. They could be the basis for the help_url.

> Should I add __init__.py to the addon root dir (NameSuite)?

It will be needed in every directory that has a test. If you flatten and only have a single tests directory, then there. If you keep the multi-level directory structure, you will need in each of the sub directorie that have tests.

bryndin:
@GaryGriffin   updated the PR

1. Tests use `unittest` exclusively.
Running tests from `/addons-source` works:
```
python3 -m unittest NameSuite.tests.controllers.test_tool
..
Ran 2 tests in 0.002s

OK
```
```
python3 -m unittest discover -s NameSuite
....................................................
Ran 52 tests in 0.020s
```

2. Added license headers to all Python files.
3. Remove unneeded files (e.g. README, screenshots, etc.)
4. Added `MANIFEST` file for the txt files used in tests
5. Dropped `NameSuite.` prefix from abs imports

> Please use a real email in the GPR file.

My email in GPR is real, it's the GitHub-specific email GitHub creates for devs.

GaryGriffin:
All tests worked. Tried Addon and got warnings (errors):

**Audit Given and Patronymic Names**
> WARNING: _manager.py: line 328: Plugin error (from 'names_tool'): No module named 'name_processor'

**Patronymic Suggestion**
> WARNING: _manager.py: line 328: Plugin error (from 'patronymics_gramplet'): No module named 'name_processor'

You need to include the name_processor files in the MANIFEST. Currently only the python files in the NameSuite directory are in the download, you need to add the following to the MANIFEST

`NameSuite/name_processor/*/*`

After adding that, the Audit works. The Suggestion fails with:

>   File "/Applications/Gramps6.0.6.app/Contents/Resources/lib/python3.13/site-packages/gramps/gen/plug/_manager.py", line 270, in load_plugin
>     _module = self.import_plugin(pdata)
>   File "/Applications/Gramps6.0.6.app/Contents/Resources/lib/python3.13/site-packages/gramps/gen/plug/_manager.py", line 304, in import_plugin
>     module = __import__(pdata.mod_name)
>   File "/Users/gary/Library/Application Support/gramps/gramps60/plugins/NameSuite/patronymics_gramplet.py", line 32, in <module>
>     from name_processor.views.gramplet import GrampletView
>   File "/Users/gary/Library/Application Support/gramps/gramps60/plugins/NameSuite/name_processor/views/gramplet.py", line 42, in <module>
>     class GrampletView:
>     ...<80 lines>...
>             ErrorDialog(_(title_key), message, self.gramplet.gui.get_window())
>   File "/Users/gary/Library/Application Support/gramps/gramps60/plugins/NameSuite/name_processor/views/gramplet.py", line 76, in GrampletView
>     def set_controller(self, controller: GrampletController | None) -> None:
>                                          ^^^^^^^^^^^^^^^^^^
> NameError: name 'GrampletController' is not defined
> 2026-06-10 18:01:47.433: WARNING: grampletpane.py: line 191: Error loading gramplet 'patronymic_suggestion_gramplet': skipping content



**Comments:**

- Since the `__init__.py` does not include code, it doesnt need the License. This file is only needed in the directories where there is a test file. The other directories do not need it.
- `tests/conftest.py` is not needed
- Is `male_names.txt` needed by user (included in MANIFEST?

GaryGriffin:
In `NameSuite/name_processor/views/gramplet.py` you conditionally import GrampletController, but in procedure `__init__` you always use it.

Always importing it fixes the issue and the Addon works.

GaryGriffin:
Testing Audit

- The wiki page is currently blank, so I dont have guidance on how to store Patronymic.
- I am getting unexpected results for the few relevant names I have. I suspect where I am storing the Patronymic is different from your expectation. I am storing in the Given Name at the end. Where should the Patronymic be stored for your Addon? Are you using the Name Editor --> Origin field == Patronymic, for instance?

Father:
Screenshot 2026-06-11 at 8 54 41 AM

Audit:

Screenshot 2026-06-11 at 8 55 32 AM

GaryGriffin:
Changed my naming convention to put Patronymic in  Surname (Origin=Patronymic) instead of Given Name. Audit Gramplet worked great.

bryndin:
> The wiki page is currently blank, so I dont have guidance on how to store Patronymic.

> Changed my naming convention to put Patronymic in Surname (Origin=Patronymic) instead of Given Name. Audit Gramplet worked great.

Confusingly, Gramps considers Patronymic names (e.g. Petrovich in Ivan Pertovich) a type of surname. For East Slavs it's not really a surname, but this historical hack works pretty well to the point Gramps [documents it in the wiki](https://www.gramps-project.org/wiki/index.php/Ru:Gramps_6.0_%D0%92%D0%B8%D0%BA%D0%B8_%D0%A0%D1%83%D0%BA%D0%BE%D0%B2%D0%BE%D0%B4%D1%81%D1%82%D0%B2%D0%BE_-_%D0%92%D0%B2%D0%BE%D0%B4_%D0%B8_%D1%80%D0%B5%D0%B4%D0%B0%D0%BA%D1%82%D0%B8%D1%80%D0%BE%D0%B2%D0%B0%D0%BD%D0%B8%D0%B5_%D0%B4%D0%B0%D0%BD%D0%BD%D1%8B%D1%85:_%D0%BF%D0%BE%D0%B4%D1%80%D0%BE%D0%B1%D0%BD%D0%BE_-_%D1%87%D0%B0%D1%81%D1%82%D1%8C_1#.D0.9D.D0.B5.D1.81.D0.BA.D0.BE.D0.BB.D1.8C.D0.BA.D0.BE_.D1.84.D0.B0.D0.BC.D0.B8.D0.BB.D0.B8.D0.B9). They later built a dedicated name format to visualize them properly and the hack became a feature.
image


Typing pat names manually and adjusting the surname type is time-consuming and far from fun. That's why I've built the gramplet.

For the large DBs (I have 2k+), manually copy-pasting pat names creates a lot of typos: names with the wrong gender, grandfather's names being accidentally used instead of the father's...   It also helps to catch children attached to the wrong parents, thanks to pat names serving as a second check.

> Is male_names.txt needed by user (included in MANIFEST?

It's a pat names corpus for tests. I have a quite advanced morphology engine to derive the correct endings for pat names. It requires testing on a corpus of real historical East Slavic names, as, to my knowledge, no complete set of linguistic rules exists, and some names are the official exceptions.
If we want to remove that file, we'd remove the entire test suite.

> Since the __init__.py does not include code, it doesnt need the License.

That is easy to change. I have a script that adds licenses to files.

> tests/conftest.py is not needed

Yes, good catch.  That's `pytest` leftover.

 > The wiki page is currently blank
 
I thought the Addon must be live first?

> In NameSuite/name_processor/views/gramplet.py you conditionally import GrampletController, but in procedure __init__ you always use it.

`GrampletController` there is purely for typechecking. The controller is passed to the view, but Python's duck-typing doesn't need it to be declared or imported.

> You need to include the name_processor files in the MANIFEST. Currently only the python files in the NameSuite directory are in the download, you need to add the following to the MANIFEST

I see.  This part is quite hard to figure out. The other addons I've checked consist of just one file, so they aren't a good example. When I test by copying Python files to my Gramps plugin dir, everything works fine with no warnings or errors.
I'll add sub-dirs to the MANIFEST, thanks.

bryndin:
@GaryGriffin  I think I figured it out.   Updated the repo.
Pls let me know if something else is off.  I'm developing on Linux,  while testing on Windows Gramps by copying files to `$HOME/AppData/Roaming/gramps/${GRAPMPS_VER}/plugins/NameSuite`

Here's the `addon` PR: https://github.com/gramps-project/addons/pull/15

GaryGriffin:
Thanks for the better description of Patronymic. Some answers:

- `male_names.txt `- I only meant to ask if this is needed by the user (that is, should it be in the MANIFEST and therefore in the distribution). It certainly is needed for tests. The tests are not part of the distribution to the user. After successfully using your new Addon, I agree that having the `male_names.txt` in the distribution (in the MANIFEST) is useful. 
- You can add the wiki help independently of the release. I prefer to have it when the initial PR is made so that any testers can see how to use a new Addon.
- `addons` repo - There are a number of other things that need done when publishing an Addon (mostly related to string translations), so you dont need to do a PR to the `addons` repo - only the `addons-source` repo. When I publish, there is a series of `make.py` commands used. Generally, I wait a week from PR creation to allow for user testing before merge/publish of a new Addon. You can close PR15. 

**Open Issues:**
Still an issue with GrampletController. Error is:

> Gramps6.0.6.app/Contents/MacOS/Gramps
>   File "/Applications/Gramps6.0.6.app/Contents/Resources/lib/python3.13/site-packages/gramps/gen/plug/_manager.py", line 270, in load_plugin
>     _module = self.import_plugin(pdata)
>   File "/Applications/Gramps6.0.6.app/Contents/Resources/lib/python3.13/site-packages/gramps/gen/plug/_manager.py", line 304, in import_plugin
>     module = __import__(pdata.mod_name)
>   File "/Users/gary/Library/Application Support/gramps/gramps60/plugins/NameSuite/patronymics_gramplet.py", line 32, in <module>
>     from name_processor.views.gramplet import GrampletView
>   File "/Users/gary/Library/Application Support/gramps/gramps60/plugins/NameSuite/name_processor/views/gramplet.py", line 42, in <module>
>     class GrampletView:
>     ...<80 lines>...
>             ErrorDialog(_(title_key), message, self.gramplet.gui.get_window())
>   File "/Users/gary/Library/Application Support/gramps/gramps60/plugins/NameSuite/name_processor/views/gramplet.py", line 76, in GrampletView
>     def set_controller(self, controller: GrampletController | None) -> None:
>                                          ^^^^^^^^^^^^^^^^^^
> NameError: name 'GrampletController' is not defined

Specifically, views/gramplet.py :
line 36:37

```
if TYPE_CHECKING:
    from name_processor.controllers.gramplet import GrampletController
```

line 71:

>         self._controller: GrampletController | None = None

This causes the Patronymic Suggestion Addon to fail. When I deleted line 36 and adjusted spacing on line 37, it worked.

**Cleanup**
You can remove the following files as they are not needed. Only directories with tests need this (for unittest discovery)

```
name_processor/__init__.py
name_processor/repositories/__init__.py
name_processor/models/__init__.py
name_processor/controllers/__init__.py
name_processor/views/__init__.py
name_processor/services/__init__.py
name_processor/services/audit_rules/__init__.py
tests/__init__.py
```

bryndin:
> Still an issue with GrampletController. Error is:

Thanks for the details, as I wasn't able to reproduce it with Windows Gramps 6.0.8.
Looks like Win and Mac versions are built with different Python versions, and the issue is related to [PEP 649](https://peps.python.org/pep-0649/) introduced in Python 3.14, where they changed the format for type hints.

Let me figure out the most compatible solution.

Do you know the Python version used to build your Gramps?  Don't worry if you don't, it's not a problem.

GaryGriffin:
> Do you know the Python version used to build your Gramps? Don't worry if you don't, it's not a problem.

Mac version of Gramps 6.0.6 - 3.13
Mac version of Gramps 6.1 (beta1) - 3.14

Unless you disagree, my intention is to cherry pick this addon (once published on 6.0) to 6.1
My testing has been exclusively on Gramps 6.0.6, so it is running 3.13 .

bryndin:
> Unless you disagree, my intention is to cherry pick this addon (once published on 6.0) to 6.1 My testing has been exclusively on Gramps 6.0.6, so it is running 3.13 .

My Win Gramps uses 3.14.4,   so that's why I wasn't able to see it.  Good catch!

I'm going to update the code to make it compatible with all 3.10+. Bear with me.

GaryGriffin:
As another point of reference, Mac Gramps 5.1 and 5.2  use python 3.11 .

GaryGriffin:
A new [guidelines](https://www.gramps-project.org/wiki/index.php/User:Eduralph/Sandbox/Gramps_6.0_Wiki_Manual_-_Addon_Development_-_Rules) wiki page for addon developers is being written. It is still very preliminary. Your comments would be appreciated.

GaryGriffin:
One suggestion. In the Audit Patronymic --> Audit Settings, it would be useful to add a Person Filter option. This would limit the part of the database that is searched. I have over 10K+ people but < 100 are Slavic names. 

In my case, one person married someone from Russia and limiting the database search to only their ancestors and relations would reduce the scope of the search.

bryndin:
> One suggestion. In the Audit Patronymic --> Audit Settings, it would be useful to add a Person Filter option. This would limit the part of the database that is searched. I have over 10K+ people but < 100 are Slavic names.
> 
> In my case, one person married someone from Russia and limiting the database search to only their ancestors and relations would reduce the scope of the search.

Since I've gotten back to the codebase,  I'm doing another pass, refactoring to make it better decoupled. This way, I'll be able to have normal E2E tests with minimal impact from Gramps imports (GTK, DB).
It will take me a few days.

I wish Gramps were more testable.  Not being able to run headless Gramps (no GTK) to test DB access with fixtures is a major downside.  Not having type annotation in Gramps is another safety downside. Seems like a low-hanging fruit, given the AI automation.

bryndin:
> A new [guidelines](https://www.gramps-project.org/wiki/index.php/User:Eduralph/Sandbox/Gramps_6.0_Wiki_Manual_-_Addon_Development_-_Rules) wiki page for addon developers is being written. It is still very preliminary. Your comments would be appreciated.

I need to read the new one,  but the comments on the old one:

P0: Lack of pointers to good examples. 90% of developers don't have time to read extensive docs. They would take and modify someone else's existing code. It was already like that pre-AI, now that ship had sailed. If developers find a bad/outdated example, it can spread bad practices across the codebase.

P1: Some data was outdated. AFAIR on "addons" repository, and on "requesting write access" to the Gramps repo, instead of forking it.

P1: The part on how addons are imported and what namespace is used kept me puzzled.  All the other addons were simple one-file Python scripts that weren't representative.

P2: The MANIFEST-related part was confusing,  and how the `addon-source` relates to `addon` repo.

P2: Moreover, I'd fine-tune the docs to be 50/50 Human/LLM parsable,  provide clear **must not** guidelines that could be added to the  `AGENTS.md` section.

P2: Since the guidelines mentioned above are... guidelines, and not enforcement. I'd develop tests, lints, etc.,  that would actually enforce validation on the `addon-source` repo.

These are a few from the top of my head.   I'll check the updated docs,  but it may take me some time. Need to balance my addon refactoring with other things.

bryndin:
> One suggestion. In the Audit Patronymic --> Audit Settings, it would be useful to add a Person Filter option. This would limit the part of the database that is searched. I have over 10K+ people but < 100 are Slavic names.

I may need your feedback on the speed of the audit.  How long does it take on your DB?
Mine is 2K, and it takes about 13 secs.

I'm considering adding a caching layer to speed it up.
Optionally, I can drop the confidence guessing, but that will be a downgrade.

GaryGriffin:
> I may need your feedback on the speed of the audit. How long does it take on your DB?

I have 17,500 people and it took 1 min 5 sec on an M3 Mac. 

> Not being able to run headless Gramps (no GTK) to test DB access with fixtures is a major downside.

I believe this is possible. Look at some of the recent changes that have unittests. SQLiteWithSelect may be a good example of test cases.

> I need to read the new one, but the comments on the old one:
> 
> P0: Lack of pointers to good examples. 90% of developers don't have time to read extensive docs. They would take and modify someone else's existing code. It was already like that pre-AI, now that ship had sailed. If developers find a bad/outdated example, it can spread bad practices across the codebase.
> 

I wrote a Sample addon a few years ago but it was more about using the API and not about guidelines. You bring up a good point. 

> P1: Some data was outdated. AFAIR on "addons" repository, and on "requesting write access" to the Gramps repo, instead of forking it.
> 

I updated the docs to clarify.

> P1: The part on how addons are imported and what namespace is used kept me puzzled. All the other addons were simple one-file Python scripts that weren't representative.
> 

Yours is one of the first that has subdirectories with files where this is an issue. 

> P2: The MANIFEST-related part was confusing, and how the addon-source relates to addon repo.
> 

Does [Package your addon](https://gramps-project.org/wiki/index.php/Addons_development#Package_your_addon) make this clear, or is this where the confusion lies?

> P2: Moreover, I'd fine-tune the docs to be 50/50 Human/LLM parsable, provide clear must not guidelines that could be added to the AGENTS.md section.
> 
> P2: Since the guidelines mentioned above are... guidelines, and not enforcement. I'd develop tests, lints, etc., that would actually enforce validation on the addon-source repo.
> 
We are moving to a CI/CD process for the addons-source repo, so guidelines are likely to be more rigorous. Lint and black are still suggestions. But there are testing processes for CI/CD that we are now running (not in production yet).

bryndin:
> Does [Package your addon](https://gramps-project.org/wiki/index.php/Addons_development#Package_your_addon) make this clear, or is this where the confusion lies?

Yes, I was quite confused by that section. After you helped me with MANIFEST, I just experiment with the packaging, checking the content of the `.tgz`.

  - P3: It may help to add `tar -tvf ../addons/gramps60/download/NameSuite.addon.tgz` like example for addon devs to verify the packaging.  
  - P1: Even better,  make `make` do a dry run and print the files that will go into that package. Then remove all the mentioning of `addons` repo, as it is handled by the build system.  Removing the whole `addons` concept from the `addons-source` devs, while handling it in the CI/CD behind the scene, will be a huge improvement!
  - P2: `make.py xxx build` build number increment magic kept increasing the number during my testing, and I had to be careful not to commit it by accident.  Is the goal to release `1.0.0`, then enforce `1.0.1` for unreleased code?  I'm quite confused there. 
  - P1:  (cont that P2) IMHO, removing that functionality, and adding bot-ran safeguards (tests/lints/etc) against seeing a PR with a non-updated version number will resolve the problem.

bryndin:
@GaryGriffin 
> I have 17,500 people and it took 1 min 5 sec on an M3 Mac.

Can you recheck? It should be much faster now.

---

Updated the Addon: 
  - Added DB caching
  - Added E2E tests
  - Refactor to fit the MVCS architecture better
  - Added more tests
  - Improved type annotations
  
Tested on Gramps 6.0.6 (Python 3.12) and 6.0.8 (Python 3.14)

GaryGriffin:
Retested on same M3 Mac running 6.0.6 . 15 seconds for 1st use. Same for further uses. Great

I was also tracking memory usage. The first pass increased mem 30M. Additional uses did not change this. Great

One unusual thing. The ProgressMeter counted to 17600, But I have 18100 people in my db. I was expecting the counter to go closer to the number of people.

bryndin:
> One unusual thing. The ProgressMeter counted to 17600, But I have 18100 people in my db. I was expecting the counter to go closer to the number of people.

Interesting. The progress meter shows "Processed / Total", where the "Total" uses Gramps's own `get_number_of_people()`
https://github.com/gramps-project/gramps/blob/aef9f35ec64b67f5912c5d19543060d43f270a9a/gramps/gen/db/base.py#L895

So there should not be any shenanigans on the Addon side. Mine is showing the correct `2004` that matches the number of people at the bottom of the tree view.

Try injecting that call somewhere else and logging the number. I wonder if we are seeing a Gramps bug.

Here's the SQL query it uses.  If you want to try it in the DB.
https://github.com/gramps-project/gramps/blob/aef9f35ec64b67f5912c5d19543060d43f270a9a/gramps/plugins/db/dbapi/dbapi.py#L664

GaryGriffin:
> One unusual thing. The ProgressMeter counted to 17600, But I have 18100 people in my db. I was expecting the counter to go closer to the number of people.

D'ohhh - my mistake. The Statistics addon matches the ProgressMeter for the addon. You are correct.

bryndin:
All good then,  should we merge it?

GaryGriffin:
@eduralph 
Please check that this PR aligns with your CI/CD proposed changes. I am wondering about the  `__init__.py` files in most directories.

GaryGriffin:
Please create the wiki help page.

eduralph:
@GaryGriffin , @bryndin - thanks for calling my attention to this PR

It's the first one I've seen submitted as a nested Python package (a name_processor/ library tree with controllers/models/services/repositories/views, plus a parallel tests/ tree). I love the layout, but it does mean we have to double-check if our tooling is up to snuff and it surfaced 2 items

1. Internationalisation The view modules bind translation to the Gramps core catalog, so the addon's own strings can never be translated:

_ = glocale.translation.gettext # views/base_tab.py, gramplet.py, tool.py, tool_audit_tab.py, tool_rename_tab.py

glocale.translation has no fallback to the addon's catalog (confirmed in gen/utils/grampslocale.py — only get_addon_translator attaches the addon .mo as a fallback). To solve this, switch each of those five modules to the standard addon binding (used by the addons here):

_trans = glocale.get_addon_translator(file) _ = _trans.gettext

Then commit a po/template.pot (the structure check requires it; it's a generated-but-checked-in artifact). Caveat: the stock make.py … init only extracts strings from the addon's top-level .py, so it will miss the ~56 _() strings in name_processor/views/. Until the tooling handles nested-package addons (I'm tracking that separately), generate the template over the nested tree, e.g.:

xgettext --language=Python --keyword=_ --keyword=:1,2c --keyword=N --from-code=UTF-8
-o NameSuite/po/template.pot
$(find NameSuite -name '.py' -not -path '/tests/*')

2. The test suite is nested in a way CI can't currently discover or load. Discovery only looks for /tests/test_.py (one level), so the suite isn't picked up; and the runner loads tests by dotted module name from the repo root, where name_processor isn't importable unless the addon dir is on sys.path. So even a flatter layout would still need CI to add the per-addon directory the way Gramps' loader does. I'll have to look into a fix for this.

On top of that, we have the first collision with namespaces. Because PR#950 reserves the top-level tests namespace to emulate the Gramps environment, an addon may not squat it. NameSuite does (from tests.fixtures…, from tests.fakes…, from tests.compat_mocks…), so under this contract the 941 developer has to change it — namespace the private test support inside the addon (relative imports within NameSuite.tests, e.g. from .fixtures import …, from .fakes.sync_task_runner import …), so it's addressed as NameSuite.tests.* and never collides with the reserved top-level tests

@GaryGriffin  - if you have buyers remorse, let me know and we'll undo 950. The issue with make.py needs to be addressed no matter what happens - either you Gary or me - and the translation would be something that @bryndin  would have to do.

PS minor Nit: the build relies on namespace-package fallback for name_processor/init.py (the MANIFEST's name_processor// pattern doesn't pick up the top-level package marker, only the nested files). It imports fine either way, but adding a NameSuite/name_processor/init.py line to MANIFEST would ship the explicit marker rather than depending on PEP 420.

GaryGriffin:
The normal translation code that I have see is below. Doesn't this work for this case?

```

# ------------------------------------------------------------------------
#
# Internationalisation
#
# ------------------------------------------------------------------------

try:
    _trans = glocale.get_addon_translator(__file__)
except ValueError:
    _trans = glocale.translation
_ = _trans.gettext


```

eduralph:
I tested this against Gramps' real get_addon_translator with a compiled ru/LC_MESSAGES/addon.mo: with _ bound from a nested module (name_processor/views/base_tab.py), a Russian user gets the untranslated string back — get_addon_translator(__file__) raises ValueError (wrong locale path, …/name_processor/views/locale instead of NameSuite/locale) and the except falls through to the core catalog. The identical idiom returns the translation when _ is bound from a module at the addon root, and also when anchored to the addon root from a shared name_processor/i18n.py. So it's the nesting, not the idiom.

GaryGriffin:
@eduralph 
Can you fix the following:

1. `make.py init` to handle generating template.pot on nested addons. Until then, I have to remember not to run `make.py init all` - which will overwrite the `NameSuite/po/template.pot` file with less coverage. This is not blocking this PR.
2. Update PR820 to handle test cases for nested addons. This is not blocking this PR.

I have always been leary of PR950, but I think it is the right answer in the long term. 

@bryndin 
Can you fix the following:

1. Update for translation fix
2. Generate `po/template.pot`

With these fixes, are we good to merge/publish?

dsblank:
I also have an addon for Gramps 6.2 that is nested (https://github.com/dsblank/addons-source/pull/1) Let me know if you have advice or requirements for such an organization.

eduralph:
@GaryGriffin - PR #955 has the update for make.py to deal with nested addons. PR 820 is updated with the fix.

bryndin:
## Addon Background

### Architecture

**MVCS + Clean Architecture principles of layering and dependency inversion.**

Gramps-like systems fits naturally, as they have a View layer (UI), Repository layer (Gramps DB), and Service layer (our own morphology and business logic).

* **Decoupling.** GTK and Gramps DB dependencies complicate testing. They;d be kept in the outermost layers.
* **Extensibility.** New features and GUI tabs should be easy to add without affecting unrelated parts of the codebase.
* **Code navigation.** MVCS systems are designed to look alike, so there is no need to guess where the business logic resides in a large codebase.
* **Testing.** I need unit and E2E tests that mimic user workflows. East Slavic morphology is messy and not well algorithmized. The project includes a dedicated corpus of names for validation.
* **Testability.** Python 3 supports Protocols (aka Interfaces) simplify building tests without hacky Python mocking.

### Gramps-Related Problems Worth Noting

* **P0:** Tight coupling with GTK and Gramps internals (esp the DB). I'm not aware of a way to run Gramps headlessly, nor how to instantiate a Gramps DB and use real fixtures. Let me know if I'm missing it.
* **P2:** Lack of complex addon examples. Most existing addons appear to be single-filed.
* **P1:** A flat, single-file structure is too limiting. Having business logic, libraries, GUI, DB access, caching, etc., in one file is hardly readable, reusable, or maintainable.
* **P1:** Test story. Similar to the above, how can I have a proper testing framework with multiple levels of tests for my addon that integrates well with the overall repository?
* **P1:** `addons-source` repository structure and CI/CD. My expectation was that addons should be designed as separate Python modules, each responsible for its own test suite. Test discovery from the repository root should pick them up automatically. Since Gramps runs addons in an isolated environment, separate from its core, I have a feeling the story is a bit more convoluted.
* **P2:** The lack of type annotations in Gramps limits the usefulness of `mypy`.
* **P3:** It would be nice to support `pytest`, `ruff`, and `mypy`. `pytest` has been the industry standard for quite a while.

### Localization

I'm keeping translation and string extraction in a separate module as future work. English strings should be sufficient for now, at least until there is a critical user base.

*Hope my feedback helps to improve the dev experience.  I wish to see more complex Addons and a lesser learning curve for Gramps development.*

--- 

If localization is required to keep the CI/CD pipeline working, I'll look into it.

bryndin:
> 1. Update for translation fix
> 2. Generate `po/template.pot`

1. I will replace my current core binding:
```python
_ = glocale.translation.gettext
```

With this

```python

try:
    _trans = glocale.get_addon_translator(__file__)
except ValueError:
    _trans = glocale.translation
_ = _trans.gettext
```
2. Run this
```
xgettext --language=Python --keyword=_ --keyword=:1,2c --keyword=N --from-code=UTF-8
-o NameSuite/po/template.pot
$(find NameSuite -name '.py' -not -path '/tests/*')
```
and commit `NameSuite/po/template.pot`.

Anything else?

eduralph:
Hi bryndin,

thanks for the feedback. I share some of your views on the architectural issues, but overall I think the choices made are defensible given the goals and objectives. I think what many developers lack are ADRs to understand *why* the things are the way they are. Especially plugin mechanics are tricky to implement in an open source product that has a limited audience. 

> * **P0:** Tight coupling with GTK and Gramps internals (esp the DB). I'm not aware of a way to run Gramps headlessly, nor how to instantiate a Gramps DB and use real fixtures. Let me know if I'm missing it.

Gramps is designed to run headlessly with dogtail. I've actually implemented a private harness to run Gramps in dogtail. There are scenarios that can't be instrumented, but that claim is not correct.

> * **P2:** Lack of complex addon examples. Most existing addons appear to be single-filed.

Most addons are simple. The really complex ones with value get folded into Gramps earlier or later. The addons are often things like importers or similar things. Given that we are open source, the dichotomy of "our code" vs "their code" isn't as stringent nor makes as much sense. That line of reasoning drives several of the architectural decisions around the plugin mechanism for addons. 

> * **P1:** A flat, single-file structure is too limiting. Having business logic, libraries, GUI, DB access, caching, etc., in one file is hardly readable, reusable, or maintainable.

Agreed, the question is if the logic warrants more complexity. 

> * **P1:** `addons-source` repository structure and CI/CD. My expectation was that addons should be designed as separate Python modules, each responsible for its own test suite. Test discovery from the repository root should pick them up automatically. Since Gramps runs addons in an isolated environment, separate from its core, I have a feeling the story is a bit more convoluted.

Here's the core item: Addons *do not run in an isolated environment* to Gramps, but are instead hooked into Gramps over a loosely defined  interface. I'd love a better defined interface -  indeed, I've written a GEMP with exactly that idea in mind - but it isn't implemented at this point. In the mean time, we've built a root __init__ that provides the most important items in order to decouple the addon developer from Gramps internals. 

> _Hope my feedback helps to improve the dev experience. I wish to see more complex Addons and a lesser learning curve for Gramps development._

I'm fairly new myself and had similar observations when I started. The team has been very welcoming to contribute and open for ideas - join the fray. You can find a proposal I made [in the Wiki](https://www.gramps-project.org/wiki/index.php/GEPS_049:_Versioned_Addon_API_surface_and_2_axis_lifecyle_model). I'd love your voice.

bryndin:
> Gramps is designed to run headlessly with dogtail. I've actually implemented a private harness to run Gramps in dogtail. There are scenarios that can't be instrumented, but that claim is not correct.

Any support for a test DB?  It'd be good to replace unittests with actual fixtures.
Headless Gramps is good news. Can you point me to the docs or examples?

No pressure. It's not worth redoing tests for this PR.  But I may need to develop a few more soon, so it's good to know.

> Here's the core item: Addons _do not run in an isolated environment_ to Gramps, but are instead hooked into Gramps over a loosely defined interface. I'd love a better defined interface - indeed, I've written a GEMP with exactly that idea in mind - but it isn't implemented at this point. In the mean time, we've built a root **init** that provides the most important items in order to decouple the addon developer from Gramps internals.

Since Gramps is a standalone program for personal use, fully isolating addons is probably overengineering. The main audience for such a feature would be Gramps Core devs. Migrations, given the 3rd-party addons, are hardly any fun. But if they aren't complaining, I guess it's all good.

> I'm fairly new myself and had similar observations when I started. The team has been very welcoming to contribute and open for ideas - join the fray. You can find a proposal I made [in the Wiki](https://www.gramps-project.org/wiki/index.php/GEPS_049:_Versioned_Addon_API_surface_and_2_axis_lifecyle_model). I'd love your voice.

I fixed a bug or two in Gramps at some point and skimmed through its code. It's quite a historical monolith, that, IMHO, is worth splitting into more manageable parts. But that's just a bystander point of view.

I'll be happy to look at the proposal. It may take me a few days, though.

bryndin:
btw,   should I add hardcode `po/template.pot` into `MANIFEST`?

GaryGriffin:
> btw, should I add hardcode po/template.pot into MANIFEST?

No, this is used by Weblate, not the user. The po files generated from this pot file are already included in the distribution, so no need to do that either.

eduralph:
> Any support for a test DB? It'd be good to replace unittests with actual fixtures. Headless Gramps is good news. Can you point me to the docs or examples?

You can steal the stuff from my test harness at https://github.com/eduralph/gramps-testbed-v2 if you like.

The real Gramps GUI runs headless, driven via AT-SPI/dogtail:

- Docker image ([engine/docker/Dockerfile.ubuntu](vscode-webview://1igs668572jq5jpgsdq2oabm2sdudu4tj6b0kboh0fuk4iue863l/engine/docker/Dockerfile.ubuntu)) bundles xvfb, dbus-x11, at-spi2-core, python3-pyatspi, and dogtail — the X-framebuffer headless stack.
- [engine/interface/base.py](vscode-webview://1igs668572jq5jpgsdq2oabm2sdudu4tj6b0kboh0fuk4iue863l/engine/interface/base.py) — GrampsAppTestCase: launches gramps -u -O TestTree under xvfb, waits for the main window frame, drives it through the accessibility tree (dogtail root/keyCombo), one Gramps process per test class, screenshots on failure, per-class timeout, and resets ~/.gramps/grampsrc between launches.
- Runners + gates: [run-interface.sh](vscode-webview://1igs668572jq5jpgsdq2oabm2sdudu4tj6b0kboh0fuk4iue863l/engine/scripts/ubuntu/run-interface.sh) (core), run-addon-interface.sh (addons), run-verify-interface.sh (per-fix), wired as the interface C4/T3 gates.
- A real interface test suite: test_smoke.py plus bug repros (2092, 8594, 11786, 13716, 13864, 14100, quiltview, combinedview…).

The DB harness you can find at 

- Canonical "TestTree" — run-interface.sh:161 does gramps -C TestTree -i …/gramps/example/gramps/example.gramps, i.e. it creates a tree and seeds it from Gramps' own example.gramps before the GUI suite runs.
- Per-test fixtures in [engine/interface/data/](vscode-webview://1igs668572jq5jpgsdq2oabm2sdudu4tj6b0kboh0fuk4iue863l/engine/interface/data/) — bug_0014100_minimal.gramps, quiltview_minimal.gramps — small .gramps trees for focused repros (a test sets TREE_NAME/imports its own minimal DB).
- Unit-side DB: the shared harness [tests/gramps_test_env.py](vscode-webview://1igs668572jq5jpgsdq2oabm2sdudu4tj6b0kboh0fuk4iue863l/tests/gramps_test_env.py) provides GrampsDbTestCase; addon unit tests mostly mock DbTxn, while ones like Sqlite round-trip example.gramps and TMGimporter drives a real in-memory DB via make_database.

> Since Gramps is a standalone program for personal use, fully isolating addons is probably overengineering. The main audience for such a feature would be Gramps Core devs. Migrations, given the 3rd-party addons, are hardly any fun. But if they aren't complaining, I guess it's all good.

I'm a great fan of overengineering but I can understand if others aren't, lol. 

> I fixed a bug or two in Gramps at some point and skimmed through its code. It's quite a historical monolith, that, IMHO, is worth splitting into more manageable parts. But that's just a bystander point of view.

Oh, it's absolutely a monolith and there is a lot of room for improvement. OTOH, given the market, it is a great piece of software. I've also been thinking what a UI Designer could make of the workflows, etc. 
 
> I'll be happy to look at the proposal. It may take me a few days, though.

No worries - there's also no promise it will actually be implemented but you came out at the same spot I did when looking at the current setup.

bryndin:
> Please create the wiki help page.

@GaryGriffin 
I've created an account on wiki,  waiting for the password to be mailed after confirming the email.  It is surprisingly taking too long for an automated system.  Resetting the password doesn't work either, reporting that there is no such user.

These extra steps could potentially be automated. For example, GitHub uses the repo README as the docs. CI/CD can transform it into the wiki format.

GaryGriffin:
> I've created an account on wiki, waiting for the password to be mailed after confirming the email. It is surprisingly taking too long for an automated system. Resetting the password doesn't work either, reporting that there is no such user.

@emyoulation Brian, can you check on this wiki password request. As I recall, someone else recently had a similar issue. I dont know how it was resolved.

emyoulation:
That will have to be checked by @sam-m888 or @Nick-Hall . They have Wiki Admin rights.

bryndin:
@eduralph let me know where to post the feedback to your GERP, and we can move it there.

---

TL;DR:  IMHO, what you're describing is **the** Gramps Public API. Taking that perspective may straighten the whole GERP.

> Part 1: The Addon API surface
> The Addon API is the set of names, signatures, and behaviours that Gramps core promises to keep stable for addon authors. It is not the entire Gramps codebase; an addon poking at gramps.gen.db.generic._Generic.__some_private__ has always been on its own.

Access to private internals ends up being the core maintainer's problem on every migration. I'd restrict it. If some functionality is missing, it'd be added to the API.
Full decoupling could be hard for the non-isolated in-process Python addons, and may not be worth the effort and added complexity.

> Versioning scheme

This all makes sense. Semver is standard in Python packages. This section could be slimmed down. If I'm not missing any internal details, the message is
  - Semver 2.0; API version is decoupled from the Gramps version.

> Detecting API changes
> Tier 1 — Define the surface explicitly. 

If you ask me,  I'd approach this one differently. Python has Protocols that were designed specifically to declare APIs/Interfaces. ABCs are the historical option, but Protocols are superior. They provide static typing (that Gramps is very past due on), and allow Addons to be isolated from the Gramps internals.

Having a whitelist of imports is
	- P0:	Provides poor isolation (Python doesn't enforce private scope). You allow an import, you leak the internals.
	- P0:	Very confusing for AI coders
	- P1:	Confusing to human developers
	- P2:	Harder to enforce

Pros:  Gramps already has a DB API that can and should be converted to Protocols anyway.
Cons:  That API can grow pretty darn big.

> Tier 2 — Mechanical detection via griffe

Great point, I'd even extend this tooling section. With the modern speed of development, the guardrails are a must.

> Part 2: Lifecycle (two orthogonal axes)

Don't have a background, not going to comment on it. These things are highly opinionated and tend to consume time in discussions. I prefer not to go there unless there is a clear pain point everyone agrees on.

> MAINTAINERS wiki integration

Oh my, pls replace Wiki with the GitHub code-based (or some automated) solution. I know it's out of scope for this GERP, just throwing it here.

> Part 3: Source repository structure

Hide the **entire** concept of `addon` repo from users. It must be an internal detail of the Gramps build system. The packaging must be testable using the `addon-source` providing tools only.

> Open questions
> Whether the dual-version scheme (Gramps release version + addon_api) 

For web services, libraries, etc., the Public API defines the main (and only) version, as that's what customers face. For Gramps as a desktop app, it could be different. Users may see a major change in functionality without a change in the API. I'm less of an expert there, but the independent Gramps & API versioning makes sense.

> Default maintenance_state for legacy addons during Phase 3 rollout.
> BROKEN visibility. 

If agreement is slow to reach,  I'd kick it into a separate low-priority proposal.

> Cross-major-version backport workflow. 

Do what Python does: support old versions  :-)

> Developer tooling around registration

I'd raise priority on this one heavily. Registration is a pain point for addon developers. It's a Python file being non-Python. No tests, linting, or type hinting. If I, as an addon dev, mess it up, I can't test it within the `addon-source` repo.

*Hope it helps*

GaryGriffin:
> > 1. Update for translation fix
> > 2. Generate `po/template.pot`
> 
> 1. I will replace my current core binding:
> 
> ```python
> from gramps.gen.const import GRAMPS_LOCALE as glocale
> _ = glocale.translation.gettext
> ```
> 
> With this
> 
> ```python
> from gramps.gen.const import GRAMPS_LOCALE as glocale
> 
> try:
>     _trans = glocale.get_addon_translator(__file__)
> except ValueError:
>     _trans = glocale.translation
> _ = _trans.gettext
> ```

@eduralph Is this complete for the CI/CD process? The template.pot seems reasonable and I am merging your make.py changes so that point is doubly covered.

eduralph:
@bryndin — small but important: get_addon_translator(__file__) loads the catalog from dirname(__file__)/locale, so binding _ in the name_processor/views/* modules makes it look under name_processor/views/locale, which doesn't exist — the .mo files live at NameSuite/locale/. I tested it against real Gramps with a compiled ru/LC_MESSAGES/addon.mo: bound in a nested module, a Russian user gets the untranslated English string (the ValueError falls through to the core catalog); anchored at the addon root it returns the translation. So the idiom needs to be anchored at the addon root, not each module's __file__.

Cleanest: bind _ once in a new name_processor/i18n.py:


import os

\# this file is <addon>/name_processor/i18n.py → the addon root (which holds locale/)
\# is two dirs up; get_addon_translator uses only dirname(path), so "_" is a placeholder.
_ADDON_ROOT = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
try:
    _trans = glocale.get_addon_translator(os.path.join(_ADDON_ROOT, "_"))
except ValueError:
    _trans = glocale.translation
_ = _trans.gettext
then in each of the five view modules, replace the glocale import and the try/except _trans block with:


from name_processor.i18n import _
I verified this loads NameSuite/locale/.../addon.mo correctly ("Given name" → "Имя").

bryndin:
@eduralph 
I see a few problems there:
  - It's a workaround for this particular addon (and the model layout)
  - It's not reusable, and harder to maintain
  - Addon devs have to understand how Gramps translation works (while they're better off improving their addons)

While I can do it to unblock this PR, the better options could be:

#### Option 1: Make Gramps `get_addon_translator()` handle it.
Say, add the logic to do `locale/` dir discovery.
_Pros:_
  - Easy to document
  - Simplest solution for addon devs, they can blindly reuse existing code samples
  - Takes effort off the `addon-source` repo maintainers, simplifies their CRs
  - Easy to make changes without migrating all the addons

_Cons:_
  - It's a Gramps level change

#### Option 2: Make each model a module (`NameSuite` will be the module)
Translation logic is a copy-paste `i18n.py` file, placed at the root of the addon module. Implement auto-discovery there.
_Pros:_
  - Simple copy-paste of the reusable file

_Cons:_
  - Still IMHO a redundant step.

---

Separate question to both @GaryGriffin and @eduralph 

Would it be better to make Addons the real Python modules?
`from NameSuite.name_processor.models.infer import ...`
instead of
`from name_processor.models...`

IMHO, this would streamline the development and make everyone's life much simpler.
I have quite some experience with Python, and I find the current system **highly** counterintuitive and puzzling. The struggle with translations is yet another indicator. We create workarounds for problems that may not need to exist. If needed, the Addon manager could be a simple wrapper around `pip` or a similar standard tool.

Modules must be self-contained,  providing and handling their own tests, test frameworks, and whatnot. It's not the `addon-source` responsibility to deal with any of that, maybe besides test self-discovery.

This'd make life way easier for both addon devs and Gramps devs/maintainers.

dsblank:
> I have quite some experience with Python, and I find the current system highly counterintuitive and puzzling. The struggle with translations is yet another indicator. We create workarounds for problems that may not need to exist. If needed, the Addon manager could be a simple wrapper around pip or a similar standard tool.

I was thinking the same thing. We have the foundation of our own mini-installer now; we could use that infrastructure... or something even easier.

eduralph:
@bryndin 

> I'd raise priority on this one heavily. Registration is a pain point for addon developers. It's a Python file being non-Python. No tests, linting, or type hinting. If I, as an addon dev, mess it up, I can't test it within the `addon-source` repo.
> 
> _Hope it helps_

Thanks for your ideas, I think there is some really good ideas to include in the GEPS - I'll revise it.

Nick-Hall:
@GaryGriffin If you are thinking about changing the way addon translations are handled, then you should also consider the possibility of compiling and distributing the aggregated translations instead of extracting them.  This would have the benefit of separating the translation files from the individual addons.  Users wouldn't see a new addon version for just a translation update.

GaryGriffin:
Should the translation topic move to a separate discussion rather than this PR? It is above my Python knowledge to suggest a path forward. I think the appropriate people ( @bryndin @eduralph @dsblank at a minimum ) need to drive to a plan for how to handle addon translations, as they seem already involved with it.

dsblank:
The translation question should be made in "discussions" in the gramps repo.

eduralph:
@bryndin - your Option 2 is the thing I proposed — both bind _ once, anchored at the addon root (the dir that holds locale/), and both let the view modules just from … import _. The reason it can't be a per-module __file__ binding isn't a stylistic choice on my end: get_addon_translator(filename) derives the catalog dir as os.path.dirname(os.path.abspath(filename)) + "/locale" (gen/utils/grampslocale.py). So a binding in name_processor/views/base_tab.py looks under name_processor/views/locale/, which doesn't exist, and a Russian user silently gets English. Anchoring at the addon root is what makes it resolve NameSuite/locale/…/addon.mo.

bryndin:
@GaryGriffin @eduralph @dsblank @Nick-Hall 

I jammed on the options for the i18n design, here's a [high-level RFC](https://gist.github.com/bryndin/275ea3c448b448e904fd41799bc8c1b2) (it's human-readable): 
  - **P0**: reduce the burden on addon developers
  - **P1**: decouple translation updates from code releases

If you are time restricted, check the [best option](https://gist.github.com/bryndin/275ea3c448b448e904fd41799bc8c1b2#alternative-c-decoupled-over-the-air-ota-language-packs
):

@Nick-Hall, 
  - Is self-hosting `Weblate` an option?  Optionally, Weblate might be willing to host it for free for a popular open-source project.
  - Would Core Gramps translation benefit from the same (lang-pack) approach?  (I'm not natively familiar with Gramps internals)
  
We can move this discussion elsewhere.
_Note: Wiki registration is broken. I can't register there. No write access._

GaryGriffin:
Your **Translation Workflow** for Addons was correct up to v5.2 . However, version 6.0 translations moved to a Weblate process (Gramps core moved to a Weblate process during 5.x), found at https://hosted.weblate.org/projects/gramps-project/addons/ . Both Gramps core and Addons have separate Weblate projects for translations. The current translation process for Addons is:

- Each addon has a po/template.pot. I run `make.py .. init` to create if it does not exist. 
- If a PR contains a change in a text string, I run `make.py .. aggregate-pot` to generate an updated consolidated strings file for Weblate. This file is at `addons-source/po/addons.pot`. Any string that is in the Gramps project string list (found at https://github.com/gramps-project/gramps/blob/maintenance/gramps60/po/gramps.pot ) is removed as a duplicate.
- Translators make changes in Weblate for the Addons project (there is a separate project for Gramps). Each change appends to a single Addons PR generated by Weblate. The current PR is https://github.com/gramps-project/addons-source/pull/938 . 
- Every few weeks, I merge this PR. It updates the addons-source/po/\*.po files (one per language). I also run `make.py .. extract-po` that updates every addon's po files stored in addons-source/.../po/\*-local.po
- When a new download is created, it includes the latest po/\*-local.po files.

bryndin:
Thanks @GaryGriffin ,   I'll update the RFC.   The "best solution" should stay the same,  but the details need a rewrite.

bryndin:
@eduralph   updated the i18n logic,  it resides in the dedicated [NameSuite/name_processor/views/i18n.py](https://github.com/gramps-project/addons-source/pull/941/changes/47c81c4ad57579a46b81cb5f505c2f972b0d45ca#diff-def19adf549f6a59576b1c0070ed7c8d8fe0332d6bf12a9e2485cab408512b9f)

bryndin:
"Gramps Addons - Current l10n process"
https://gist.github.com/bryndin/d95b239902c66f6072eb86732f89810a

This is the summary of the current process, as far as I understand it. @GaryGriffin is the process correct? Let me know if you have better documentation.

bryndin:
Here's the updated RFC:  https://gist.github.com/bryndin/b0fbd268b4785d0e6600b8270daca75e
The top part is the generalized architecture, followed by the concrete implementation in Gramps.

Description of the current L10n process in Gramps:
https://gist.github.com/bryndin/d95b239902c66f6072eb86732f89810a

GaryGriffin:
Suggest you start a discussion at https://github.com/gramps-project/gramps/discussions as this will require gramps core changes. This also gives a wider discussion area.

GaryGriffin:
@eduralph does the PR align with the CI proposed process and translations? From your perspective, it is ready to merge?

eduralph:
- Structure + template.pot: aligned.
- CI/make.py flow: the make.py changes he's merging cover the aggregate-pot/Weblate side; structurally the addon (.gpr.py, MANIFEST, po/template.pot, tests/) conforms.
- Translations: not ready to merge as-is — the runtime binding resolves the wrong locale/ dir, so translations silently won't apply. Needs the one-line i18n.py fix, after which it's aligned.

_trans` = glocale.get_addon_translator(os.path.join(_ADDON_ROOT, "name_processor.gpr.py"))

would do.

bryndin:
> * Structure + template.pot: aligned.
> * CI/make.py flow: the make.py changes he's merging cover the aggregate-pot/Weblate side; structurally the addon (.gpr.py, MANIFEST, po/template.pot, tests/) conforms.
> * Translations: not ready to merge as-is — the runtime binding resolves the wrong locale/ dir, so translations silently won't apply. Needs the one-line i18n.py fix, after which it's aligned.
> 
> _trans` = glocale.get_addon_translator(os.path.join(_ADDON_ROOT, "name_processor.gpr.py"))
> 
> would do.

@eduralph  should there be a repository-wide test for that,  so that contributors like me can self-test it?

eduralph:
> > * Structure + template.pot: aligned.
> > * CI/make.py flow: the make.py changes he's merging cover the aggregate-pot/Weblate side; structurally the addon (.gpr.py, MANIFEST, po/template.pot, tests/) conforms.
> > * Translations: not ready to merge as-is — the runtime binding resolves the wrong locale/ dir, so translations silently won't apply. Needs the one-line i18n.py fix, after which it's aligned.
> > 
> > _trans` = glocale.get_addon_translator(os.path.join(_ADDON_ROOT, "name_processor.gpr.py"))
> > would do.
> 
> @eduralph should there be a repository-wide test for that, so that contributors like me can self-test it?

No, unfortunately not but is a good idea for a rule in a precommit hook. Definitely something to add to it once we get a CI process in place.

bryndin:
> No, unfortunately not but is a good idea for a rule in a precommit hook. Definitely something to add to it once we get a CI process in place.

My understanding, a direct implementation would require all the Gramps dependencies, which is a no-go for the addon contributors.
I wonder if that could be achieved via some tricky mocking,  but I'm not an expert and don't have an answer.

bryndin:
> _trans` = glocale.get_addon_translator(os.path.join(_ADDON_ROOT, "name_processor.gpr.py"))

@eduralph updated, can you pls check?

eduralph:
@bryndin - all green, no more nit-picks

bryndin:
I've posted the **RFC: Redesigning the L10n Architecture for Gramps Addons**
https://github.com/gramps-project/gramps/discussions/2462

PR #942, opened 2026-06-05
Fix PostgreSQL collation check to support versions before 12

dsblank:
## Summary

- Replace the DROP+CREATE collation pattern with a `pg_collation` existence
  check — only creates the collation when absent
- Removes dependency on `CREATE COLLATION IF NOT EXISTS` (PostgreSQL 12+) and
  on `DROP COLLATION IF EXISTS` executed on every connection open
- Works on all PostgreSQL versions that support `CREATE COLLATION` (9.1+)
- Also includes `column_exists` and `drop_column` methods needed for schema
  migration support, and applies black formatting throughout

## Background

Users on PostgreSQL < 12 hit a syntax error on startup because the previous
code called `DROP COLLATION IF EXISTS` then `CREATE COLLATION`, which still
required `IF NOT EXISTS` in an earlier version of this file. The pg_collation
catalog query is the idiomatic, version-agnostic approach.

Fixes #65803.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/PostgreSQL/postgresql.py
+++ b/PostgreSQL/postgresql.py
@@ -23,39 +23,41 @@
 Backend for PostgreSQL database.
 """
 
-#-------------------------------------------------------------------------
+# -------------------------------------------------------------------------
 #
 # Standard python modules
 #
-#-------------------------------------------------------------------------
+# -------------------------------------------------------------------------
 import psycopg2
 import os
 import re
 
-#-------------------------------------------------------------------------
+# -------------------------------------------------------------------------
 #
 # Gramps modules
 #
-#-------------------------------------------------------------------------
+# -------------------------------------------------------------------------
 from gramps.plugins.db.dbapi.dbapi import DBAPI
 from gramps.gen.utils.configmanager import ConfigManager
 from gramps.gen.config import config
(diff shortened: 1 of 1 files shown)


dsblank:
See discussion here: https://gramps.discourse.group/t/database-error-on-gramps-6-0-6-update-line-1-create-collation-if-not-exists-de-de-locale-de-de-utf-8/9580

GaryGriffin:
This PR is for master. Shouldnt this be to gramps60 and then cherry picked to gramps61?  Your other PostgreSQL [PR943](https://github.com/gramps-project/addons-source/pull/943) is on gramps60, for instance. 

AFAIK, only @Nick-Hall  deals with master branch.

dsblank:
> Shouldnt this be to gramps60

My bad. (Actually it was Claude but I will instruct him properly). I'll fix.

dsblank:
@GaryGriffin fixed!

Nick-Hall:
> AFAIK, only @Nick-Hall deals with master branch.

We haven't used the _master_ branch in _addons-source_ for years.

PR #947, opened 2026-06-09
HasTagSubstr: version target to 6.0

GaryGriffin:
Fix addon to have correct version target on gramps60 branch


hgohel:
Thanks @GaryGriffin 
For reference, this is [bug 14246](https://gramps-project.org/bugs/view.php?id=14246)

PR #955, opened 2026-06-22
make.py: extract i18n strings from nested-package addons

eduralph:
## What

`make.py … init` (and the `as-needed` pot rebuild) collected translatable
strings with `glob.glob("{addon}/*.py")`, scanning only an addon's **top-level**
`*.py`. An addon laid out as a nested Python package keeps its `_()` strings in
sub-package modules, so `po/template.pot` silently omitted them — and the
addon-structure check, which only verifies the file exists, can't tell.

## Change

- Add `addon_py_files(addon)`: top-level files first, in the exact order
  `glob.glob("{addon}/*.py")` yields them (so a flat addon's `template.pot` is
  unchanged), then nested-package modules (sorted, deterministic), excluding the
  addon's own `tests/` tree at any depth (test strings are not shipped or
  translated).
- Use it at both `xgettext` call sites (`init` and the `as-needed` rebuild).
  Downstream consumers (`compile_addon` / `aggregate_pot` / `extract_po`) read
  the generated pots and are unaffected.

## Validation

Against the NameSuite submission (#941): the old top-level-only extraction
captured **5** strings (from the `.gpr.py` alone); the new recursive extraction
captures **55** across 6 files, including the five `name_processor/views/*` UI
modules. Flat addons are unchanged — `addon_py_files` returns the same file
list, in the same order, as today's `glob.glob`.

`tests/test_i18n_extraction.py` drives `make.py init` over synthetic addon
trees: nested strings are extracted, the addon's `tests/` tree is excluded, and
a flat addon references only its top-level files.

### DIFF
--- a/make.py
+++ b/make.py
@@ -189,6 +189,28 @@ def cleanup(addon_dir):
     shutil.rmtree("%s/locale" % addon_dir, ignore_errors=True)
 
 
+def addon_py_files(addon):
+    """Return the addon's .py files for xgettext string extraction.
+
+    Top-level files come first, in the exact order ``glob.glob("{addon}/*.py")``
+    yields them, so a flat addon's template.pot is unchanged. Nested-package
+    modules follow (sorted, for deterministic output), excluding the addon's own
+    tests/ tree at any depth (test strings are not shipped or translated).
+    """
+    top = glob.glob(f"{addon}/*.py")
+    top_set = set(top)
+    nested = []
+    for dirpath, dirnames, filenames in os.walk(addon):
+        if "tests" in dirnames:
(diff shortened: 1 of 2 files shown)


GaryGriffin:
Should I cherry pick this to gramps61 or will that fail?

eduralph:
Cherry-pick it — it won't fail. I tested the cherry-pick onto maintenance/gramps61: it applies cleanly (no conflicts), make.py compiles, and the tests/test_i18n_extraction.py regression passes there. The two xgettext extraction sites and the helper's insertion point are identical on gramps60 and gramps61 (only a 2-line offset on the as-needed site, which the patch context handles), so the change is correct on gramps61, not just applicable.

PR #959, opened 2026-06-28
Gram.py Script: add evaluate_expression, fix bugs, add tests, target gramps 6.1

dsblank:
## Summary

- Add `evaluate_expression()` method — runs code in the full GrampyScript scope and returns stdout immediately (used by GrampsAssistant to introspect the database without user interaction)
- Add `_instance` singleton — allows other addons (e.g. GrampsAssistant) to get a reference to the active GrampyScript gramplet
- Add `database` variable to script execution scope, giving scripts direct access to the Gramps DB object
- Fix `save_script` crash with `FileNotFoundError` when no filename has been set yet — falls through to Save As instead
- Fix `get_columns()` returning `None` implicitly when no `row()` call is found — now consistently returns `[]`
- Update `gramps_target_version` to `6.1`
- Add `tests/` with unit tests for `DataDict2`, `DataList2`, `NoneData`, and `get_columns`

## Dependents

**#960 (GrampsAssistant) depends on this PR** — it uses `_instance` and `evaluate_expression()` added here. Merge this first.

## Test plan

- [ ] Run `python -m pytest tests/` in `GrampyScript/` — all 42 tests pass
- [ ] Open GrampyScript sidebar panel, click Script → Save without having saved before — should open Save As dialog instead of crashing
- [ ] Run a script with no `row()` calls — `get_columns` returns `[]` (no crash)

### DIFF
--- a/GrampyScript/GrampyScript.gpr.py
+++ b/GrampyScript/GrampyScript.gpr.py
@@ -23,7 +23,7 @@ register(
     name=_("Gram.py Script"),
     description=_("Run a special Gramps Python script"),
     status=STABLE,
-    version = '0.0.5',
+    version = '0.0.3',
     fname="GrampyScript.py",
     authors=["Doug Blank"],
     authors_email=["doug.blank@gmail.com"],
--- a/GrampyScript/GrampyScript.gpr.py
+++ b/GrampyScript/GrampyScript.gpr.py
@@ -23,7 +23,7 @@ register(
     name=_("Gram.py Script"),
     description=_("Run a special Gramps Python script"),
     status=STABLE,
-    version = '0.0.3',
+    version = '0.0.6',
     fname="GrampyScript.py",
     authors=["Doug Blank"],
     authors_email=["doug.blank@gmail.com"],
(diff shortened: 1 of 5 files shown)


GrampyScript/GrampyScript.gpr.py
```diff
@@ -23,7 +23,7 @@
     name=_("Gram.py Script"),
     description=_("Run a special Gramps Python script"),
     status=STABLE,
-    version = '0.0.5',
+    version = '0.0.3',
```
dsblank:
```suggestion
    version = '0.0.6',
```

GaryGriffin:
No need to increment version number in gpr - that happens when published.

dsblank:
Yeah, it was just that the number went down from a previous version due to me moving between branches, and didn't want to make it even more confusing :smile:

GaryGriffin:
Changes in GrampyScript are in both this PR and PR960. Shouldn't this be in just this PR?

dsblank:
> Changes in GrampyScript are in both this PR and PR960. Shouldn't this be in just this PR?

Yes, fixed.

PR #961, opened 2026-06-28
Handles references without page number. Slight cleanup of regex

Ludwig-T:


GaryGriffin:
Adding template.pot when merging. This will allow Weblate to translate the 2 strings.

PR #968, opened 2026-07-01
Filter rule to match parents of family filter

jjdup:


GaryGriffin:
Should this be on branch gramps60 and then cherry picked also to gramps61?

jjdup:
I rebased to 6.0

PR #970, opened 2026-07-03
BirthdaysGramplet: sort persons by month and day of birth

javadr:
Previously entries were sorted by proximity to today's date (nearest birthdays first). Now they are sorted by month then day, giving a consistent calendar-order listing.


GaryGriffin:
The documentation of this gramplet at https://gramps-project.org/wiki/index.php/Addon:BirthdaysGramplet describes upcoming birthdays, not starting at the beginning of the year. This revision seems to be contrary to that.

> This Birthdays Gramplet displays the upcoming birthdays based around the current date.

javadr:
You are right, I will close this one and make a new PR with an option from configuration to activate this option. The default would be its original behavior.

PR #973, opened 2026-07-04
Add DateOfDeathGramplet - gramplet listing death dates sorted by …

javadr:
A new gramplet that displays all deceased persons' death dates, sorted by month then day, with clickable person links and age at death.

Inspired by the BirthdaysGramplet, adapted for death dates.


javadr:
Here's a PR description for the changes:

- Sort mode dropdown — config to switch between "Proximity to current date" (today → upcoming → past) and "Month and day" (calendar order), mirroring the proven BirthdaysGramplet pattern.

- Cross-calendar fix — gregorian(date_of_death) is now called before constructing death_this_year, fixing sort order when the death date uses a non-Gregorian calendar (e.g., Jalali/Islamic).

javadr:
I'd like to have it on both the 6.0 and 6.1 branches. Since I have Gramps 6.0 on my local machine, that's my preferred target. I haven't tested it extensively, but I believe it should work with earlier versions as well.

GaryGriffin:
Thanks for this addition. Couple comments:
- Please add author and author_email to GPR
- help_url does not exist. Typically it is Addon:NameOfGramplet
- Translations are generated by Weblate and will be overwritten. Only po/template.pot is used (sent to Weblate) and Weblate returns all of the xx-local.po files based on its translated strings.

javadr:
authors and authors_email have been added.

PR #974, opened 2026-07-04
Gramplet/birthdays gramplet

javadr:
Here's a PR comment for the changes:

BirthdaysGramplet: add sort mode option, set default to proximity
- Added a "Sort birthdays by" dropdown option in build_options() with two modes:
  - Month and day — sort by calendar date (Jan 1 → Dec 31)
  - Proximity to current date — sort by how soon the next birthday is (upcoming first)
- Default mode set to proximity
- Filled all empty msgstr entries across all 22 locale files via Google Translate API — every translation file is now 100% translated

### DIFF
--- a/BirthdaysGramplet/BirthdaysGramplet.gpr.py
+++ b/BirthdaysGramplet/BirthdaysGramplet.gpr.py
@@ -3,6 +3,7 @@
 #
 # Copyright (C) 2010  Peter Potrowl <peter017@gmail.com>
 # Copyright (C) 2019  Matthias Kemmer <matt.familienforschung@gmail.com>
+# Copyright (C) 2026  Javad Razavian <javadr@gmail.com>
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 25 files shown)


javadr:
The latest commit ([71261d8](https://github.com/gramps-project/addons-source/pull/974/commits/71261d883732a160a3cf468ba9236f4dd500b395))  is a hot fix to a 16 year-old bug!

It fixes proximity sort for non-Gregorian birth dates

Date(y, m, d) always creates a Gregorian date, but the code was extracting month/day directly from birth_date — so a Persian date of Farvardin 1 (~March 21) became January 1 Gregorian, producing wrong proximity values and broken sort order.

Fix: convert birth_date to Gregorian via gregorian() before constructing birthday_this_year, so month/day values match the calendar today uses.

GaryGriffin:
String translations are done in Weblate and then pushed to all of the addons. The changes to xx-local.po will be overwritten by Weblate.

GaryGriffin:
Do you want this on gramps60 branch instead of 61? 

All Addons PRs for gramps60 are being cherry picked to 61 also.

javadr:
Since I have Gramps 6.0 on my local machine, I would prefer to have it available on both branches.

I'm not completely sure, but I have a feeling it should work with earlier versions as well.

dsblank:
Great find with 71261d8 fix! I'm going to make sure we don't make that mistake in other places in gramps core, or other addons. Thanks!

dsblank:
Added a new method in https://github.com/gramps-project/gramps/pull/2459, but it won't be available until gramps 6.2.

GaryGriffin:
The cherry pick to gramps60 failed. You will need to create a new PR for these changes for the gramps60 branch.

PR #976, opened 2026-07-04
Update associationstool.py to v1.2.1(gramps60)

romjeromealt:


GaryGriffin:
The diff from the 5.2 version is the VERSION_TUPLE check. Why does this have to change? I thought the 5.2 version would work on any version > 5.2 and would not need any changes.

romjeromealt:
oops, you are right, my bad! I will keep the logic set by @emyoulation

PR #977, opened 2026-07-06
GrampyScript: fix AttributeError when committing edited records

dsblank:
## Summary
- `callback()` in `GrampyScript.py` called the non-existent `db._table(class, "commit_func")` instead of `db._get_table_func(class, "commit_func")`
- Any script that edits data (e.g. sets a field after calling `begin_changes()`) crashed with:
  ```
  AttributeError: 'SQLite' object has no attribute '_table'
  ```
- Fixes the typo so edits actually commit to the database

## Test plan
- [x] Ran a GrampyScript script that clears `attribute_list` on all citations after `begin_changes()`; previously crashed with the AttributeError, now commits successfully

### DIFF
--- a/GrampyScript/GrampyScript.py
+++ b/GrampyScript/GrampyScript.py
@@ -854,7 +854,7 @@ for person in people():
         if action == "set":
             if "_class" in data:
                 if self.db._get_table_func(data["_class"]):
-                    method = self.db._table(data["_class"], "commit_func")
+                    method = self.db._get_table_func(data["_class"], "commit_func")
                     method(data._object, self.TRANSACTION)
                 else:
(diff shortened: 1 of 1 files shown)


dsblank:
@GaryGriffin you can make this same change to maintenance/gramps61 branch too (or let me know if you want a separate PR). Thanks!

GaryGriffin:
I am cherry picking EVERY 6.0 PR to 6.1 at this point. Once 6.1 is officially released, I will stop the auto cherry pick. 

Only if the cherry pick fails will a separate PR for 6.1 be needed.

dsblank:
I have an update for 6.1 so I'll include it there.

PR #978, opened 2026-07-06
GrampyScript: bundled example scripts, Open-dialog previews, UI polish

dsblank:
## Summary
- Adds `GrampyScript/scripts/` with 13 worked example `.gram.py` scripts covering iteration, filtering, family relationships, pie/histogram charts, batch editing with transactions, CSV-ready reports, active-person summaries, selected-rows/data-quality checks, importing a shared `.py` helper module, reusing a Gramps custom filter, and deleting an object
- This folder is also the default location for Script > Open... / Save As..., so a user's own saved scripts naturally collect next to the examples
- Adds a preview pane to the Open dialog: highlighting a `.gram.py` file shows its (translatable) title/description, falling back to the file's own leading comment for uncatalogued scripts
- Descriptions live in `script_descriptions.py`, kept translatable via the addon's existing xgettext pipeline without cluttering the example scripts themselves with `_()` calls
- Adds `update_script_descriptions.py`, a dev tool that keeps that catalog in sync with `scripts/` (adds stubs for new files, drops stale entries, warns on title drift) without touching hand-written text it didn't generate
- UI tweaks: moved the Execute button directly below the script editor (was far below, under the results tabs), added a persistent filename label to the status area, and prompt to save unsaved changes before New or Open
- Fixes an `AttributeError` in `callback()` (`db._table` → `db._get_table_func`) that crashed any script performing an edit after `begin_changes()` — same fix as #977, needed here since the new `06_mark_unsourced_people_private.gram.py` example exercises that exact path
- Adds `import` support: the scripts folder and the currently open script's own folder are added to `sys.path` before running, so a script can `import` a plain `.py` helper module placed alongside it (see `script_helpers.py` / `11_import_example.gram.py`) without any custom `@include` directive
- Adds `custom_filter(name, namespace="Person")`: runs an existing Gramps sidebar custom filter by name and yields matching rows, printing a `Warning:` line to the Output tab (instead of raising) if the name doesn't match a filter — consistent with how `selected()`/`filtered()` already degrade to "nothing to show" rather than erroring (see `12_custom_filter_example.gram.py`)
- Adds `delete(obj)`: looks up the right `del_func` for the object's class the same way commits already look up `commit_func`, so a script can delete a record without knowing the underlying `remove_person`/`remove_family`/etc. call (see `13_delete_unused_repositories.gram.py`, which only deletes repositories nothing else refers to — safe to run, unlikely to touch anything in a typical tree)
- Fixes `active_event`, which returned a raw handle instead of a `DataDict2` like every other `active_*` constant (`self.get_active` → `self.get_active_data`)

## Tab completion (new)
- Adds Tab-triggered, live-filtering command completion to the script editor, built on [jedi](https://jedi.readthedocs.io/). Tab completes plain Python names, the DSL's own functions (`people()`, `custom_filter()`, `selected()`/`filtered()`, ...), and nested attribute chains on Gramps records — including `person.primary_name.first_name` through a user's own loop variable, and through list subscripts like `primary_name.surname_list[0].surname`
- `completion.py` wraps `jedi.Interpreter` for the actual lookups; `completion_popup.py` is a standalone `Gtk.Popover` controller (Tab opens/accepts, Up/Down navigate, Escape/click/focus-out dismiss, arbitrary typing live-refilters), kept independent of the Gramplet class so it's testable against a plain `Gtk.TextView`
- `stub_generator.py` derives static type stubs directly from Gramps' own `cls.get_schema()`, so jedi can infer a generator/loop-variable's row type (e.g. what `people()` yields) statically, without ever executing anything — a live template object would need to call `DataDict2`'s computed properties (`father`, `birth`, ...) to see what they return, which only yields empty results for blank/template data anyway
- `namespace_builder.py` supplies the remaining names that are safe to introspect as live objects (`today`, `counter`, `database`)
- `datadict2.py`: `DataDict2` gained a `__dir__` override so introspection (jedi's runtime fallback) sees dynamic dict keys like `primary_name`, not just its declared `@property` methods
- Also fixes the editor's `ScrolledWindow`/`TextView` having no wrap mode or explicit scroll policy, which let long lines widen the whole gramplet instead of scrolling within it
- Requires `jedi` (added to `GrampyScript.gpr.py`'s `requires_mod`), which ships with Gramps 6.1

## Test plan
- [x] `cd GrampyScript && python -m pytest tests/` — 110 tests pass (95 pure-logic + 15 GTK popup tests run under Xvfb against real widgets)
- [x] Verified `update_script_descriptions.py` against add/remove/rename scenarios, and that it round-trips the current `script_descriptions.py` byte-for-byte when nothing changed
- [x] Verified `xgettext --language=Python --keyword=_ ... --from-code=UTF-8` (the flags `make.py` uses) correctly extracts the new translatable strings from `script_descriptions.py`
- [x] Verified the new `sys.path` import logic and `decade()` helper standalone (no GTK required)
- [x] Verified `del_func` exists for all primary object types (Person, Family, Event, Place, Source, Citation, Repository, Note, Media) plus Tag
- [x] Instantiated the real `GrampyScript` Gramplet (real `DbState`, real SQLite-backed database, real `Gtk.TextView`) and drove actual Tab keypresses through `on_key_press`, confirming completion end-to-end outside of unit tests
- [ ] Manual: open Gramps with the addon loaded, confirm the Open-dialog preview, button placement, filename label, and unsaved-changes prompt look right; run the example scripts (including the new import/custom-filter/delete ones) against a real family tree

### DIFF
--- a/GrampyScript/GrampyScript.gpr.py
+++ b/GrampyScript/GrampyScript.gpr.py
@@ -32,4 +32,5 @@ register(
     gramplet_title=_("Gram.py Script"),
     help_url="Addon:GrampyScript",
     height=800,
+    requires_mod=["jedi"],
 )
--- a/GrampyScript/GrampyScript.py
+++ b/GrampyScript/GrampyScript.py
@@ -950,7 +950,7 @@ for person in people():
         if action == "set":
             if "_class" in data:
                 if self.db._get_table_func(data["_class"]):
-                    method = self.db._table(data["_class"], "commit_func")
+                    method = self.db._get_table_func(data["_class"], "commit_func")
                     method(data._object, self.TRANSACTION)
                 else:
                     raise Exception("%r class is not a PrimaryObject" % data["_class"])
--- a/GrampyScript/GrampyScript.py
+++ b/GrampyScript/GrampyScript.py
@@ -59,6 +59,8 @@ from gramps.gui.editors import (
 from datadict2 import DataDict2, NoneData, set_sa
 from script_descriptions import SCRIPT_DESCRIPTIONS
 from script_utils import get_columns, extract_header_comment, SCRIPTS_DIR
+from namespace_builder import build_namespace
+from completion_popup import CompletionController
 
 _ = glocale.translation.gettext
 
@@ -367,7 +369,9 @@ for person in people():
 
         self.editor = Gtk.ScrolledWindow()
         self.editor.set_shadow_type(Gtk.ShadowType.IN)
+        self.editor.set_policy(Gtk.PolicyType.AUTOMATIC, Gtk.PolicyType.AUTOMATIC)
         self.editor_textview = Gtk.TextView()
+        self.editor_textview.set_wrap_mode(Gtk.WrapMode.WORD_CHAR)
         self.editor.add(self.editor_textview)
         font_desc = self.editor_textview.get_pango_context().get_font_description()
         font_desc.set_family(
@@ -377,6 +381,7 @@ for person in people():
 
         self.editor_textview.connect("key-press-event", self.on_key_press)
         self.editor_textview.connect("button-press-event", self.on_textview_click)
+        self.editor_textview.connect("focus-out-event", self.on_editor_focus_out)
         key, mods = Gtk.accelerator_parse("<Alt>c")
         self.editor_textview.add_accelerator(
             "copy-clipboard", self.accel_group, key, mods, Gtk.AccelFlags.VISIBLE
@@ -404,6 +409,9 @@ for person in people():
             "comment", foreground="gray", style=Pango.Style.ITALIC
         )
         self.ebuf.connect("changed", self.on_buffer_changed)
+        self.completion = CompletionController(
+            self.editor_textview, get_namespace=lambda: build_namespace(self.dbstate.db)
+        )
 
         widget.pack_start(self.editor, True, True, 0)
 
@@ -663,6 +671,7 @@ for person in people():
 
     def on_buffer_changed(self, buffer):
         self.highlight_syntax()
+        self.completion.on_buffer_changed()
 
     def highlight_syntax(self):
         start_iter = self.ebuf.get_start_iter()
@@ -876,10 +885,18 @@ for person in people():
             return str(item)
 
     def on_textview_click(self, widget, event):
+        self.completion.close()
         if event.button == 1:  # Left mouse button
             widget.grab_focus()
 
+    def on_editor_focus_out(self, widget, event):
+        self.completion.close()
+        return False
+
     def on_key_press(self, textview, event):
+        if self.completion.on_key_press(event):
+            return True
+
         if event.keyval == Gdk.KEY_Tab:
             # buffer = textview.get_buffer()
             iter_ = self.ebuf.get_iter_at_mark(self.ebuf.get_insert())
--- a/GrampyScript/GrampyScript.py
+++ b/GrampyScript/GrampyScript.py
@@ -478,13 +478,17 @@ for person in people():
             border: 1px solid gray;
             padding: 1px;
         }
+        .bold-label {
+            font-weight: bold;
+            padding: 1px;
+        }
         """
         provider = Gtk.CssProvider()
         provider.load_from_data(css)
 
         self.filename_label = Gtk.Label()
         self.filename_label.set_xalign(0)
-        self.filename_label.get_style_context().add_class('bordered-label')
+        self.filename_label.get_style_context().add_class('bold-label')
         self.filename_label.get_style_context().add_provider(
             provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION
         )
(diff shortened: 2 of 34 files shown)


dsblank:
We can leave this for just 6.1... some encouragement to upgrade!

dsblank:
@GaryGriffin this one is ready.

GaryGriffin:
Running unittest, there are issues

- `python3 -m unittest GrampyScript.tests.test_stub_generator` : ModuleNotFoundError: No module named 'jedi'
- `python3 -m unittest GrampyScript.tests.test_completion` : ModuleNotFoundError: No module named 'jedi'
- `python3 -m unittest GrampyScript.tests.test_completion_popup` : ModuleNotFoundError: No module named 'jedi'
- `python3 -m unittest GrampyScript.tests.test_script_descriptions` : Passed tests but many Warning messages

> /Development/addons-source/GrampyScript/update_script_descriptions.py:106: ResourceWarning: unclosed file <_io.TextIOWrapper name='/Development/addons-source/GrampyScript/scripts/01_list_people.gram.py' mode='r' encoding='utf-8'>
>   source = open(path, encoding="utf-8").read()
> ResourceWarning: Enable tracemalloc to get the object allocation traceback

dsblank:
@GaryGriffin I fixed the warnings. The other errors are that the addon is designed to mini-install jedi. I don't know how your testing env does that.

GaryGriffin:
My testing env does not do mini-installs. Not sure that Github Actions would do that either.
@eduralph Will this PR work for testing in the proposed CI ?

dsblank:
(If you just want to run the tests, you can "pip install jedi")

eduralph:
> My testing env does not do mini-installs. Not sure that Github Actions would do that either. @eduralph Will this PR work for testing in the proposed CI ?

@GaryGriffin Yes, it'll work in the proposed CI  - it works slightly different that the way you work locally.

The CI doesn't depend on Gramps' runtime "mini-install." It reads requires_mod directly from each addon's .gpr.py and pip-installs the union before any test job runs (the "Install addon runtime deps (derived from requires_mod)" step, via addon_python_deps.py). Since this PR adds requires_mod=["jedi"], CI does pip install jedi up front, so import jedi resolves and test_completion / test_completion_popup / test_stub_generator pass. Because jedi is a pure-Python the Windows lane installs it fine too.

dsblank:
I fixed some more things, and is now ready again :smile_cat:

dsblank:
> The CI doesn't depend on Gramps' runtime "mini-install." 

@eduralph we probably should use it, to test that whatever package is listed that it can actually be installed that way. This is now in Gramps 6.1:

```python
from gramps.gen.utils.pypi import install_package
from gramps.gen.const import LIB_PATH

installed = install_package("jedi", LIB_PATH)
print(installed)   # list of package names actually installed
import jedi        # now importable
```

eduralph:
> > The CI doesn't depend on Gramps' runtime "mini-install."
> 
> @eduralph we probably should use it, to test that whatever package is listed that it can actually be installed that way. This is now in Gramps 6.1:
> 
> ```python
> from gramps.gen.utils.pypi import install_package
> from gramps.gen.const import LIB_PATH
> 
> installed = install_package("jedi", LIB_PATH)
> print(installed)   # list of package names actually installed
> import jedi        # now importable
> ```

@dsblank  - using install_package in the addon is the wrong approach in the current architecture IMHO. The original design was that the dependency loading is covered by Gramps itself based on the gpr.py hence the root package for unit testing which covers that "service" that Gramps normally provides for all addons when unit testing. 

You added the bespoke loading mechanism to Gramps as fallback if pip isn't available, but the loading responsibility should remain with Gramps, not the addon, ie not calling the install_package line. Gramps isn't the system under test, so the unit test shouldn't cover that part of code - I use the happy path for that which is having pip and load it in that fashion. 

This is a good example of where my GEMP Versioning of an API surface would highlight this: you'd be calling unsupported internals.

dsblank:
> The original design was that the dependency loading is covered by Gramps itself based on the gpr.py hence the root package for unit testing which covers that "service" that Gramps normally provides for all addons when unit testing.

Ok, then we probably should have some **integration** testing.

> You added the bespoke loading mechanism to Gramps as fallback if pip isn't available, 

Most uses of `requires_mod` will happen in packaged Gramps (AIO, flatpack, etc). IMHO this should be tested in integration tests somewhere.

eduralph:
> > The original design was that the dependency loading is covered by Gramps itself based on the gpr.py hence the root package for unit testing which covers that "service" that Gramps normally provides for all addons when unit testing.
> 
> Ok, then we probably should have some **integration** testing.
> 
> > You added the bespoke loading mechanism to Gramps as fallback if pip isn't available,
> 
> Most uses of `requires_mod` will happen in packaged Gramps (AIO, flatpack, etc). IMHO this should be tested in integration tests somewhere.

Yes, you are right - I was thinking of Ubuntu, but most will be on Windows, etc. You fortunately already built the E2E tests, leaving only the gap of real tests with addons. 

I was struggling with the framing, sorry for that. There is already a plugin loading test as part of the 820 CI fix planned but doesn't include your fallback loading. I could extract the testing alltogether and include the additional code you've written to cover exactly that.

dsblank:
> I was thinking of Ubuntu

I wish that was all we had to worry about :smile_cat: Proud Linux user since 1992, now using Ubuntu.

PR #980, opened 2026-07-08
test: shared test harness + plugin-registration tests

eduralph:
Carved out of #820 so the addon-facing tests can land **before** the CI pipeline itself. These four items only need Gramps — they have zero dependency on the `.github/scripts/` CI code — so they are reviewable and mergeable on their own. Splitting them out also shrinks #820 toward the "one concern per PR" shape the maintainers asked for.

## What's in this PR

- **`tests/gramps_test_env.py`** — `GrampsTestCase` / `GrampsDbTestCase` base classes (stdlib `unittest`, mirroring upstream Gramps' own test style). Boots the plugin manager once per process and hands out fresh in-memory DBs.
- **`tests/test_plugin_registration.py`** — registers every addon through the real Gramps `PluginRegister` / `BasePluginManager`, verifies `gramps_target_version`, valid id/name/version, and smoke-tests import/export entry functions.
- **`tests/test_plugin_load_gate.py`** — regression test that drives the plugin-load *gating* path (a hard, non-dependency load failure must fail the test rather than pass silently).
- **`make.py`** — a new `test` command that runs the repo-root `tests/` suite via `unittest` discovery (the exact discovery the CI job uses), gated on `GRAMPSPATH` like the other commands.

`tests/__init__.py` (the GI version pin) is already on `maintenance/gramps60`, so it isn't part of this diff.

## Running the tests

```bash
GRAMPSPATH=/path/to/gramps python3 make.py gramps60 test
```

`GRAMPSPATH` must point at a Gramps **6.0** checkout so the addon target versions match. `test_plugin_registration` boots the full plugin registry and needs the GTK typelibs present; `test_plugin_load_gate` is import-light and runs anywhere.

## Relationship to #820

The three remaining root tests in #820 — `test_addon_system_deps.py`, `test_requires_mod_dedup.py`, `test_run_addon_tests_paths.py` — stay there because they import modules out of `.github/scripts/`, which doesn't exist until the CI PR merges. Once this PR lands, #820 rebases and sheds these files from its diff, leaving it strictly CI infrastructure.

### DIFF
--- a/make.py
+++ b/make.py
@@ -54,6 +54,21 @@ Examples:
        Runs the repo-root `tests/` suite (unittest discovery for
        `test_*.py`). GRAMPSPATH must point at a Gramps checkout so the
        addon plugins are importable. Exits non-zero on any failure.
+
+       Default mode is a headless developer SMOKE run, not a full gate:
+         - only listed addons (include_in_listing=True) are load/version/
+           dependency checked; unlisted ones are covered only by the
+           listing-manifest test;
+         - failures that are really about the environment — no display, a
+           missing GI/typelib, a native GTK abort — are advisory, not fatal;
+         - tests needing extra tools skip (e.g. i18n extraction without
+           `xgettext`, GUI loads without a display).
+       For a full gate, run under a complete runtime (Xvfb + gettext + all
+       GI typelibs) with GRAMPS_ADDON_TEST_STRICT=1: unmet dependencies,
+       display/GTK failures and unverified addon dependencies then become
+       hard failures. (Slow registry-scan timeouts stay advisory — they are
+       a performance signal, not a correctness one; a genuine import hang is
+       always a hard failure.)
 """
 import shutil
 import glob
@@ -1079,6 +1094,12 @@ elif command == "extract-po":
         extract_po(addon)
 
 elif command == "test":
+    import logging
+
+    # Quiet the "no compiled locale" warning Gramps logs at import: this
+    # command imports Gramps (via check_gramps_path) before the test package's
+    # own suppression runs, and translations are irrelevant to the tests.
+    logging.getLogger(".gramps.gen.utils.grampslocale").setLevel(logging.ERROR)
     check_gramps_path(command)
     import unittest
 
--- a/make.py
+++ b/make.py
@@ -49,6 +49,11 @@ Examples:
    python3 make.py gramps60 extract-po
        Extracts strings from the aggregated `po/{lang}.po` files into the
        `{lang}-local.po` files for each addon.
+
+   GRAMPSPATH=path python3 make.py gramps60 test
+       Runs the repo-root `tests/` suite (unittest discovery for
+       `test_*.py`). GRAMPSPATH must point at a Gramps checkout so the
+       addon plugins are importable. Exits non-zero on any failure.
 """
 import shutil
 import glob
@@ -1072,6 +1077,17 @@ elif command == "extract-po":
     ]:
         print(addon)
         extract_po(addon)
+
+elif command == "test":
+    check_gramps_path(command)
+    import unittest
+
+    suite = unittest.TestLoader().discover(
+        "tests", pattern="test_*.py", top_level_dir="."
+    )
+    result = unittest.TextTestRunner(verbosity=2).run(suite)
+    if not result.wasSuccessful():
+        exit(1)
 
 else:
     raise AttributeError("unknown command")
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -6,7 +6,14 @@ the GTK 3 / GDK 3 stack a real Gramps GUI session uses. A test that imports a
 ``gramps.gui.*`` module directly never runs the launcher's own
 ``require_version``; without this the GI stack can resolve to GTK 4 on a host
 where that is the default — emitting ``PyGIWarning`` and risking the wrong stack.
+
+This module also runs before Gramps is imported, so it is where we silence
+log/warning spam that is expected when testing *uncompiled, source-tree* addons
+(no ``locale/*.mo``) and would otherwise bury the test results.
 """
+
+import logging
+import warnings
 
 try:
     import gi
@@ -17,3 +24,17 @@ except Exception:
     # No PyGObject, or the 3.0 typelibs are unavailable — leave the environment
     # untouched; this only fixes the version when it can.
     pass
+
+# Gramps warns once per addon that lacks a compiled locale dir (source-tree
+# addons ship ``po/*.po`` only), plus a startup warning about the main locale
+# dir — expected here, not a failure. Raising the logger to ERROR is robust:
+# logging levels are not reset by the test runners, so this holds on every
+# entrypoint.
+logging.getLogger(".gramps.gen.utils.grampslocale").setLevel(logging.ERROR)
+# PyGObject deprecation shout from gramps.gui.glade on import — not our concern.
+# NOTE: this warning *filter* is honoured by ``make.py test`` (TextTestRunner),
+# but a bare ``python -m unittest`` resets warning filters mid-run and may still
+# print it once. The documented entrypoint (make.py) stays clean; suppressing it
+# everywhere would mean importing gramps.gui.glade eagerly here (pulls in Gtk on
+# every run), which isn't worth it for one cosmetic line.
+warnings.filterwarnings("ignore", message=r".*shouldn't use __slots__.*")
--- a/tests/gramps_test_env.py
+++ b/tests/gramps_test_env.py
@@ -104,23 +104,26 @@ _GUI_DIALOG_NAMES = (
     "RunDatabaseRepair",
 )
 
-def neutralize_gui_dialogs() -> None:
-    """Replace ``gramps.gui.dialog`` dialogs with no-ops for the test process.
+#: Written to stdout by a neutralised dialog, followed by its class name.
+#: Distinctive enough not to collide with an addon's own output. The isolated
+#: load subprocess in ``test_plugin_registration`` greps its captured stdout for
+#: this: neutralising the dialog stops the load from hanging, but the addon
+#: still has a bug, and the suite must say so rather than swallow it.
+DIALOG_MARKER = "__GRAMPS_TEST_DIALOG_AT_IMPORT__"
 
-    Some addons pop a *blocking* modal (``ErrorDialog(...).run()``) at import or
-    load-on-reg time when an optional dependency is missing — e.g.
-    ``lxmlGramplet`` on a host without ``python3-lxml``. Loading such an addon
-    in a test would otherwise hang on the modal (under a display) or scatter
-    dialogs on the developer's screen. Safe to call more than once.
-    """
-    try:
-        import gramps.gui.dialog as gd
-    except Exception:
-        return
+#: Names of the dialog classes constructed since the stub was installed, for
+#: callers that neutralise and load in the same process.
+dialogs_raised: list[str] = []
+
+
+def _make_no_dialog(dialog_name: str) -> type:
+    """Build a no-op stand-in for ``gramps.gui.dialog.<dialog_name>``."""
 
     class _NoDialog:
         def __init__(self, *args: Any, **kwargs: Any) -> None:
             self.response = 0
+            dialogs_raised.append(dialog_name)
+            print(f"{DIALOG_MARKER} {dialog_name}", flush=True)
 
         def run(self, *args: Any, **kwargs: Any) -> int:
             return 0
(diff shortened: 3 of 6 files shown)


GaryGriffin:
Tried `make.py gramps60 test `on my Mac. Failed but not sure what is the issue to fix. I think it is lxml gramplet, but not sure. That gramplet does not have a requires in the gpr, so why time out? Was an env variable undefined? 

```
All listed addons must target the Gramps series they're running against. ... ok

ERROR: test_load_all_addon_modules (tests.test_plugin_registration.TestPluginLoading.test_load_all_addon_modules)
Load every addon plugin; gate on non-dependency load failures.
  File "/Volumes/Storage/Development/addons-source/tests/test_plugin_registration.py", line 187, in test_load_all_addon_modules
    result = subprocess.run(
        [
    ...<17 lines>...
        env={**os.environ, "PYTHONPATH": ADDONS_ROOT},
    )
  File "/Users/GrampsTest/gtk/inst/lib/python3.13/subprocess.py", line 558, in run
    stdout, stderr = process.communicate(input, timeout=timeout)
                     ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/GrampsTest/gtk/inst/lib/python3.13/subprocess.py", line 1221, in communicate
    stdout, stderr = self._communicate(input, endtime, timeout)
                     ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/GrampsTest/gtk/inst/lib/python3.13/subprocess.py", line 2131, in _communicate
    self._check_timeout(endtime, orig_timeout, stdout, stderr)
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/GrampsTest/gtk/inst/lib/python3.13/subprocess.py", line 1265, in _check_timeout
    raise TimeoutExpired(
    ...<2 lines>...
            stderr=b''.join(stderr_seq) if stderr_seq else None)
subprocess.TimeoutExpired: Command '['/Users/GrampsTest/gtk/inst/bin/python3', '-c', "import sys; sys.path.insert(0, '/Volumes/Storage/Development/addons-source');from gramps.gen.plug import BasePluginManager;from gramps.gen.const import PLUGINS_DIR;pmgr = BasePluginManager.get_instance();pmgr.reg_plugins(PLUGINS_DIR, None, None);pmgr.reg_plugins('/Volumes/Storage/Development/addons-source', None, None);from gramps.gen.plug import PluginRegister;preg = PluginRegister.get_instance();pdata = preg.get_plugin('lxml Gramplet');mod = pmgr.load_plugin(pdata);sys.exit(0 if mod else 1)"]' timed out after 30 seconds

Ran 16 tests in 129.029s

FAILED (errors=1)

```

GaryGriffin:
Reran and the output was much more understandable. Great progress - this was sorta what I was looking for when we started - a way to test the addons from my local machine. 



Something weird on the Gexiv2. I see the following but only one generated error. The former passes, the latter fails. Is there a syntax issue if there is a comma string in the versions? 
```
EditExifMetadata/editexifmetadata.gpr.py:    requires_gi=[("GExiv2", "0.10")],
PhotoTaggingGramplet/PhotoTaggingGramplet.gpr.py:    requires_gi=[("GExiv2", "0.10,0.12,0.14,0.16")],
```

My current error log is below. Do you get the same as I get running locally? I want to set my expectations for Mac usability. Did this actually load the requires_mod/requires_gi packages? 

```

Skipped for unmet dependencies (2):
  graphview (missing: gi:GooCanvas-2.0,3.0)
  Photo Tagging (missing: gi:GExiv2-0.10,0.12,0.14,0.16)
Failed for environmental reasons — signal crash or GTK/display init, likely need a display server (1):
  Collections Clipboard Gramplet (gi.repository.GLib.GError: gtk-icon-theme-error-quark: Icon 'stock_link' not present in theme Adwaita (0))
FAIL
test_addons_discovered (tests.test_plugin_registration.TestPluginRegistration.test_addons_discovered)
At least some plugins should be registered. ... ok
test_all_plugins_have_valid_metadata (tests.test_plugin_registration.TestPluginRegistration.test_all_plugins_have_valid_metadata)
Every registered plugin must have id, name, and version. ... ok
test_target_version_matches_gramps_install (tests.test_plugin_registration.TestPluginRegistration.test_target_version_matches_gramps_install)
All listed addons must target the Gramps series they're running against. ... ok

FAIL: test_load_all_addon_modules (tests.test_plugin_registration.TestPluginLoading.test_load_all_addon_modules)
Load every addon plugin; gate on non-dependency load failures.
  File "/Volumes/Storage/Development/addons-source/tests/test_plugin_registration.py", line 347, in test_load_all_addon_modules
    self.fail("\n".join(lines))
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^
AssertionError: 1 of 174 addon(s) failed to load. These are real load failures and must be fixed:

  ✗ lxml Gramplet (hung during import, >120s)

Not failing this test (advisory only; full lists in the log above):
  1 could not initialise GTK/a display — run under Xvfb to load these; not an addon bug
  2 skipped for an unmet optional dependency (requires_mod/gi/exe)

Ran 24 tests in 224.865s

FAILED (failures=1)
```

eduralph:
@GaryGriffin - it was abit more work than I originally thought but it should be clean now. There were some tricky items to differentiate between loading hangs and other things, but it should all be good. Let me know what you think about it.

eduralph:
@dsblank - I took your idea and ran with it. The PR 820 already had some tests in place, I just needed to wire up your addition and change it a little bit.

GaryGriffin:
Confused. I ran `make.py gramps60 tests` (without merging PR981). I expected to still choke on lxml issue but it did not. Is this now hidden? 

Also, still get warning on Photo Tagging but not EditExIfMetadata, both checking Gexiv2 (but different strings). Seems strange there are Warnings for exactly the ones that have a comma-separated version in the requires_gi but not for the ones without.

eduralph:
> Confused. I ran `make.py gramps60 tests` (without merging PR981). I expected to still choke on lxml issue but it did not. Is this now hidden?
> 
> Also, still get warning on Photo Tagging but not EditExIfMetadata, both checking Gexiv2 (but different strings). Seems strange there are Warnings for exactly the ones that have a comma-separated version in the requires_gi but not for the ones without.

Good points ... I buried the GUI issues (including lxml) and that is probably not the right approach to take. It protects the test from choking but also starts hiding issues. It now records the dialog, and the load test reports it as a hard failure in both modes, naming the addon and the dialog class. So lxml goes back to red — with an accurate message this time, "raised ErrorDialog at import" rather than "hung" — stays red until #981 merges, and any future addon that does the same gets caught rather than silently neutralised. 

GExiv2 — that's my bug, Gramps supports a comma-separated version list in requires_gi. It splits, and accepts if any listed version resolves. My _check_dependencies() in tests/test_plugin_registration.py didn't. 

To your earlier question — "did this actually load the requires_mod/requires_gi packages?" — no. The dependency check only probes (find_spec / gi.require_version / PATH lookup) to decide whether it's fair to attempt the load; the actual import happens in the isolated subprocess afterwards. So a missing typelib is a skip, never a failure, outside strict mode.

GaryGriffin:
Not cherry picking to gramps61 unless you explicitly tell me to check it.

GaryGriffin:
One suggestion - running this creates a file debug.log . If at the end of the run, it is zero length, suggest you delete it. Only leave it if it has messages.

GaryGriffin:
FYI - I ran this PR (`make.py gramps61 test`) against gramps61 build. The following two errors occurred.

- GrampsAssistant has a dependency on the addon Grampy Script. Both exist in the 61 branch. What should change to make this pass?  Adding @dsblank to confirm since these are both new addons. And both work correctly.
- Gexiv2 - It is made as a dylib as part of the Mac bundle for 6.1 . `lib/libgexiv2-<version>.dylib` .  Adding @jralls in case I am mis-speaking. This is different and why it is no longer a requires_gi in the GPR.

```

FAIL: test_declared_dependencies_resolve_to_registered_plugins (tests.test_addon_dependency_loading.TestAddonDependencyLoading.test_declared_dependencies_resolve_to_registered_plugins)
Every ``depends_on`` id must name a plugin Gramps actually registered.
  File "/Volumes/Storage/Development/addons-source/tests/test_addon_dependency_loading.py", line 201, in test_declared_dependencies_resolve_to_registered_plugins
    self.fail(
    ~~~~~~~~~^
        "Addons declaring depends_on ids that are not registered "
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        "(the Gramps loader cannot satisfy these):\n  "
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        + "\n  ".join(sorted(unresolved))
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
AssertionError: Addons declaring depends_on ids that are not registered (the Gramps loader cannot satisfy these):
  grampsassistant -> Grampy Script

FAIL: test_load_all_addon_modules (tests.test_plugin_registration.TestPluginLoading.test_load_all_addon_modules)
Load every addon plugin; gate on non-dependency load failures.
  File "/Volumes/Storage/Development/addons-source/tests/test_plugin_registration.py", line 435, in test_load_all_addon_modules
    self.fail("\n".join(lines))
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^
AssertionError: 2 of 173 addon(s) failed to load cleanly. These are real defects and must be fixed:

  ✗ Edit Image Exif Metadata (Plugin error (from 'editexifmetadata'): GExiv2 is not installed)
  ✗ Photo Tagging (Plugin error (from 'PhotoTaggingGramplet'): GExiv2 is not installed)

Not failing this test (advisory only; full lists in the log above):
  1 could not initialise GTK/a display — run under Xvfb to load these; not an addon bug

```

dsblank:
Found a gap in `neutralize_gui_dialogs()` while triaging a warning-dialog report on #993 (WordClouds).

`_GUI_DIALOG_NAMES` covers `ErrorDialog`, `WarningDialog`, `OkDialog`, `InfoDialog`, `QuestionDialog`, `QuestionDialog2`, `DBErrorDialog`, `RunDatabaseRepair` — but not `MessageHideDialog`. Core Gramps' own `gramps/plugins/gramplet/gramplet.gpr.py` uses `MessageHideDialog` (not `WarningDialog`) for its "GExiv2 module not loaded" check, run as module-level code whenever `has_display()` is `True` and GExiv2 isn't importable:

```python
if uistate:
    MessageHideDialog(title, message, "interface.ignore-gexiv2", parent=uistate.window)
else:
    MessageHideDialog(title, message, "interface.ignore-gexiv2")
```

`MessageHideDialog.__init__` calls `self.top.run()` — a real blocking modal (`gramps/gui/dialog.py`).

Since `get_plugin_manager_and_registry()` calls `pmgr.reg_plugins(PLUGINS_DIR, None, None)` — `PLUGINS_DIR` being core's own `plugins/` tree, not just `ADDONS_ROOT` — this modal is in-scope for every `test_plugin_registration.py` run / `make.py test` invocation, and it isn't neutralized. On a host with a live display and no GExiv2 available, this would hang rather than fail cleanly, until whatever wall-clock timeout wraps the process kills it.

Not a problem inside the CI container itself, since the Dockerfile bundles `gir1.2-gexiv2-0.10` — the `available` branch in `gramplet.gpr.py` short-circuits before reaching the dialog. But it'll hang for anyone running the harness locally (`make.py test`, or importing `gramps_test_env` directly) on a machine without GExiv2 and a real display — e.g. a Mac dev environment.

Suggest adding `"MessageHideDialog"` to `_GUI_DIALOG_NAMES`. Happy to send a one-line PR for it if that's easier than you doing it.

dsblank:
Following up with the concrete one-liner: in `tests/gramps_test_env.py`, add `"MessageHideDialog"` to `_GUI_DIALOG_NAMES` (line 104):

```python
_GUI_DIALOG_NAMES = (
    "ErrorDialog",
    "WarningDialog",
    "OkDialog",
    "InfoDialog",
    "QuestionDialog",
    "QuestionDialog2",
    "DBErrorDialog",
    "RunDatabaseRepair",
    "MessageHideDialog",
)
```

`_make_no_dialog` already builds a generic stand-in regardless of the target dialog's constructor signature, so nothing else needs to change — `neutralize_gui_dialogs()` just iterates the tuple and swaps each matching attribute on `gramps.gui.dialog`.

PR #981, opened 2026-07-09
lxml: don't show a blocking dialog at import when lxml/gzip is missing

eduralph:
## Problem

`lxml/lxmlGramplet.py` calls `ErrorDialog(...)` at **module import time** when `gzip` or `lxml` is unavailable:

```python
try:
    from lxml import etree, objectify
    LXML_OK = True
    ...
except ImportError:
    LXML_OK = False
    ErrorDialog(_('Missing python3 lxml'), _('Please, try to install "python3 lxml" package.'))
```

Gramps' `ErrorDialog.__init__` calls `.run()`, so this is a **blocking modal** raised before any GUI action. Consequences:

- It stalls plugin loading until someone clicks Close.
- It assumes a GUI is running. Imported without a display (CLI, or automated tooling), it errors or aborts — on a host missing `python3-lxml` the module hangs or scatters dialogs.

The gramplet already degrades gracefully — it falls back to `xml.etree` when lxml is absent (`LXML_OK = False`) — so the missing dependency is **not fatal**.

## Fix

- At import, set `GZIP_OK` / `LXML_OK` silently and `LOG` the condition — no dialog.
- Show the notice in `init()`, when the gramplet is actually opened and the GUI is up, so a user who opens the gramplet still gets the helpful message.
- Bump `lxmlGramplet` `1.2.3` → `1.2.4`.

`etreeGramplet.py` is unaffected (its `ErrorDialog` calls are already inside user-triggered methods, and it does not import lxml at module top).

## Verification

Importing the module with `lxml` unavailable now returns cleanly (`LXML_OK = False`, one log line) instead of raising a blocking modal — verified headless under Xvfb.

Surfaced while adding CI test infrastructure (#980), where loading this addon on a runner without `python3-lxml` hung the plugin-load test; that test harness works around it, but the addon behaviour is the real bug and is fixed here.

### DIFF
--- a/lxml/lxmlGramplet.py
+++ b/lxml/lxmlGramplet.py
@@ -69,7 +69,6 @@ try:
     GZIP_OK = True
 except ImportError:
     GZIP_OK = False
-    ErrorDialog(_('Where is gzip?'), _('"gzip" is missing'))
     LOG.error('No gzip')
 
 #-------------------------------------------------------------------------
@@ -91,8 +90,7 @@ try:
     LIBXSLT_VERSION = etree.LIBXSLT_VERSION
 except ImportError:
     LXML_OK = False
-    ErrorDialog(_('Missing python3 lxml'), _('Please, try to install "python3 lxml" package.'))
-    LOG.debug('No lxml')
+    LOG.warning('No lxml; XPATH/XSLT features are unavailable')
 
 #-------------------------------------------------------------------------
 #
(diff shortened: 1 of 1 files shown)


GaryGriffin:
No need to increment version. That is done automatically at publish time.

eduralph:
> No need to increment version. That is done automatically at publish time.

Sorry, that's what happens if you don't use the proper harness that loads those kind of rules into context. Admittedly, it was late - I should have just waited until the next day.

GaryGriffin:
This PR deletes the GPR file. You just needed to delete the change to the GPR.

eduralph:
> This PR deletes the GPR file. You just needed to delete the change to the GPR.

github's ****** interface ... it said "delete file" in the change set. I just wanted to remove it from the changeset, not make it delete the whole file. Doesn't make sense, but ok ... 

Anyways, it's correct now

PR #985, opened 2026-07-15
Themes: fix TypeError on preferences_activate(initial_panel=...)

dsblank:
*Without this fix, the addon crashes on loading preferences in Gramps 6.1*

## Summary
- `GrampsPreferences.__init__` in Gramps core gained an `initial_panel` keyword so callers (e.g. `ViewManager.preferences_activate`) can open Preferences directly on a specific panel.
- The Themes addon's `themes_load.py` monkey-patches `GrampsPreferences.__init__ = MyPrefs.__init__`, but `MyPrefs.__init__` didn't accept the new keyword, causing:
  ```
  TypeError: MyPrefs.__init__() got an unexpected keyword argument 'initial_panel'
  ```
- `MyPrefs.__init__` now accepts `initial_panel=None` and calls `self.select_panel(initial_panel)` after setup, mirroring core's behavior (guarded with `hasattr` for compatibility with older Gramps versions lacking `select_panel`).

## Test plan
- [x] Added `Themes/tests/test_initial_panel.py`: confirms `MyPrefs.__init__` accepts `initial_panel`, calls `select_panel('colors')` when given, and does not call it when omitted.
- [x] Verified the new tests fail with the original `TypeError`/missing-parameter error against the pre-fix code, and pass after the fix.
- [x] `python3 -m unittest tests.test_initial_panel -v` passes.


GaryGriffin:
I thought [950](https://github.com/gramps-project/addons-source/pull/950) addressed the pinning of Gdk and Gtk to v3 for all tests. Why is this needed here too?

dsblank:
> I thought https://github.com/gramps-project/addons-source/pull/950 addressed the pinning of Gdk and Gtk to v3 for all tests. Why is this needed here too?

If there is a recommendation on what to do, it should be in an AGENTS.md file. Otherwise how will developers know what to do?

GaryGriffin:
@eduralph suggestions for messaging?

eduralph:
Gary, you are right, this is exactly the reason we introduced this in #950.

It covers Gdk and Gtk pinning for unit testing in the addons, in run-time it's Gramps that covers that. 

@dsblank - I had built a rule set that we can include into AGENTS.md that would also include that; you can find it [here](https://github.com/eduralph/gramps-testbed-v2/blob/main/wiki/pages/06%20-%20Addon%20development/16-guidelines.md) for now. 

I also published these to the [Wiki](https://gramps-project.org/wiki/index.php/User:Eduralph/Sandbox/Gramps_6.0_Wiki_Manual_-_Addon_Development_-_Rules) as test (it still needs a little bit of touching up) which includes it as HTML and a bigger overhaul of how to develop Addons. 

Here's also [more](https://github.com/eduralph/gramps-testbed-v2/blob/main/wiki/pages/06%20-%20Addon%20development/08-testing.md#why-tests__init__py-exists) on the subject. I'm sorry it's still in my repo.

I'll focus on getting the content out there and into the repository this weekend. Thoughts and feedback welcome of course.

dsblank:
@GaryGriffin fixed. This is a blocker for people using Gramps 6.1 with this addon. Prevents preferences from being opened.

PR #986, opened 2026-07-15
Form: drop column-size sum-to-100 warning

dsblank:
## Summary
- `<size>` on `<column>` is never used to compute layout in the Form gramplet itself (entry widths are sized from their text content, not the XML value)
- PDFForms consumes it purely as a relative weight (`_col_widths()`), which works correctly regardless of what the values sum to
- The sum-to-100 warning added for bug 11010 was firing on 78 shipped forms with no actual functional problem behind it, so it's removed along with its tests

## Test plan
- [x] `python3 -m unittest tests.test_form_validator tests.test_integration_form -v` — 40 tests pass


GaryGriffin:
This reverts [822](https://github.com/gramps-project/addons-source/pull/822) . 
That PR was also cherry picked to gramps60

Adding @eduralph

eduralph:
@dsblank - can you tell me a bit more about this? It was based on a confirmed bug (and Mantis isn't responding right now), so I'm trying to understand why the whole thing was removed.

dsblank:
Basically it wasn't used at all in Forms, and only as a percentage in PDFForms

Nick-Hall:
> Basically it wasn't used at all in Forms, and only as a percentage in PDFForms

It is used for the Census reports.  I'm not sure how many people still use them.

eduralph:
Ok, that's more a judgement call in terms of "should we have done that at all" instead of anything fundamentally wrong with the fix itself. As I'm not 100% familiar with  all features. Thanks

dsblank:
This PR does not affect Census reports/forms.

What this PR removes: `get_form_warnings()` (Form/form_validator.py), a function that only logged a WARNING to the log file when a section's <column> `size` values didn't sum to 100. It was purely diagnostic (Gramps bug 11010).

What it doesn't touch: the actual rendering logic in Form/form.py (lines 261–277), which reads each column's `size` value and uses it as-is when building the form — regardless of whether the sizes across a section sum to 100. That code is unchanged by this PR.

I also checked Form/CensusCheckQuickview.py and CensusCheckUpQuickview.py (the Census-specific tooling in this addon) — neither reads `size` at all; they only match on `form_type == "Census"`.

So: Census forms will render exactly as before. The only behavior change is that a form author authoring a Census form with column sizes not summing to 100 will no longer see a warning in the log — the form still loads and renders with whatever sizes were specified, same as before this PR.

PR #987, opened 2026-07-17
PersonRelationshipFilter

dsblank:
A new gramplet, **Relationship Filter** (previously called **Flat Filter**), that drops into the Person view's Gramplet bar as a sidebar-style filter. Unlike the built-in Person filter (which only matches on the active person's own attributes), this one searches by the *names of relatives*:

- the person's own name
- father's / mother's name
- spouse's name
- up to two named siblings
- up to two named children
- a "probably alive on date" filter

Each name field is split into given-name and surname boxes searched independently (substring or regex), and results combine into a single `GenericFilter` applied to the Person view.

## Motivation

The stock Person sidebar filter can't answer "find people whose mother is named X" or "find the children of someone named Y" — you have to know the target person's own name or ID. In a large tree, it's often the *relative's* name you remember, not the person you're actually looking for. This gramplet lets you search on that relationship directly instead of navigating the tree by hand.

## History and bugs fixed here

This is a new addon for addons-source, based on earlier work by Paul Womack (BugBear, 2024). While bringing it up to date for Gramps 6.1, I found and fixed two correctness bugs in that original code:

* **`IsSiblingofNamedSibling` crashed with `HandleError`** whenever applied to a person with no recorded parents (any root ancestor, immigrant, etc.). `Person.get_main_parents_family_handle()` returns `None` when there's no parent family, but the code passed that straight into `get_family_from_handle()`, which raises rather than returning `None` on a `None` handle. Fixed with a guard, matching the pattern used elsewhere in this file and in core (e.g. `persondetails.py`).
* **`RegExpPersonal`/`RegExpFamily` searched mismatched `Name` fields.** The personal (given-name) field list listed `title` twice instead of including the call name, so a personal-name search never matched a person's call name. The family/surname field list included `title` and `call` name — both personal-name fields — so a surname search could false-positive on an unrelated title or call name. Split the field lists to match the Given/Family grouping in the Name Editor.

Also:

* Renamed a variable shadowing the `all` builtin.
* Reformatted with `black`; added a type annotation so `mypy` is clean.
* Added `tests/test_person_relationship_filter_rules.py` — builds a small family tree directly in an in-memory database and exercises every rule (`RegExpPersonal`/`RegExpFamily` field matching, `HasNamedFather/Mother/Child/Spouse`, `IsSiblingofNamedSibling`, `HasName`), including both regressions above. `tests/__init__.py` is left empty, matching the convention used by the rest of the repo's per-addon test suites (only `RepositoriesReport` pins GTK/GDK at the package level; every other addon's `tests/__init__.py` is empty and each test module self-guards, which this one does too).

## Test plan

- [x] `python3 -m pytest PersonRelationshipFilter/tests/ -v` — 19 passed
- [x] Verified both regression tests fail against the pre-fix code (`HandleError` for the sibling crash, `AssertionError` for the call-name mismatch) and pass again with the fix
- [x] `black --check PersonRelationshipFilter/` — clean
- [x] `mypy --ignore-missing-imports PersonRelationshipFilter/PersonRelationshipFilter.py PersonRelationshipFilter/tests/` — clean

### DIFF
--- a/PersonRelationshipFilter/tests/test_person_relationship_filter_rules.py
+++ b/PersonRelationshipFilter/tests/test_person_relationship_filter_rules.py
@@ -1,0 +1,326 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2026  Doug Blank <doug.blank@gmail.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+"""
+Unit tests for the relationship-matching filter rules defined in
+``PersonRelationshipFilter.py``.
+
+The tests build a small family tree directly in an in-memory database
+(no dependency on the shared ``example.gramps`` fixture) so each
+rule's relationship traversal and name-field matching can be checked
+precisely. Two regressions are covered explicitly:
+
+* ``IsSiblingofNamedSibling`` used to raise ``HandleError`` when
+  applied to a person with no recorded parents, because
+  ``get_family_from_handle(None)`` raises rather than returning
+  ``None``.
+* ``RegExpPersonal``/``RegExpFamily`` searched mismatched ``Name``
+  fields (``title`` was listed twice in the personal field list, and
+  ``call`` name was searched by the family/surname rule instead), so
+  personal search never matched a person's call name and surname
+  search could false-positive on an unrelated title or call name.
+"""
+
+# ------------------------
+# Python modules
+# ------------------------
+import os
+import sys
+import unittest
+
+# The addon imports Gtk at module load — skip cleanly if gi/Gtk are not
+# available, mirroring what other addons' test suites do.
+try:
+    import gi
+
+    gi.require_version("Gtk", "3.0")
+    gi.require_version("Gdk", "3.0")
+except (ImportError, ValueError, AttributeError) as err:
+    raise unittest.SkipTest("GTK 3.0 / PyGObject not available: %s" % err)
+
+# ------------------------
+# Gramps modules
+# ------------------------
+# The addon directory goes on sys.path so ``import PersonRelationshipFilter``
+# resolves the flat ``PersonRelationshipFilter.py`` module directly (there is
+# no wrapping package — the file lives right in the addon directory).
+ADDON_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+if ADDON_DIR not in sys.path:
+    sys.path.insert(0, ADDON_DIR)
+
+try:
+    import gramps
+except ImportError as err:
+    raise unittest.SkipTest("gramps package not available: %s" % err)
+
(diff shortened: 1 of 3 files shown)


GaryGriffin:
Don't need the pinning of Gtk and Gdk to 3.0 in the test.

PersonRelationshipFilter/tests/test_person_relationship_filter_rules.py
```diff
@@ -0,0 +1,326 @@
...
+# The addon imports Gtk at module load — skip cleanly if gi/Gtk are not
+# available, mirroring what other addons' test suites do.
+try:
+    import gi
+
+    gi.require_version("Gtk", "3.0")
+    gi.require_version("Gdk", "3.0")
+except (ImportError, ValueError, AttributeError) as err:
+    raise unittest.SkipTest("GTK 3.0 / PyGObject not available: %s" % err)
+
```
dsblank:
```suggestion
```

GaryGriffin:
Running unittest failed

```shell
$ python3 -m unittest PersonRelationshipFilter.tests.test_person_relationship_filter_rules

  File "/Volumes/Storage/Development/addons-source/PersonRelationshipFilter/tests/test_person_relationship_filter_rules.py", line 71, in <module>
    from PersonRelationshipFilter import (
    ...<8 lines>...
    )
ImportError: cannot import name 'HasName' from 'PersonRelationshipFilter' (unknown location)

During handling of the above exception, another exception occurred:

  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/GrampsTest61/gtk/inst/lib/python3.14/unittest/__main__.py", line 18, in <module>
    main(module=None)
    ~~~~^^^^^^^^^^^^^
  File "/Users/GrampsTest61/gtk/inst/lib/python3.14/unittest/main.py", line 103, in __init__
    self.parseArgs(argv)
    ~~~~~~~~~~~~~~^^^^^^
  File "/Users/GrampsTest61/gtk/inst/lib/python3.14/unittest/main.py", line 142, in parseArgs
    self.createTests()
    ~~~~~~~~~~~~~~~~^^
  File "/Users/GrampsTest61/gtk/inst/lib/python3.14/unittest/main.py", line 153, in createTests
    self.test = self.testLoader.loadTestsFromNames(self.testNames,
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
                                                   self.module)
                                                   ^^^^^^^^^^^^
  File "/Users/GrampsTest61/gtk/inst/lib/python3.14/unittest/loader.py", line 207, in loadTestsFromNames
    suites = [self.loadTestsFromName(name, module) for name in names]
              ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/Users/GrampsTest61/gtk/inst/lib/python3.14/unittest/loader.py", line 137, in loadTestsFromName
  File "/Volumes/Storage/Development/addons-source/PersonRelationshipFilter/tests/test_person_relationship_filter_rules.py", line 82, in <module>
    raise unittest.SkipTest("PersonRelationshipFilter module unavailable: %s" % err)
unittest.case.SkipTest: PersonRelationshipFilter module unavailable: cannot import name 'HasName' from 'PersonRelationshipFilter' (unknown location)
```

dsblank:
Fixed — this was the same namespace-package import trap documented in `DataEntryGramplet/tests/test_data_entry_gramplet.py`: when `unittest` loads the file as `PersonRelationshipFilter.tests.test_...`, the outer `PersonRelationshipFilter` is already a namespace package in `sys.modules` by the time the test module's own import line runs, so a bare `from PersonRelationshipFilter import HasName` looks for `HasName` as an attribute of that namespace package instead of importing the `PersonRelationshipFilter.py` submodule that defines it.

Fix: use the fully-qualified `from PersonRelationshipFilter.PersonRelationshipFilter import (...)`, matching the pattern DataEntryGramplet already uses. Verified `python3 -m unittest PersonRelationshipFilter.tests.test_person_relationship_filter_rules -v` now passes all 19 tests.

dsblank:
@GaryGriffin this should be ready to merge.

PR #988, opened 2026-07-17
PersonRelationshipFilter (gramps52 port)

dsblank:
## Summary

Port of the **PersonRelationshipFilter** gramplet (gramps-project/addons-source#987) to the `maintenance/gramps52` branch, targeting `gramps_target_version="5.2"`.

A new gramplet, **Relationship Filter**, that drops into the Person view's Gramplet bar as a sidebar-style filter. Unlike the built-in Person filter (which only matches on the active person's own attributes), this one searches by the *names of relatives*:

- the person's own name
- father's / mother's name
- spouse's name
- up to two named siblings
- up to two named children
- a "probably alive on date" filter

Each name field is split into given-name and surname boxes searched independently (substring or regex), and results combine into a single `GenericFilter` applied to the Person view.

## Motivation

The stock Person sidebar filter can't answer "find people whose mother is named X" or "find the children of someone named Y" — you have to know the target person's own name or ID. This gramplet lets you search on that relationship directly instead of navigating the tree by hand.

## Test plan

- [x] `python3 -m pytest PersonRelationshipFilter/tests/ -v` — 19 passed

### DIFF
--- a/PersonRelationshipFilter/PersonRelationshipFilter.gpr.py
+++ b/PersonRelationshipFilter/PersonRelationshipFilter.gpr.py
@@ -1,0 +1,38 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2015      Nick Hall
+# Copyright (C) 2024      Paul Womack (BugBear)
+# Copyright (C) 2026      Doug Blank
+#
+# This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


dsblank:
@GaryGriffin this should also work on 5.1, if you alter the gpr.py file.

dsblank:
@GaryGriffin I think this is ready to merge.

PR #989, opened 2026-07-17
PersonRelationshipFilter (gramps51 port)

dsblank:
## Summary
- Adds the PersonRelationshipFilter gramplet, ported to the Gramps 5.1 API: a Person sidebar filter matching on father/mother/spouse/sibling/child names (substring or regex) plus a "probably alive" date filter.
- Guards `Filter.__filter_clicked` against re-entrant `build_tree()` calls (a suspected cause of an intermittent, non-reproducible segfault reported by a user running the 5.1 build).
- Fixes `help_url` to point at this addon's own wiki page (`Addon:PersonRelationshipFilter`) instead of a leftover reference to a different addon name.

Companion ports: #987 (6.1), #988 (5.2).

## Test plan
- [ ] `pytest PersonRelationshipFilter/tests/` passes
- [ ] Load example.gramps in Gramps 5.1, add the gramplet, search Father/Mother/Spouse/Sibling/Child fields, confirm results and no crash across repeated searches

### DIFF
--- a/PersonRelationshipFilter/PersonRelationshipFilter.gpr.py
+++ b/PersonRelationshipFilter/PersonRelationshipFilter.gpr.py
@@ -1,0 +1,38 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2015      Nick Hall
+# Copyright (C) 2024      Paul Womack (BugBear)
+# Copyright (C) 2026      Doug Blank
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+register(
+    GRAMPLET,
+    id="Person Relationship Filter",
+    name=_("Person Relationship Filter"),
+    description=_("Gramplet providing a person filter on relationships"),
+    version="1.0.0",
+    gramps_target_version="5.1",
+    status=STABLE,
+    fname="PersonRelationshipFilter.py",
+    height=200,
+    gramplet="PersonRelationshipFilter",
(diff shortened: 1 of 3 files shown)


dsblank:
@GaryGriffin I have no idea why this addon might have an issue with Macs. But I did make a change here.

GaryGriffin:
Uploaded change but it still intermittently crashes on 5.1.6 . 

I thought to file a bug with a crash report, but Mantis only goes back to 5.2.0, not 5.1.6 .

GaryGriffin:
@dsblank Do you want me to hold this until the Mac crash is resolved? I suspect it is a core issue with a system library, which will not be fixed (would require a new bundle). Or do you want me to publish with the known caveat it is unstable on Mac on 5.1 and the fix is to upgrade to at least 5.2? 

Can you add the wiki help page as it is blank.

dsblank:
@emyoulation do you have an opinion on:

> Do you want me to hold this until the Mac crash is resolved? I suspect it is a core issue with a system library, which will not be fixed (would require a new bundle). Or do you want me to publish with the known caveat it is unstable on Mac on 5.1 and the fix is to upgrade to at least 5.2?

emyoulation:
I suspect that sharing it now as a separate 5.1 "experimental status" addon will be the most effective approach. 

That will allow a broader cross-section of users to experiment and comment. But segregating it from the 5.2+ version will decouple any bloat from that might not be necessary with the other versions.  (I suspect that the macOS 5.1 install bundle might have skipped some libraries. It might be necessary to bundle those in the 5.1 addon.) 

And it will allow native macOS users to experiment with Claude (or their favorite AI of the moment) interactively to troubleshoot or to add diagnostics.

GaryGriffin:
> (I suspect that the macOS 5.1 install bundle might have skipped some libraries. It might be necessary to bundle those in the 5.1 addon.)
> 

Not really possible to overwrite in the addon a portion of core functionality. When I have seen this type of crash on Mac, it is often due to gtk or cairo bugs. System core (crash) file indicates this is likely the area of issue, but I am not skilled in diagnosing this. There was nothing 'skipped' in the Mac 5.1 bundle (This is what I use for my production system). 

> And it will allow native macOS users to experiment with Claude (or their favorite AI of the moment) interactively to troubleshoot or to add diagnostics.

I suspect that the intersection of (Mac User | Gramps 5.1 User | AI User) is the NULL set. IMO, the response is to have them upgrade to at least 5.2 if they want this functionality. 

My suggestion is to add a known issue to the Wiki Help page for the addon and I can merge/publish it on 5.1 too. I want to avoid bug reports since this is a hard crash of gramps.

dsblank:
@GaryGriffin IMHO you can close this.

GaryGriffin:
Closing PR. The README specifically states the min branch is 5.2 .

PR #993, opened 2026-07-19
Add WordClouds gramplet

dsblank:
## Summary

Adds three word-cloud gramplets — **Given Name Word Cloud**, **Surname Word
Cloud**, and **Place Word Cloud** — that render names/places as a
size-weighted, click-to-navigate word cloud instead of the old plain-text
link list.

Screenshot from 2026-07-19 09-33-52

- `wordcloudwidget.py` — a Cairo/Pango `WordCloudWidget` (`Gtk.DrawingArea`)
  that lays out words with a spiral placement algorithm, sized and colored
  by frequency.
- `slideroption.py` — a `SliderOption`/`GuiSliderOption` pair (a
  `NumberOption` rendered as a slider + text entry), registered via
  `BasePluginManager.register_option()` so **no changes to Gramps core are
  required**.
- `cloudgramplet.py` — shared `CloudGramplet` base class handling the
  common option set (top-N count, colors, layout quality, filter
  missing/unknown values) and word-cloud wiring.
- `givennamewordcloudgramplet.py`, `surnamewordcloudgramplet.py`,
  `placewordcloudgramplet.py` — the three concrete gramplets.

This is based on ideas from ClmntPnd's Gramps core PR
[gramps-project/gramps#2223](https://github.com/gramps-project/gramps/pull/2223),
which added the WordCloudWidget-based rendering (including a Place Cloud
gramplet and `autosave_options` support) directly to Gramps core. This PR
adapts that work into a self-contained addon instead, so it:

- Works with unmodified, current stable Gramps releases (no core patches).
- Uses new gramplet ids/class names (`Given Name Word Cloud` /
  `GivenNameWordCloudGramplet`, `Surname Word Cloud` /
  `SurnameWordCloudGramplet`, `Place Word Cloud` /
  `PlaceWordCloudGramplet`) that are distinct from the built-in
  `Given Name Cloud` / `Surname Cloud` gramplets, to avoid any plugin
  registration collision between the addon and the core gramplets of the
  same name.
- Uses the standard Save-button option flow (rather than the core-patched
  `autosave_options` instant-apply behavior), since that required a
  `grampletpane.py` change.

## Test plan

- [x] `tests/test_imports.py` — all modules import cleanly; each gramplet
      class is a proper `Gramplet` subclass.
- [x] `tests/test_wordcloudwidget.py` — font-size/color interpolation and
      bounding-box overlap helpers.
- [x] `tests/test_slideroption.py` — `SliderOption` value/min/max/step
      behavior and `GuiSliderOption` widget wiring (run under `xvfb-run`,
      since it needs a real display connection).
- [x] `tests/test_cloudgramplet_logic.py` — `get_items()` for each
      gramplet against a real in-memory Gramps database, including the
      `filter_missing` option and backlink-based place filtering.
- [x] Manual: install the addon, open each of the three gramplets in a
      running Gramps 6.1, verify word clicks navigate correctly and
      options apply on Save.

### DIFF
--- a/WordClouds/WordClouds.gpr.py
+++ b/WordClouds/WordClouds.gpr.py
@@ -1,0 +1,67 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2026  Douglas S. Blank <doug.blank@gmail.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+register(
+    GRAMPLET,
+    id="Given Name Word Cloud",
+    name=_("Given Name Word Cloud"),
+    description=_("Gramplet showing all given names as a word cloud"),
+    status=STABLE,
+    version="1.0.0",
+    fname="givennamewordcloudgramplet.py",
+    height=300,
+    expand=True,
+    gramplet="GivenNameWordCloudGramplet",
+    gramplet_title=_("Given Name Word Cloud"),
+    gramps_target_version="6.1",
+    help_url="WordClouds",
+)
+
+register(
+    GRAMPLET,
+    id="Surname Word Cloud",
+    name=_("Surname Word Cloud"),
+    description=_("Gramplet showing all surnames as a word cloud"),
+    status=STABLE,
+    version="1.0.0",
+    fname="surnamewordcloudgramplet.py",
+    height=300,
+    expand=True,
+    gramplet="SurnameWordCloudGramplet",
+    gramplet_title=_("Surname Word Cloud"),
+    gramps_target_version="6.1",
+    help_url="WordClouds",
+)
+
+register(
+    GRAMPLET,
+    id="Place Word Cloud",
+    name=_("Place Word Cloud"),
+    description=_("Gramplet showing all places as a word cloud"),
+    status=STABLE,
+    version="1.0.0",
+    fname="placewordcloudgramplet.py",
+    height=300,
+    expand=True,
+    gramplet="PlaceWordCloudGramplet",
+    gramplet_title=_("Place Word Cloud"),
+    gramps_target_version="6.1",
+    help_url="WordClouds",
+)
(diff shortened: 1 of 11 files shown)


GaryGriffin:
One unittest had warning:

```
python3 -m unittest WordClouds.tests.test_imports     
/Users/GrampsTest61/gtk/inst/lib/python3.14/site-packages/gi/overrides/__init__.py:159: PyGIDeprecationWarning: GLib.unix_signal_add_full is deprecated; use GLibUnix.signal_add_full instead
  value = getattr(proxy, attr)
....
Ran 4 tests in 0.448s

OK
```

Another unittest popped up a WarningDialog. This could be Mac-specific as gexiv2 is built within the Mac bundle and is not in a separate pip package. Plz confirm that you do not see this on Linux/Win.

Screenshot 2026-07-24 at 10 39 46 AM


```
python3 -m unittest WordClouds.tests.test_cloudgramplet_logic

/Users/GrampsTest61/gtk/inst/lib/python3.14/site-packages/gi/events.py:803: DeprecationWarning: 'asyncio.AbstractEventLoopPolicy' is deprecated and slated for removal in Python 3.16
  AbstractEventLoopPolicy = getattr(asyncio, "AbstractEventLoopPolicy", object)
/Users/GrampsTest61/gtk/inst/lib/python3.14/site-packages/gi/events.py:890: DeprecationWarning: 'asyncio.get_event_loop_policy' is deprecated and slated for removal in Python 3.16

.......
Ran 7 tests in 5.238s

OK

```
@dsblank - Plz confirm if I can merge these with these warnings.
@eduralph - any issues with CI if these tests are merged?

dsblank:
@GaryGriffin The deprecation warnings in `test_imports` are harmless (upstream PyGObject/asyncio noise, not from this addon).

I dug into the WarningDialog and it's not a WordClouds bug — it's core Gramps. `gramps/plugins/gramplet/gramplet.gpr.py` has module-level code that pops a `MessageHideDialog` ("GExiv2 module not loaded") whenever `has_display()` returns `True` and GExiv2 isn't importable. Our test's `make_database("sqlite")` call goes through `BasePluginManager.get_instance()` (`gramps/gen/db/utils.py`), which scans *every* registered `.gpr.py`, including that core one — so this fires for any addon test that ends up touching the plugin manager, on any machine with a live display and no GExiv2, not just WordClouds. On Linux here it doesn't reproduce because there's no display attached in that environment.

Your read on it being Mac-specific was right — it's the classic Mac GExiv2-bundling gap, unrelated to this PR's code.

For now: re-running with `GDK_BACKEND=-` set (per this repo's documented test-running convention) should suppress it, since that makes `has_display()` return `False` even with a physical display attached.

@eduralph — heads up, this also affects the new shared harness on gramps60: `neutralize_gui_dialogs()` in `tests/gramps_test_env.py` stubs out `ErrorDialog`/`WarningDialog`/`OkDialog`/etc. but not `MessageHideDialog`, which is what core's `gramplet.gpr.py` uses here. Since `get_plugin_manager_and_registry()` scans `PLUGINS_DIR` (core's own plugins tree, not just addons), `test_plugin_registration.py` / `make.py test` could hang on this exact modal on a GExiv2-less host with a display. Not a problem in the CI container itself since the Dockerfile bundles `gir1.2-gexiv2-0.10`, but it'll bite local runs. Following up with a comment on #980 about it.

Safe to merge WordClouds as-is from my side.

PR #967, opened 2026-06-29
Update associationstool.py to v1.2.1(gramps52)

romjeromealt:

PR #996, opened 2026-07-20
Fix trailing whitespace and a shadowed gettext _ in AssociationsTool

eduralph:
## Problem

Two pre-existing defects on `maintenance/gramps60` that the addon CI pipeline (#820) surfaces the moment its lint gate runs:

1. **Trailing whitespace** on three tracked Python files (27 lines): `ArchiveAssist/ArchiveAssist.py` (22), `FilterRules/matcheventfilterrole.py` (4), `FilterRules/matchparentoffilterfamily.py` (1). The pipeline's trailing-whitespace check fails on them.

2. **A real latent bug in AssociationsTool**: `_build_with_progress()` uses `for _ in range(batch_size)`, which makes `_` a function-local throughout the method and shadows the module-level `_ = _trans.gettext`. On the error paths this raises `UnboundLocalError` (the `_("Error initializing data: %s")` handler, which references `_` before the loop binds it) and `TypeError` (the in-loop `_("Error processing person …")` handler, which calls `_` after the loop bound it to an int). ruff `F82` flags it.

## Fix

- Strip the trailing whitespace (whitespace-only; `git diff -w` is empty).
- Rename the unused loop counter to `_batch_step` so `_` resolves to the gettext function everywhere in the method.

Four files, no pipeline/CI files. Splitting these out of #820 so that PR stays pipeline-only; #820's lint lane goes green once this merges and is merged back. Found while running #820's own gates against the tree during an adversarial review of that PR.

### DIFF
--- a/ArchiveAssist/ArchiveAssist.py
+++ b/ArchiveAssist/ArchiveAssist.py
@@ -94,7 +94,7 @@ def parse_ref(text: str) -> dict:
     match = pattern.search(text)
     if not match:
         return {}
-    
+
     archive = match.group("archive").strip()
     to_remove = ["kyrkoarkiv", "stadsarkiv"]
     for word in to_remove:
@@ -127,7 +127,7 @@ def parse_ref(text: str) -> dict:
 # Gramplet class
 # ------------------------
 class ArchiveAssist(Gramplet):
-    
+
     # I am bad at GUI. Double check...
     def init(self):
         if getattr(self, "_initialized", False):
@@ -170,7 +170,7 @@ class ArchiveAssist(Gramplet):
         vbox.pack_start(self.status_label, False, False, 0)
 
         return vbox
-    
+
     def get_or_create_repository(self, name, trans):
         for handle in self.dbstate.db.get_repository_handles():
             repo = self.dbstate.db.get_repository_from_handle(handle)
@@ -195,11 +195,11 @@ class ArchiveAssist(Gramplet):
         if not parsed:
             self.status_label.set_text("Could not parse the reference string.")
             return
-            
+
         # Find existing Source (by title)
         src = None
         src_handle = None
-        
+
         for handle in self.dbstate.db.get_source_handles():
             candidate = self.dbstate.db.get_source_from_handle(handle)
             for attr in candidate.get_attribute_list():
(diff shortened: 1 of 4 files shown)


romjeromealt:
> A real latent bug in AssociationsTool: _build_with_progress() uses for _ in range(batch_size), which makes _ a function-local throughout the method and shadows the module-level _ = _trans.gettext. On the error paths this raises UnboundLocalError (the _("Error initializing data: %s") handler, which references _ before the loop binds it) and TypeError (the in-loop _("Error processing person …") handler, which calls _ after the loop bound it to an int). ruff F82 flags it.

My bad, I did not see it during review for #976
Also, testing was under gramps52. Also, as translation strings (_related to _ variable_) were loaded at first runtime (or addon load), such backend hidden method did not always raise an error? e.g., batch size < 50. Anyway, good catch. Thanks!

romjeromealt:
There was a blind change (_according to my working branch_) on #967 (_gramps52_)
Also read that it seems to have a change on translation handling into gramps60 and +.

IA pointed out possible issues, but this one was not explicit (there was two locations...), despite matched.
```
Solution:

    Never use _ or LOG as variable names in loops or unpacking.
    Use explicit variable names (e.g., privacy, citation, note) if you must unpack tuples.
    Better yet: avoid unpacking entirely by using the public API (Fix 1).

Example of What to Avoid:

for ref in person.serialize()[-1]:
    _, _, _, two, value = ref  # ❌ Shadows gettext's _
    # or
    LOG, _, _, two, value = ref  # ❌ Shadows the LOG logger
```

https://github.com/gramps-project/addons-source/pull/967#issuecomment-4844075549

eduralph:
> > A real latent bug in AssociationsTool: _build_with_progress() uses for _ in range(batch_size), which makes _ a function-local throughout the method and shadows the module-level _ = _trans.gettext. On the error paths this raises UnboundLocalError (the _("Error initializing data: %s") handler, which references _ before the loop binds it) and TypeError (the in-loop _("Error processing person …") handler, which calls _ after the loop bound it to an int). ruff F82 flags it.
> 
> My bad, I did not see it during review for #976 Also, testing was under gramps52. Also, as translation strings (_related to _ variable_) were loaded at first runtime (or addon load), such backend hidden method did not always raise an error? e.g., batch size < 50. Anyway, good catch. Thanks!

No worries - it surfaced during my work on a CI for addons-source, so it's doing what it's supposed to. :-)

PR #943, opened 2026-06-05
PostgreSQL: replace _hack_query() with proper dialect methods

dsblank:
## Summary

- Removes the monolithic \`_hack_query()\` string-substitution function from the \`Connection\` class
- Adds \`_sql_type()\` override on \`PostgreSQL\` to return \`bytea\` instead of \`BLOB\`
- Adds \`_quote_column()\` override to handle PostgreSQL reserved words: returns \`desc_\` for \`desc\` (and any other reserved words), preserving the column name used by all existing PostgreSQL databases and avoiding any migration
- Overrides \`_create_secondary_columns()\`, \`_update_secondary_values()\`, and \`get_media_handles()\` on \`PostgreSQL\` to call \`self._quote_column()\` at each SQL generation site — so reserved words are quoted before the SQL is built, not patched after
- \`Connection.execute()\` handles the remaining SQL syntax differences:
  - \`?\`→\`%s\` (paramstyle)
  - \`REGEXP\`→\`~\` (regex operator)
  - \`LIMIT offset, count\`→\`LIMIT count OFFSET offset\` (SQLite two-arg form → standard SQL), via inline lambda
  - \`LIMIT -1\`→\`LIMIT ALL\` (SQLite unlimited → PostgreSQL)
  - \`ON media(desc)\`→\`ON media(desc_)\` — targeted fix for the one \`CREATE INDEX\` site in \`_create_schema\` that has no override hook (marked TODO for removal once gramps PR #2178 is merged)
- Adds \`PostgreSQL/tests/test_sql_translations.py\` — 20 unit tests covering all rewrite rules, \`_sql_type\`, and \`_quote_column\`; psycopg2 is stubbed so no real database is needed

## Relationship to upstream

A companion PR [gramps #2178](https://github.com/gramps-project/gramps/pull/2178) adds a \`_quote_column()\` no-op hook to the \`DBAPI\` base class and calls it at the \`CREATE INDEX\` and \`ORDER BY\` sites. Once that is merged, the three method overrides and the targeted \`ON media(desc)\` replacement in \`execute()\` can all be removed — each is marked with a comment pointing to PR #2178. The \`_quote_column()\` definition, \`_sql_type()\` override, and the \`execute()\` paramstyle/operator translations remain as legitimately PostgreSQL-specific.

## Test plan

- [x] 20 unit tests pass (\`python3 -m pytest PostgreSQL/tests/test_sql_translations.py -v\`)
[...]

PR #1001, opened 2026-07-28
Update SharedPostgreSQL to match PostgreSQL addon

DavidMStraub:
This PR backports @dsblank's recent improvements to the PostgreSQL addon (#943) to the SharedPostgreSQL addon (which is used by many Gramps Web deployments).

Claude's summary:

- Replace the blind `_hack_query()` substring rewriting (`"?"→"%s"`, `"desc"→"desc_"`, `"BLOB"→"bytea"`, etc.) with targeted, word-boundary-safe SQL dialect translation and a `_quote_column()` hook for reserved-word columns. The old approach corrupted any identifier containing "desc" as a substring (e.g. `description` silently became `desc_ription` in every query).
- Fix PostgreSQL collation check to work on PostgreSQL < 12 by querying `pg_collation` instead of relying on `CREATE COLLATION IF NOT EXISTS`.
- Fix a parameter-count mismatch in `_commit_raw()`'s INSERT branch (3 columns, only 2 `?` placeholders), which would raise a database error if ever triggered. Only reachable via legacy schema-upgrade code paths (schema version 14–16, roughly Gramps 3.3–4.1 era), so unlikely to have affected real trees in practice.
- Add a `dialect` class attribute and `_sql_type()` BLOB→bytea mapping for forward-compatibility with upstream Gramps dialect support.
- Add unit tests covering the SQL translation rules and column naming, adapted for this addon's schema.

Notes:

- Existing databases already have physical columns named `desc_` and `desc_ription` (artifacts of the old blind substring rewrite). These legacy names are preserved via `_quote_column()` rather than renamed, so no migration is required for existing trees.

### DIFF
--- a/SharedPostgreSQL/shareddbapi.py
+++ b/SharedPostgreSQL/shareddbapi.py
@@ -289,7 +289,9 @@ class SharedDBAPI(DbGeneric):
         self.dbapi.execute(
             "CREATE INDEX citation_gramps_id " "ON citation(treeid,gramps_id)"
         )
-        self.dbapi.execute("CREATE INDEX media_desc " "ON media(treeid,desc)")
+        self.dbapi.execute(
+            f"CREATE INDEX media_desc ON media(treeid,{self._quote_column('desc')})"
+        )
         self.dbapi.execute("CREATE INDEX media_gramps_id " "ON media(treeid,gramps_id)")
         self.dbapi.execute("CREATE INDEX place_title " "ON place(treeid,title)")
         self.dbapi.execute(
@@ -699,7 +701,7 @@ class SharedDBAPI(DbGeneric):
             self.dbapi.execute(
                 "SELECT handle FROM media "
                 "WHERE treeid = ? "
-                "ORDER BY desc "
+                f"ORDER BY {self._quote_column('desc')} "
                 'COLLATE "%s"' % locale.get_collation(),
                 [self.dbapi.treeid],
             )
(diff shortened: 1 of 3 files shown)


dsblank:
@DavidMStraub can you also include this changed:

https://github.com/gramps-project/addons-source/pull/1005

DavidMStraub:
Both 6.0 and 6.1, right?

dsblank:
FYI: One issue that I just discovered as I'm working on gramps-web-api: the old hack changed "description" into "desc_ription". This may have a backwards compatibility effect as we switch to less hacky ways of handling SQL dialects.

DavidMStraub:
> FYI: One issue that I just discovered as I'm working on gramps-web-api: the old hack changed "description" into "desc_ription". This may have a backwards compatibility effect as we switch to less hacky ways of handling SQL dialects.

Yes, Claude found this out and already included it in this PR:

https://github.com/DavidMStraub/addons-source/blob/d50c7c94afa45e876343da338aa5a7edc08d484c/SharedPostgreSQL/sharedpostgresql.py#L58-L62

DavidMStraub:
@GaryGriffin I'll create a follow-up PR matching #1005, not include it in this one.

GaryGriffin:
So this is ready to merge? 
I will cherrypick to gramps61 also.

DavidMStraub:
Yes, ready!

PR #1007, opened 2026-08-02
Add GOQLFilter addon

dsblank:
## Screen Capture Example

image

## Summary
- Adds a gramps-object-query-language (GOQL) filter gramplet and a matching "MatchesExpression" custom-filter rule for each primary object type (Person, Family, Event, Place, Repository, Source, Citation, Media, Note).
- The gramplet is a multi-line where-expression editor (Tab completion, syntax highlighting, session-scoped history) that runs the expression as a real `GenericFilter` via `view.generic_filter` -- the same mechanism the built-in rule-based sidebar filter uses -- and can hand the same expression to Gramps' own `EditFilter` dialog to save it as an ordinary Custom Filter, reusable anywhere a Custom Filter is (other views' sidebar filters, reports, exports), not just in this gramplet.
- `MatchesExpression` pushes the expression down to SQL when the database is unproxied and DB-API-backed (SQLite, or the SharedPostgreSQL addon), falling back to per-object evaluation otherwise -- the same fast/slow split gramps-web-api's `object_query` resource makes.
- Requires the `gramps-object-query-language` package (`pip install gramps-object-query-language`); `requires_mod` in the `.gpr` registrations surfaces that as a Plugin Manager prerequisite.

## Connections

1. https://github.com/dsblank/gramps-object-query-language - the query language
2. https://gramps-project.org/wiki/index.php?title=Addon:GrampsObjectQueryLanguage - help page

## Test plan
- [x] `black --check` passes on all files in `GOQLFilter/`
- [x] Unit tests included under `GOQLFilter/tests/` (`test_goql.py`, `test_goql_completion.py`, `test_goql_completion_popup.py`, `test_goql_highlight.py`) plus an integration test (`test_integration_whereexprrule.py`)
- [ ] CI run of the test suite (this addon's tests rely on the repo-root GTK3 pin in `tests/__init__.py`, not exercised when run standalone on a host that defaults to GTK4)

### DIFF
--- a/GOQLFilter/goql.gpr.py
+++ b/GOQLFilter/goql.gpr.py
@@ -1,0 +1,199 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2026       Douglas Blank
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+"""Gramplets providing a gramps-object-query-language filter for each
+primary object view."""
+
+_HELP = "Addon:GrampsObjectQueryLanguage"
+_AUTHORS = ["Douglas Blank"]
+_AUTHORS_EMAIL = ["doug.blank@gmail.com"]
+
+register(
+    GRAMPLET,
+    id="Person GOQL Filter",
+    name=_("Person GOQL Filter"),
+    description=_("Gramplet providing a gramps-object-query-language person filter"),
+    version="1.0.0",
+    gramps_target_version="6.1",
+    status=STABLE,
+    fname="goql.py",
+    height=260,
+    gramplet="PersonQueryFilter",
+    gramplet_title=_("GOQL Filter"),
+    navtypes=["Person"],
+    authors=_AUTHORS,
+    authors_email=_AUTHORS_EMAIL,
+    help_url=_HELP,
+    requires_mod=["gramps_object_query_language"],
+)
+
+register(
+    GRAMPLET,
+    id="Family GOQL Filter",
+    name=_("Family GOQL Filter"),
+    description=_("Gramplet providing a gramps-object-query-language family filter"),
+    version="1.0.0",
+    gramps_target_version="6.1",
+    status=STABLE,
+    fname="goql.py",
+    height=260,
+    gramplet="FamilyQueryFilter",
+    gramplet_title=_("GOQL Filter"),
+    navtypes=["Family"],
+    authors=_AUTHORS,
+    authors_email=_AUTHORS_EMAIL,
+    help_url=_HELP,
+    requires_mod=["gramps_object_query_language"],
+)
+
+register(
+    GRAMPLET,
+    id="Event GOQL Filter",
+    name=_("Event GOQL Filter"),
+    description=_("Gramplet providing a gramps-object-query-language event filter"),
+    version="1.0.0",
+    gramps_target_version="6.1",
+    status=STABLE,
+    fname="goql.py",
+    height=260,
+    gramplet="EventQueryFilter",
+    gramplet_title=_("GOQL Filter"),
+    navtypes=["Event"],
+    authors=_AUTHORS,
+    authors_email=_AUTHORS_EMAIL,
+    help_url=_HELP,
+    requires_mod=["gramps_object_query_language"],
+)
+
+register(
+    GRAMPLET,
+    id="Place GOQL Filter",
+    name=_("Place GOQL Filter"),
+    description=_("Gramplet providing a gramps-object-query-language place filter"),
+    version="1.0.0",
+    gramps_target_version="6.1",
+    status=STABLE,
+    fname="goql.py",
+    height=260,
+    gramplet="PlaceQueryFilter",
+    gramplet_title=_("GOQL Filter"),
+    navtypes=["Place"],
+    authors=_AUTHORS,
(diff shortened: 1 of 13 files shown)


GaryGriffin:
I ran unittests as:
`python3 -m unittest tests.test_make_listing  GOQLFilter.tests.TESTNAME`
and all ran successfully after putting `gramps_object_query_language` into the GOQLFilter directory manually on Mac Gramps6.1 beta. 

Integration test further required prepending `GDK_BACKEND=- ` or it gets stuck waiting for user input.

dsblank:
Claude says that testing works like:

```
export GDK_BACKEND=-
export GRAMPS_RESOURCES=/path/to/gramps/build/share   # if gramps isn't already resolving it
python3 -m unittest tests.test_make_listing GOQLFilter.tests.test_goql \
    GOQLFilter.tests.test_goql_completion_popup \
    GOQLFilter.tests.test_goql_highlight \
    GOQLFilter.tests.test_goql_completion \
    GOQLFilter.tests.test_integration_whereexprrule -v
```

@GaryGriffin Is that not correct? Please let me know exactly what should work and I will make it so.

dsblank:
Claude also says:

docs/addon-development/07-testing.md now states even more strongly: "MUST: never call gi.require_version... The corollary: run from the repository root. The pins execute when the root tests package loads." I tested that literally:

```
cd addons-source
GDK_BACKEND=- python3 -m unittest GOQLFilter.tests.test_goql -v

File ".../GOQLFilter/goql.py", line 74, in <module>
    from gramps.gui.editors import EditFilter
File ".../gramps/gui/editors/editaddress.py", line 41, in <module>
ValueError: Namespace Gtk is already loaded with version 4.0
```

Same crash as before. Running from the repo root, by itself, does not import the root tests/__init__.py — importing GOQLFilter.tests.test_goql only pulls in the GOQLFilter and GOQLFilter.tests packages, which are unrelated to the top-level tests package. That only happens if a root-level test module is named explicitly in the same invocation (what the reviewer did with tests.test_make_listing) or via a full-repo discover that happens to touch it first. The doc's "run from root, dotted path" claim doesn't hold for a single-module invocation on this machine — I have a live repro if you want to see it directly.

Interesting side detail from the traceback: gramps/gui/editors/editaddress.py itself calls gi.require_version("Gtk", "3.0") at import time — so core Gramps GUI modules don't actually rely purely on a one-time startup pin either; they re-pin defensively at each import site. That undercuts the doc's premise that "the pin is never the addon's job."

dsblank:
CC: @eduralph

GaryGriffin:
The `tests.test_make_listing` include the pin, hence why I put that before the specific test. Here is what I did (I run my tests from a sudo terminal window which does not have a gfx head, and the integration test pops a gfx window)

**SETUP:**
- Copy gramps_object_query_language folder to the GOQFilter folder so unittest could find it. 
- `jhbuild shell` to get an environment that has gramps in the PATH. This is Mac-specific.

**TESTING:**
```
python3 -m unittest tests.test_make_listing  GOQLFilter.tests.test_goql
python3 -m unittest tests.test_make_listing  GOQLFilter.tests.test_goql_completion
python3 -m unittest tests.test_make_listing  GOQLFilter.tests.test_goql_completion_popup
python3 -m unittest tests.test_make_listing  GOQLFilter.tests.test_goql_highlight
GDK_BACKEND=- python3 -m unittest tests.test_make_listing  GOQLFilter.tests.test_integration_whereexprrule
```

The warnings I got were not related to your tests. On Integration test:
```
/Users/GrampsTest61/gtk/inst/lib/python3.14/site-packages/gi/overrides/__init__.py:159: PyGIDeprecationWarning: GLib.unix_signal_add_full is deprecated; use GLibUnix.signal_add_full instead
  value = getattr(proxy, attr)
```

dsblank:
> /Users/GrampsTest61/gtk/inst/lib/python3.14/site-packages/gi/overrides/__init__.py:159: PyGIDeprecationWarning: GLib.unix_signal_add_full is deprecated; use GLibUnix.signal_add_full instead
>  value = getattr(proxy, attr)

Not sure what that is :shrug:

dsblank:
Claude says:

It's a known, harmless PyGObject quirk, not a bug in your PR.

What's happening: PyGObject's gi.overrides.__init__ (the file/line in the traceback) fires this PyGIDeprecationWarning merely from importing Gio/GLib on new GLib builds (2.88.0+) — unix_signal_add_full moved to a separate GLibUnix module, and PyGObject warns about it during the override-proxy setup regardless of whether the code ever calls that function. It's known upstream: GLib.unix_signal_add_full raises deprecation warning on import, not use (pygobject#757).

Why it showed up on Gary's run: GOQLFilter/goql.py:59 does from gi.repository import Gdk, GLib, Gtk. That import is enough to trigger PyGObject's warning on his environment (/Users/GrampsTest61/gtk/inst/...python3.14..., i.e. a bleeding-edge gtk-osx build with a very new GLib/PyGObject). No code in this PR calls unix_signal_add_full — I checked, and neither GOQLFilter nor Gramps core references it.

Bottom line: it's a pre-existing environment/PyGObject-version artifact, not something introduced by the GOQLFilter addon, and it would fire for any addon or Gramps module that imports GLib on his setup. It's a warning, not an error, and shouldn't block the PR. Worth telling Gary it's cosmetic/environmental and unrelated to the addon's logic.

Doug says:

I filtered out the warning for clean output.

dsblank:
@Nick-Hall it sure would be nice to have an updated beta/rc release (especially on mac) so that @GaryGriffin can actually test everything, including the mini-installer.

Nick-Hall:
> would be nice to have an updated beta/rc release

I agree.  I've been chasing the progress of the FamilySearch integration, but nothing seems to be happening.

It is probably time for another beta. I'll take a look at the outstanding PRs.

GaryGriffin:
Do we really want to selectively suppress this specific warning in this specific test? I have other similar warnings on other unittests - are we going to selectively suppress those too? There are 16 addons that import GLib, for instance. Do all these need addressed?

- python3 -m unittest tests.test_make_listing Form.tests.test_integration_form
- python3 -m unittest tests.test_make_listing SurnameMappingGramplet.tests.test_surnamemappinggramplet_imports
- python3 -m unittest tests.test_make_listing SourceReferences.tests.test_sourcereferences_imports

I can just ignore these when found in testing, as they are not relevant to the PR that is being tested.

dsblank:
> I can just ignore these when found in testing, as they are not relevant to the PR that is being tested.

Ok :smile: In another thread there was a suggestion that noisy output should be prevented. If it doesn't bother you, it is fine to leave them in.

PR #1009, opened 2026-08-04
Add GrampsWebApiDb: use a Gramps Web API server as a live database backend

dsblank:
## Summary

`GrampsWebApiDb` is a `DATABASE`-type addon that lets Gramps (desktop or
scripted) open a [gramps-web-api](https://github.com/gramps-project/gramps-web-api) server (e.g. a [gramps-connect](https://github.com/dsblank/gramps-connect) or Gramps
Web instance) as a regular family tree — read and write, no
export/import step, no separate sync tool to run by hand.

It's not fast (every write is a network round trip), but it means the
same tree hosted online is simultaneously usable from Gramps desktop and
from the web app, live.

## How it works

Rather than implementing `DbReadBase`/`DbWriteBase` from scratch (~175
methods), `WebApiDB` subclasses the stock `SQLite` DBAPI backend.
`DbGeneric` already implements every `get_*`/`iter_*`/`commit_*`/
`remove_*` method generically on top of a small SQL connection, so this
addon only needs to supply two things:

- **A local SQLite mirror** for fast reads, kept in sync via
  `GET /transactions/history/?after=<high-water-mark>` — the same
  per-object transaction log the server's own undo system uses. The
  high-water-mark is persisted in the mirror's own metadata table
  (`_get_metadata`/`_set_metadata`), so reopening an existing tree is
  incremental, not a full re-download.
- **Write-through**, hooked at `transaction_commit()` (the single point
  `DbTxn.__exit__` calls after any completed local transaction) rather
  than each individual `commit_person`/`commit_family`/etc. Local edits
  push to `POST /transactions/` automatically.

`webapi_client.py` is a small, self-contained HTTP client (stdlib
`urllib` only, no extra dependency) — trimmed from and crediting the
`WebApiHandler` class in the `GrampsWebSync` addon (same repo). It's now
a hand-synced vendored copy of a standalone `gramps-api-client`
package (not yet published; same code, class renamed `Client`) — the
addon keeps its own copy since Gramps addons can't declare a pip
dependency. That standalone package also ships a
`gramps-api-client generate-key` CLI as an alternative to the
`mint_api_key()` script-based flow described below.

## Credentials

A single environment variable, `GRAMPS_WEB_API_KEY`, shaped
`"<REFRESH_TOKEN>*<BASE64URL(URL)>"`. This is deliberately *not* a login
dialog: `requires_login()` returns `False`, and the same env var also
works as a bare SDK credential (`WebApiHandler.from_env()`) for scripts
that talk to the server directly, without Gramps involved at all — one
credential, two consumers.

The `TOKEN` half is a non-expiring JWT refresh token
(`gramps-web-api`'s `JWT_REFRESH_TOKEN_EXPIRES` is `False` by default),
obtained once via `WebApiHandler.mint_api_key(url, username, password)`
(or the `gramps-api-client generate-key` CLI mentioned above). This
is a **known, deliberate stopgap**, not a real personal-access-token:
`gramps-web-api` has scoped, revocable token infrastructure already
(`AccessToken`, hashed, per-scope) but it's currently hardcoded to a
single scope (`anniversaries_ics`) and not wired into general request
auth. Until that's generalized server-side, this key is exactly as
powerful as the account it was minted from, and — unlike a real
personal-access-token — isn't individually revocable (a password change
doesn't invalidate it). This tradeoff and the reasoning are documented
in `webapi_client.py`'s module docstring.

## Usage

1. Mint a `GRAMPS_WEB_API_KEY` -- it bakes in the host, and logs in as a
   particular user/database on that server. Via the CLI, for example:

```python
$ gramps-api-client generate-key --url "http://localhost:5003/api" --user "gramps"
Password:...
<the key>
```

2. `export GRAMPS_WEB_API_KEY="<the key>"`
3. Add the addon in Gramps 
4. Start Gramps; in Preferences, change the database backend to "Gramps
   Web API Database".

image

5. Create a new family tree. The type just reflects what the `GRAMPS_WEB_API_KEY` points to at that time.

image

Gramps opens the tree already populated with the server's data (the
initial sync), and from there it behaves like any other family tree:
edit, delete, undo, and redo all work, with every change pushed back to
the server live.

## Status

- ✅ Read sync (`_sync_from_server`), incremental, verified against a
  live server
- ✅ Write-through (`transaction_commit` hook), verified add/update/delete
  round-trip against a live server, confirmed via an independent fresh
  mirror that pushes actually reach the server
- ✅ Verified live in Gramps desktop itself (not just scripts): backend
  selectable in the New Family Tree dialog, syncs and edits correctly
- ✅ Push conflict detection — pushes no longer go out with `force=1`;
  the server's optimistic-concurrency check (comparing each item's
  pre-edit snapshot against its current data) now actually runs, and a
  rejection raises `WebApiPushConflict`, which `transaction_commit()`
  catches to resync the local mirror from the server instead of
  silently overwriting. Real conflict *resolution* (merge, prompt the
  user) is still out of scope — the losing local edit is simply lost
  from the server's perspective.
- ✅ Undo/redo integration — `undo()`/`redo()` are overridden to push to
  the server too, not just the local mirror: undo sends the original
  transaction's payload to `POST /transactions/?undo=1` (the server
  reverses it itself), redo just re-pushes it forward, same as an
  ordinary commit. Both go through the same conflict-detection/resync
  path as a plain commit. Verified end-to-end against a live server
  (add a person, undo, confirm via a fresh mirror sync that the server
  no longer has it, redo, confirm it's back). Gramps' own undo history
  is in-memory/per-session, not persisted, so this only matters within
  a single running session.
- ❌ No media file sync

Registered `status=UNSTABLE` to reflect the remaining gaps. Targets
`gramps_target_version="6.0"`.

## Testing

- Automated: `python3 -m unittest GrampsWebApiDb.tests.test_webapi_client
  GrampsWebApiDb.tests.test_grampswebapidb -v` — 67 tests, covering
  `webapi_client.py`'s auth/retry/rate-limit/conflict logic and
  `grampswebapidb.py`'s sync/apply-change/transaction-commit/undo-redo
  logic, with `urlopen` and the SQLite base class mocked (no live server
  needed).
- Manual, end-to-end against a real gramps-web-api server: read sync,
  write-through, incremental re-sync, push-conflict detection + resync,
  undo/redo (round-tripped through independent fresh mirrors to confirm
  the server's state, not just the local one), and live use from Gramps
  desktop (New Family Tree dialog, backend selectable, syncs and edits
  correctly). Also verified `make_database("grampswebapidb")` loads the
  addon through Gramps's real plugin manager, not just via direct
  import.

### DIFF
--- a/GrampsWebApiDb/MANIFEST
+++ b/GrampsWebApiDb/MANIFEST
@@ -1,0 +1,1 @@
+GrampsWebApiDb/README.md
--- a/GrampsWebApiDb/README.md
+++ b/GrampsWebApiDb/README.md
@@ -1,0 +1,38 @@
+GrampsWebApiDb is a Gramps database backend that uses a Gramps Web API
+server (e.g. gramps-connect or Gramps Web) as a live database, mirrored
+locally in SQLite for speed. Reads are served from the local mirror, which
+is kept current via the server's transaction-history feed; local edits are
+pushed back to the server as they're committed.
+
+## Credentials
+
+The addon takes a single credential, via the `GRAMPS_WEB_API_KEY`
+environment variable, shaped `<REFRESH_TOKEN>*<BASE64URL(URL)>`. There is
+deliberately no login dialog and no per-tree settings.ini. Use
+`WebApiHandler.mint_api_key(url, username, password)` (see
+`webapi_client.py`) once to turn a username/password into this key.
+
+**Security tradeoff:** the token embedded in `GRAMPS_WEB_API_KEY` is a
+standard JWT *refresh* token obtained from the server's normal `/token/`
+login endpoint — the same endpoint and flow the official web client uses,
+not an undocumented or exploited access path. gramps-web-api leaves refresh
+tokens non-expiring by default, so this key is a long-lived, general-purpose
+credential carrying the full permissions of the account that minted it. It
+is *not* the same as a real scoped, independently revocable personal access
+token (gramps-web-api has that machinery, but it isn't generally wired into
+request auth yet). Practically, that means:
+
+* A leaked `GRAMPS_WEB_API_KEY` is as damaging as a leaked password — it
+  grants full account access until the underlying password is changed.
+  There is no "revoke this key" action independent of that.
+* Treat it accordingly: don't commit it, don't log it, and store it the
+  same way you'd store a password.
+
+This is a documented engineering tradeoff, made because the properly-scoped
+alternative isn't available server-side today — not a vulnerability in
+gramps-web-api or a loophole being exploited.
+
+## See also
+
+* `grampswebapidb.py` for the sync/write-through design (module docstring).
+* `webapi_client.py` for the token fetch/refresh implementation.
--- a/GrampsWebApiDb/README.md
+++ b/GrampsWebApiDb/README.md
@@ -8,9 +8,20 @@ pushed back to the server as they're committed.
 
 The addon takes a single credential, via the `GRAMPS_WEB_API_KEY`
 environment variable, shaped `<REFRESH_TOKEN>*<BASE64URL(URL)>`. There is
-deliberately no login dialog and no per-tree settings.ini. Use
+deliberately no login dialog and no per-tree settings.ini. Mint one once
+via username/password, either from the command line with the standalone
+`gramps-web-api-client` package (not yet published; pip-installable from
+its own repo, e.g. `pip install -e path/to/gramps-web-api-client`):
+
+```bash
+export GRAMPS_WEB_API_KEY=$(gramps-web-api-client generate-key --url https://your-server/api --username youruser)
+```
+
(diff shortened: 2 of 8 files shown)


dsblank:
@GaryGriffin this is ready to merge when you are.

I added help page here: https://gramps-project.org/wiki/index.php/Addon:GrampsWebApiDb

GaryGriffin:
This is currently `status=unstable` . But your discourse thread says BETA. If it is unstable, then user has to run gramps in debug mode to access it. Is that the intent? This is not mentioned in the Help page. 

I dont think debug mode is available for Mac, for instance. I have tested that it loads and passes unittests, but I have not tested if it communicates with a Web server. Please confirm you want `unstable` and I will publish it.

dsblank:
Sorry, forgot to push the last commit.

dsblank:
Should be ready now. :crossed_fingers:

PR #1013, opened 2026-08-11
JSON: use .jsonl extension for export/import

dsblank:
## Summary
- The JSON export writes one JSON object per line (JSON Lines format), via `write_line` appending `\n` after each `object_to_string(obj)` call — not a single JSON document.
- Updated `extension` in `JSON.gpr.py` from `json` to `jsonl` for both the EXPORT and IMPORT registrations so the produced/expected file extension matches the actual format.

## Test plan
- [x] Export a family tree via the JSON exporter in Gramps and confirm the file save dialog now defaults to `.jsonl`.
- [x] Re-import the resulting `.jsonl` file and confirm it round-trips correctly.

### DIFF
--- a/JSON/JSON.gpr.py
+++ b/JSON/JSON.gpr.py
@@ -10,7 +10,7 @@ register(
     export_function="exportData",
     export_options="WriterOptionBox",
     export_options_title=_("JSON options"),
-    extension="json",
+    extension="jsonl",
     help_url="Addon:JSON_Export_Import#Export_JSON",
 )
 
@@ -24,6 +24,6 @@ register(
     status=STABLE,
     fname="JSONImport.py",
     import_function="importData",
-    extension="json",
+    extension="jsonl",
     help_url="Addon:JSON_Export_Import#Import_JSON",
 )


GaryGriffin:
I exported my tree and then created a new tree. The Database Path and Backup Path were correctly set when I created the new Family Tree. The Base Media Path was blank. This is as expected.

When I imported the jsonl file, it did not set the Base Media Path as part of the import, so it could not find any of the media.  Importing a .gramps file correctly sets the Base Media Path. Based on the footer of the Settings window, shouldn't this be in the jsonl file?  

Screenshot 2026-08-11 at 3 51 33 PM

Of course, this was not something that was changed in this PR, as the PR only changed the file extension.

dsblank:
@GaryGriffin it is true that this import/export doesn't handle a full database and metadata---like a bunch of other exporters. In fact, if you use a filter to export, and import that you'll get a corrupted database.

This, like other import/export addons, is a pretty-raw database dump. We need to better describe this in the UI. But this is how it is, for now.

PR #1014, opened 2026-08-12
GrampsWebApiDb: sync media files, fix silent write-side drift

dsblank:
Four gaps found by comparing `GrampsWebApiDb` against **GrampsWebSync**, which solves the same sync problem interactively. Three of them let local changes silently never reach the server.

Each was verified by reverting the fix and confirming the targeted tests fail.

## 1. Media files never synced

`webapi_client.py` was trimmed down from GrampsWebSync's `WebApiHandler` and dropped its media-file-sync half. Media *records* (path, checksum, description) synced fine; the files they point at never moved in either direction.

Ports `get_missing_files()` / `download_media_file()` / `upload_media_file()` (credit @DavidMStraub, same repo and license), adapted to this class's bearer-header auth and retry conventions. Driven by a `_sync_media_files()` pass on its own coarser timer rather than GrampsWebSync's explicit wizard step — unlike the record feed there is no cheap "what changed since last time" signal for file presence, so each pass rescans.

## 2. Local batch commits were never pushed

DBAPI skips its per-object undo log for a `batch=True` transaction, so `transaction_to_json()` produced an empty payload and `_push_payload()` returned immediately — no push, no log line, no error. That covers **every bulk importer** and several stock Tools (Check and Repair Database, Media Manager, Reorder Gramps IDs, Sort Events, …). The mirror simply drifted ahead of the server, and nothing noticed: the poll only ever looks for newer *server-side* transactions, never "am I ahead?".

`transaction_begin()` now snapshots per-type handle sets, and `transaction_commit()` diffs them against post-commit state to reconstruct the change list (appeared → add, disappeared → delete, survived with `.change >= start_time` → update). Entries replay through the existing non-batch path, so each object picks up a real `"old"` snapshot and gets normal conflict handling. A `_pulling` flag keeps `_sync_from_server()`'s own batch replays from being echoed straight back at the server.

This is the write-side mirror of the `_full_resync()` fallback already in place for the equivalent pull-side blind spot.

## 3. Pushes that failed while offline were dropped

The local commit is never rolled back, but nothing remembered the push still needed to go out — the existing "until the next successful push or read sync" was aspirational, not implemented.

Failed pushes now persist via `_set_metadata()` (surviving close/reopen) and are retried at the top of every `_sync_from_server()`, **in order, stopping at the first still-undeliverable entry** rather than skipping ahead — so a later causally-dependent edit (a Family added after the Person it references) can't reach the server ahead of one still stuck. The queue is capped, and a 4xx other than 429 is treated as the server's settled answer and logged rather than queued; queueing one would retry forever and eventually evict genuinely retryable work.

## 4. Server permissions and version now checked at `load()`

`POST /transactions/` requires `AddObject` + `EditObject` + `DeleteObject` *together*, so a Member-level account could open the tree, read perfectly well, and then have every single edit rejected.

`ViewPrivate` matters more subtly. The history feed 403s loudly without it — but `GET /exporters/gramps/file` does **not** refuse the request; it passes `view_private=has_permissions({PERM_VIEW_PRIVATE})` into the export task and returns a privacy-filtered export, which `_full_resync()` would then import over a wiped local mirror, quietly dropping every private record locally.

Both are now named explicitly up front, at no extra round trip (permissions ride in the JWT claims). A tree opened read-only is held to the read permission only. `GET /metadata/` additionally gates on the server's Gramps version, turning an incompatible pre-6.0 server from a bare `KeyError` mid-sync into a sentence naming both versions.

## Also: background pushes for large payloads

Where the server supports it (gramps-web-api ≥ 2.7), a payload at or above a threshold goes out with `?background=1` so it isn't cut off mid-write by the request timeout — most of all the single large payload reconstructed after a bulk import (#2 above). Small pushes stay synchronous, keeping the crisp "400 means conflict" semantics.

Two server-side asymmetries are absorbed inside `push_transaction()` so callers see identical behavior either way:

- the server only *actually* backgrounds the work if it has a Celery queue configured — otherwise `run_task()` runs it inline and answers **200**, not 202;
- on that inline path a conflict comes back as HTTP **500** rather than 400, because `run_task()` catches `process_transactions()`'s `ValueError` and re-aborts it.

Both status codes and a failed background task are checked for the `"Object has changed"` sentinel. Without that, a conflict on those paths would read as a transient server error and — given the retry queue from #3 — be retried forever.

## Notes

- **Cost:** batch reconciliation is O(total objects) per batch commit (`.change` isn't a queryable column in DBAPI's schema), and the touched objects are written twice locally. Same order as whatever opened the batch transaction in the first place; documented inline.
- Background mode does **not** make the UI more responsive — polling blocks the GTK main thread just as a long POST would. It only prevents the connection dropping mid-write. Genuine async remains out of scope, as the module docstring already notes.
- No `version=` bumps in `.gpr.py` (left to the release build, per `CLAUDE.md`).

## Testing

121 new tests, 238 total, all passing:

```
GRAMPS_RESOURCES=/path/to/gramps GDK_BACKEND=- \
  python3 -m unittest GrampsWebApiDb.tests.test_grampswebapidb \
                      GrampsWebApiDb.tests.test_webapi_client
```

No functional change to the read/sync path against a correctly-permissioned, up-to-date server.

### DIFF
--- a/GrampsWebApiDb/grampswebapidb.py
+++ b/GrampsWebApiDb/grampswebapidb.py
@@ -141,6 +141,111 @@ user) is still out of scope. If the push fails for a non-conflict reason
 is not rolled back -- the local mirror just drifts from the server until
 the next successful push or read sync.
 
+Not every local batch=True commit is a pull-side replay, though: the same
+trans.batch guard that makes _sync_from_server()'s own replay silent to
+transaction_to_json() applies equally to *local* bulk operations run
+against this open tree from outside this file entirely -- ImportXml/
+ImportGedcom/ImportCsv/..., and stock Tools like Check and Repair
+Database, Media Manager, Extract Information from Names, Rename Event
+Types, Reorder Gramps IDs, and Sort Events all open their own DbTxn with
+batch=True for performance. Left alone, any of those would apply locally
+and never reach the server: transaction_commit() would see the same empty
+transaction_to_json() payload it correctly sees for _sync_from_server()'s
+own pull-side batch replay, with nothing in the payload itself to tell the
+two apart. transaction_begin() (called by DbTxn.__enter__, so before the
+batch operation's body runs) tells them apart with a _pulling flag set
+only around _sync_from_server()'s own batch DbTxns (including the ones
+_full_resync() opens) -- everywhere else, a batch=True transaction gets a
+snapshot of every primary object type's handle set stashed on the
+transaction itself (_handles_by_class()). transaction_commit() diffs that
+snapshot against the post-commit state (_reconcile_batch_commit()) to
+reconstruct what changed: a handle that appeared is an add, one that
+disappeared is a delete, and one that persisted but whose .change moved to
+at or after the transaction's own start_time was an update -- _commit_
+base() always stamps .change on every commit, batch or not; only the
+undo-log recording that batch skips is what normally would have let
+transaction_to_json() see this directly. The reconstructed entries are
+handed to _retry_after_conflict() -- not because anything conflicted, but
+because it already does exactly what's needed here: replay each entry as
+a fresh, ordinary (non-batch) local edit, so it picks up a real "old"
+snapshot from DBAPI's own commit path and goes out through the normal
+push path one object at a time, with the same conflict handling a live
+edit gets. This costs roughly double the local writes for whatever the
+batch operation actually touched (once for the batch commit, once more
+for this replay) -- accepted as the price of correctness, the same trade
+_full_resync() already makes for the equivalent pull-side blind spot.
+
+A second, previously-unhandled kind of silent drift: when a push's own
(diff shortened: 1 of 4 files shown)


GaryGriffin:
It might be useful (some time in the future) to have a demo server set up and instructions for setting up this gramplet to connect. Might lower the threshold for users to adopt this gramplet.

DavidMStraub:
> It might be useful (some time in the future) to have a demo server set up and instructions for setting up this gramplet to connect. Might lower the threshold for users to adopt this gramplet.

I don't see why it wouldn't work with https://demo.grampsweb.org/ - definitely worth trying.

dsblank:
> I don't see why it wouldn't work with https://demo.grampsweb.org/ - definitely worth trying.

Good idea!

One more PR for some edge cases in #1015 and you see:

image

I'll write this up in the discourse.

PR #1017, opened 2026-08-20
GrampsWebApiDb: sync hardening and read-only-account support

dsblank:
## Highlight

This continues hardening how the local mirror stays in sync with a Gramps Web API server, and adds support for accounts that don't have full edit rights on the server. An account limited to read-only (or partial) access can now open a Family Tree and keep it up to date, instead of being refused outright; any edit it isn't actually allowed to make is now rejected clearly and immediately, rather than appearing to save and then silently never reaching the server.

This continues an underlying philosophy for this addon: merging between the local copy and the server always happens automatically — there's no interactive "resolve this conflict" prompt to get in the way — and every conflict does get resolved one way or another. When that automatic resolution has to make a judgment call (or an edit couldn't be delivered to the server at all, as with the permission case above), a "message" note is attached to the affected record so a person can review it later, on their own schedule, rather than being interrupted in the moment.

## Details

**Read-only / permission-limited accounts**
- Opening a Family Tree no longer requires full edit rights on the server. An account that can only read is still checked for the one permission reading truly depends on, but is otherwise let in — the mirror opens, and keeps polling and staying current.
- If that account then tries to make an edit (add, change, delete, undo, redo) it isn't allowed to make, the edit is refused right away, with a message naming what's missing, instead of quietly succeeding locally and only failing (silently, later) when it tries to reach the server.
- A permission that changes *after* a tree is already open still surfaces the old way (a rejected sync attempt with a note left on the affected record) — this only changes the case where the limitation is already known when the tree is opened.

**Prior review-pass fixes also included in this branch**
- More accurate automatic conflict resolution when the same field is edited both locally and on the server, with a note left behind explaining what was kept versus discarded whenever that happens.
- That same note-leaving behavior now also covers an edit that permanently fails to reach the server for any other reason.
- A couple of mirror-resync edge cases (birth/death event ordering, and detecting a stalled poll) are more robust.

## How to test

For the permission-related change, `https://demo.grampsweb.org` is a convenient place to test since it's a public server you likely already have (or can create) an account on with a role short of full Editor access. Mint an API key from an account like that (this addon's own "Mint API Key" tool works fine for this), point `GRAMPS_WEB_API_KEY` at it, and open a tree:
- Confirm the tree opens normally and stays in sync with the server (make a change from another account/browser and confirm it shows up here).
- Try making a local edit (e.g. add a note or change a field) and confirm you get a clear rejection message instead of the edit silently disappearing or the app behaving oddly.
- As a sanity check, repeat with a full-Editor-role account and confirm normal editing still works exactly as before.

### DIFF
--- a/GrampsWebApiDb/grampswebapidb.py
+++ b/GrampsWebApiDb/grampswebapidb.py
@@ -794,6 +794,87 @@ _WRITE_PERMISSIONS = ("AddObject", "EditObject", "DeleteObject")
 _REQUIRED_ROLE_NAME = "Editor"
 
 
+class _MissingWritePermissionError(DbWriteFailure):
+    """DbWriteFailure raised specifically by _missing_write_permission_
+    error() -- a distinct subclass purely so _install_quiet_popup_
+    filter()'s logging.Filter can recognize this one, already-explained,
+    expected condition and single it out from any other DbWriteFailure
+    (e.g. exportxml.py's backup-write failures), which must keep
+    reaching Gramps' generic crash dialog unchanged.
+    """
+
+
+def _notify_missing_write_permission(message):
+    """Show a calm, native dialog carrying the same explanation as the
+    DbWriteFailure _missing_write_permission_error() is about to raise.
+
+    Neither a friendlier exception message nor a custom dialog title can
+    change what Gramps' own top-level crash dialog says -- ErrorView
+    (gui/logger/_errorview.py) hardcodes "Gramps has experienced an
+    unexpected error" and "it would be advisable to restart Gramps
+    immediately" regardless of exception type or message, which is
+    needlessly alarming for something as ordinary and recoverable as a
+    read-only account attempting an edit. This dialog is the actually
+    useful one; _install_quiet_popup_filter() (below) keeps the alarming
+    one from also popping up right after it. has_display()-gated the
+    same best-effort way _notify_fatal_poll_error() already is, since
+    this DATABASE plugin must stay importable and usable without a
+    display (CLI use) or gi/Gtk installed at all -- headless just skips
+    the dialog, same as _notify_fatal_poll_error().
+    """
+    if not has_display():
+        return
+    try:
+        from gramps.gui.user import User as GuiUser
+    except ImportError:
+        return
+    GuiUser().notify_error(_("Can't save this change"), message)
+
+
+def _install_quiet_popup_filter():
+    """Add a logging.Filter to the root logger's GtkHandler, if one is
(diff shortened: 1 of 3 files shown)


GaryGriffin:
Logged in to grampsweb demo as member. When I tried to add an attribute, gramps generated warning (expected) and error (unexpected). This is intimidating to the user (Error telling them to restart gramps) and unclear whether they can continue or not. 

Not sure there is any way more graceful to address - trying to write to a read-only.

```
2026-08-23 09:42:19.217: WARNING: grampswebapidb.py: line 2259: The account authenticating via GRAMPS_WEB_API_KEY is missing server permission(s) ("AddObject, EditObject, DeleteObject") this addon needs to push local changes back to the server; grant it the "Editor" role (or ask an administrator to) to enable editing. Opening this Family Tree read-only against the server: the local mirror will keep polling and staying current, but any local edit will be rejected when it tries to push.
2026-08-23 09:42:53.247: ERROR: grampsapp.py: line 187: Unhandled exception
  File "/Volumes/Storage/Applications/Gramps6.0.6.app/Contents/Resources/lib/python3.13/site-packages/gramps/gui/editors/editperson.py", line 977, in save
    with DbTxn(
         ~~~~~^
        _("Edit Person (%s)") % self.name_displayer.display(self.obj),
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        self.db,
        ^^^^^^^^
    ) as trans:
    ^
  File "/Volumes/Storage/Applications/Gramps6.0.6.app/Contents/Resources/lib/python3.13/site-packages/gramps/gen/db/txn.py", line 82, in __exit__
    self.db.transaction_commit(self)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/Users/gary/Library/Application Support/gramps/gramps60/plugins/GrampsWebApiDb/grampswebapidb.py", line 2783, in transaction_commit
    raise _missing_write_permission_error(self._missing_write_permissions)
gramps.gen.db.exceptions.DbWriteFailure: This local edit was not saved: the account authenticating via GRAMPS_WEB_API_KEY is missing server permission(s) needed to push changes to this Family Tree: AddObject, EditObject, DeleteObject. Grant it the "Editor" role on the server (or ask an administrator to) to enable editing.

```

dsblank:
Thanks for that report, and suggestion. I'll catch that error add a nice message. Will finish that tomorrow.

dsblank:
Updated to give a nice, message about the read-only status of the db. Added additional tests.

PR #1021, opened 2026-08-30
Gramps Web Sync: trust private CAs from all macOS keychains

my-flow:
## Problem

On macOS, Gramps Web Sync cannot connect to a Gramps Web instance served with a
privately issued certificate — one signed by a company root pushed via MDM, or by a
local CA generated for a self-hosted instance. The sync fails at connect time with:

```
URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED]
certificate verify failed: unable to get local issuer certificate>
```

The same URL loads without complaint in Safari and `curl`, and the same username and
password work in the web UI. That mismatch makes the failure easy to misread as a
credentials or server problem.

## Cause

`create_macos_ssl_context()` harvests certificates from exactly one keychain:

```python
"/System/Library/Keychains/SystemRootCertificates.keychain"
```

That is Apple's **read-only** list of public root CAs. A privately issued CA is never
installed there — it goes into the admin keychain (`/Library/Keychains/System.keychain`)
or the user's login keychain. Safari and `curl` consult the full keychain search list,
which is why they succeed where the addon fails.

There is no fallback either. The context is seeded from `ssl.create_default_context()`,
which normally loads OpenSSL's default CA store — but the OpenSSL shipped inside the
Gramps macOS bundle has its `OPENSSLDIR` baked in at build time:

```
$ strings Gramps.app/Contents/Resources/lib/libcrypto.3.dylib | grep etc/ssl
/Users/john/Development/gramps-tarball-11.0-arm64/inst/etc/ssl
```

That path exists only on the build machine, so on a user's Mac the default store
resolves to nothing and the context starts out **empty**. Every trust anchor has to be
supplied explicitly from the keychains — which makes the single-keychain read the sole
source of trust, not merely a supplement to it.

## Note for reviewers

The change reads the admin keychain and the `security list-keychains` search list in
addition to Apple's roots. That promotes every certificate in those keychains to a
trust anchor, without consulting per-certificate **trust settings** — so a certificate
a user has stored but explicitly marked *untrusted* would still be honoured. That
limitation is inherited from the existing implementation, not introduced here, and it
keeps the change small and dependency-free.

Reading trust settings properly means going through the Security framework rather than
parsing `security(1)` output — [`truststore`](https://pypi.org/project/truststore/)
does exactly that and would let this whole function be deleted. I've kept that out of
this PR since adding a runtime dependency to an addon is a separate decision, but I'm
happy to follow up if maintainers would prefer that direction.

No `version` bump in `.gpr.py`, per `AGENTS.md`.

### DIFF
--- a/GrampsWebSync/webapihandler.py
+++ b/GrampsWebSync/webapihandler.py
@@ -126,13 +126,9 @@ def _macos_keychains() -> list[str]:
 def create_macos_ssl_context() -> ssl.SSLContext:
     """Create an SSL context trusting the CAs in the user's macOS keychains.
 
-    A privately issued CA — a corporate root, or one made for a self-hosted
-    Gramps Web instance — lands in the admin or login keychain, never in the
-    read-only ``SystemRootCertificates`` list. Reading that list alone fails
-    with ``CERTIFICATE_VERIFY_FAILED`` on a server Safari opens happily.
-
-    There is no fallback: the bundled OpenSSL's ``OPENSSLDIR`` points into the
-    build machine's tree, so the default store is empty on a user's Mac.
+    Searches the hard-coded system keychains and the user keychains reported
+    by ``security list-keychains``, so that a privately issued CA installed in
+    the admin or login keychain is trusted alongside Apple's public roots.
     """
     ctx = ssl.create_default_context()
     pem_blocks: list[bytes] = []
--- a/GrampsWebSync/webapihandler.py
+++ b/GrampsWebSync/webapihandler.py
@@ -28,6 +28,8 @@ import logging
 import os
 import platform
 import socket
+import ssl
+import subprocess
 import time
 from collections.abc import Callable
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


GrampsWebSync/webapihandler.py
```diff
@@ -91,25 +93,75 @@ def parse_version(version) -> tuple[int, int]:
...
+    return keychains
+
+
+def create_macos_ssl_context() -> ssl.SSLContext:
+    """Create an SSL context trusting the CAs in the user's macOS keychains.
+
+    A privately issued CA — a corporate root, or one made for a self-hosted
+    Gramps Web instance — lands in the admin or login keychain, never in the
+    read-only ``SystemRootCertificates`` list. Reading that list alone fails
+    with ``CERTIFICATE_VERIFY_FAILED`` on a server Safari opens happily.
```
jralls: History lessons belong in the commit message--and it's already there. The docstring should say what the current function does: Search the hard-coded system keychains and the user keychains obtained from `security list-keychains`. 

Note that the next bit, about OPENSSLDIR, isn't quite right either. The compiled in value points into the build machine (and I don't see a way to change that in the Open SSL configure script) but OpenSSL looks for it in the environment, so we could set it in gramps_launcher.py to point somewhere else. I think we should,  using `security list-keychains` is better.

jralls (changes requested):
The code looks fine, just the one docstring that will be anachronistic as soon as this gets merged.

dsblank:
If there is an issue here, there may also be one in GrampsWebApiDb as well:

https://github.com/gramps-project/addons-source/tree/maintenance/gramps60/GrampsWebApiDb

FYI, gramps-web-api hasn't moved to gramps61 yet.

DavidMStraub:
I don't own any :green_apple: device and I also can't judge if all this complication is really needed (or whether the addon itself is the right place to fix it).

Just want to share for background, since it's not accessible in this repo's git blame, where the original function was introduced:

Here https://github.com/DavidMStraub/gramps-web-sync/pull/27.

Original issue discussion: https://github.com/DavidMStraub/gramps-web-sync/issues/26

@andreaiorio would you be able to comment on this PR?

GrampsWebSync/webapihandler.py
GaryGriffin: @jralls are you saying that a better solution would be to set OPENSSLDIR ( based on `security list-keychains` ) in the gramps_launcher.py , rather than the changes in this PR? 

Or are you saying that this PR is ok to merge from your perspective?

EDIT: Or is this a documentation issue? If there is a private CA, then it needs to be in the OPENSSLDIR ?

GaryGriffin:
Should this be on gramps60 branch instead of gramps61? All changes on 60 are cherry-picked to 61. Changes on 61 are not back-ported to 60.

GrampsWebSync/webapihandler.py
jralls: @my-flow Sorry, that last sentence should have been "I **don't** think we should, using `security list-keychains` is better.".

jralls (approved):

andreaiorio:
Thanks for tagging me @DavidMStraub 
I don't have access to a Mac environment to test it locally right now, but the logic and explanation makes sense to me.

GaryGriffin:
@DavidMStraub are you approving merging this in 60 and cherry picking to 61?

DavidMStraub:
I can't test it, but it looks reasonable and the tests look meaningful. AI didn't flag anything either. So, no blocker from my side.

PR #1023, opened 2026-09-03
GOQLFilter: update vendored gramps-object-query-language to 0.5.0 (backlinks)

dsblank:
## Summary
- Refreshes the gramps60 branch's vendored copy of `gramps-object-query-language` (in `GOQLFilter/gramps_object_query_language_copy/`) from 0.4.1 (`b717eab5`) to 0.5.0 (`ffdcde3a`)
- Picks up backlinks support: `exists(backlinks, ...)` and `count(backlinks, ...)` let a `where_expr` query the referrer side of Gramps' own `reference` table (who points at this object), filtered by the referrer's class via `_class ==`/`!=`/`in`, without needing a forward relationship defined on the referencing type
- `query.py`, `query_lang.py`, and `evaluator.py` are copied verbatim from upstream per the vendoring notice in `__init__.py`; only the pinned version/commit in that notice was hand-edited

## Test plan
- [x] `python3 -m unittest GOQLFilter.tests.test_integration_whereexprrule -v` — 13 tests, all pass
- [x] Full `GOQLFilter/tests` suite run via `unittest discover` — one unrelated pre-existing failure (`test_goql.py` import errors on a GTK 3.0/4.0 namespace conflict when all test modules load in one process), not caused by this change

https://claude.ai/code/session_01FCcftyvqz1gSDMKQFHQVAW

### DIFF
--- a/GOQLFilter/gramps_object_query_language_copy/__init__.py
+++ b/GOQLFilter/gramps_object_query_language_copy/__init__.py
@@ -29,7 +29,7 @@ handle.
 VENDORED COPY -- do not edit directly. This is a verbatim copy of the
 "gramps_object_query_language" package (PyPI: gramps-object-query-language,
 https://github.com/dsblank/gramps-object-query-language), pinned at
-version 0.4.1 / commit b717eab575e99d8dc1dad958bf44442fe64f4276, imported
+version 0.5.0 / commit ffdcde3a97e3a6dc194779a14dd2d234d113f4c1, imported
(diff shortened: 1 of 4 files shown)


dsblank:
@GaryGriffin  PR contains an updated copy of the gramps-object-query-language. It is internal to the GOQLFilters, allowing users to write queries like:

```
not exists(backlinks)
```

which will find all of the objects in the view you are in that don't have any links to it. This, like all of the other queryable items, is done in SQL for speed.

PR #1024, opened 2026-09-05
SharedPostgreSQL: ANALYZE tables touched by a batch commit

dsblank:
## Summary

- Fixes stale query-planner statistics after importing a new tree into a `SharedPostgreSQL` install that already has other trees' data (https://github.com/gramps-project/gramps-web-api/discussions/969).
- `SharedDBAPI.transaction_commit()` now runs `ANALYZE` on the tables touched by a *batch* transaction right after it commits. `batch=True` is what Gramps' importers (GEDCOM, CSV, vCard, XML, Pro-Gen) and bulk tools (Check and Repair, Reorder IDs, etc.) already use, so this only fires for the operations that can create the row-count skew described below — ordinary single-record edits are untouched.

## Why

`SharedPostgreSQL` pools every tree's rows into shared tables (`person`, `event`, `citation`, ...) distinguished by a `treeid` column, with every index led by `(treeid, ...)`. PostgreSQL's autovacuum only re-`ANALYZE`s a table once its changed-row count crosses `autovacuum_analyze_threshold + autovacuum_analyze_scale_factor × (rows already in the table)` — 50 + 10% by default, sized against the *whole* table. On a fresh, single-tree database, an import is most of the table, so this threshold trips almost immediately. On a `SharedPostgreSQL` install that already has meaningful data across multiple trees, a new tree's import can easily be a small fraction of the pooled table — well under 10% — so autoanalyze may never fire for it. The planner then keeps using statistics that predate the new tree's rows indefinitely, and can pick a badly wrong plan (e.g. a secondary index instead of the primary key) for that tree specifically.

## Verification

Reproduced against a real PostgreSQL 16 instance: seeded a `person` table shaped like `SharedPostgreSQL`'s actual schema (`PRIMARY KEY (treeid, handle)`, secondary indexes led by `treeid`) with 200,000 rows for "tree 1", then inserted 500 rows for a new "tree 2" without running `ANALYZE`. A plain `treeid + handle` lookup used the `person_surname` secondary index (502 buffer touches, 499 rows discarded by a filter) instead of the primary key (4 buffer touches) — reproducing the reported symptom. The defaults confirm it structurally: `autovacuum_analyze_threshold=50` + `autovacuum_analyze_scale_factor=0.1` means autovacuum needs ~20,050 modified rows on that table before it re-analyzes — a 500-row single-tree import never gets close.

Then called the real (unmocked) `transaction_commit()` for a batch transaction against the same database: with no manual `ANALYZE`, the plan flipped from the bad one above to the primary key.

## Test plan

- [x] New unit tests in `SharedPostgreSQL/tests/test_transaction_analyze.py` covering `_get_txn_tables()`, `_analyze_tables()`, and that `transaction_commit()` only analyzes for batch transactions (8 tests)
- [x] Existing `SharedPostgreSQL/tests/` suite still passes (58/58 total)
- [x] End-to-end manual verification against a real PostgreSQL 16 instance (described above)

https://claude.ai/code/session_01JqixScuMyvWXBcARLnY3Wv

### DIFF
--- a/SharedPostgreSQL/shareddbapi.py
+++ b/SharedPostgreSQL/shareddbapi.py
@@ -387,6 +387,8 @@ class SharedDBAPI(DbGeneric):
             # FIXME: need a User GUI update callback here:
             self.reindex_reference_map(lambda percent: percent)
         self.dbapi.commit()
+        if txn.batch:
+            self._analyze_tables(self._get_txn_tables(txn))
         if not txn.batch:
             # Now, emit signals:
             # do deletes and adds first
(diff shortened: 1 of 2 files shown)


dsblank:
CC: @DavidMStraub

DavidMStraub:
I would have liked to review this before  merging.

GaryGriffin:
Sorry for merging before your review. My mistake.

dsblank:
I'll create a new PR with fixes.

PR #1025, opened 2026-09-07
Revert: SharedPostgreSQL ANALYZE-on-batch-commit (no-op fix)

dsblank:
## Summary

Reverts #1024 ("SharedPostgreSQL: ANALYZE tables touched by a batch commit", commit f704d5fe8).

The diagnosis behind that PR is correct — on a shared database, autovacuum's analyze threshold is sized against the whole table across all trees, so a batch import for one tree can leave the planner using stale statistics and pick a badly wrong plan (e.g. a secondary index instead of the primary key) for that tree's rows. But the fix as merged does not run:

- Every `trans.add()` / `transaction.add()` call site in `shareddbapi.py` is guarded by `if not trans.batch:` (`_commit_base`, `_update_backlinks`, `_do_remove`, `_remove_backlinks`). A batch transaction's `txn` dict is therefore *never populated* — by construction it stays an empty `defaultdict` for the life of the transaction.
- The merged code added `if txn.batch: self._analyze_tables(self._get_txn_tables(txn))`, but `_get_txn_tables(txn)` iterates that same, necessarily-empty `txn`, so it always returns `set()` and no `ANALYZE` is ever issued.

Verified against a real PostgreSQL 16 instance: a 500-row batch import leaves `len(txn) == 0` and the analyze count for the touched tables unchanged (0), while a non-batch transaction of a single row does populate `txn` and does trigger an `ANALYZE` through the same code path (positive control: calling `_analyze_tables` directly does work — that's what the commit's before/after numbers actually measured, not the batch-commit trigger).

The four new tests in `test_transaction_analyze.py` pass because `_FakeTxn` is built as a `defaultdict` populated the way a *non-batch* commit populates one, then labelled `batch=True` — a combination the real commit path can never produce. So the tests and the (non-functional) code agree with each other without exercising the actual trigger.

## Why revert rather than fix forward

Per-commit `ANALYZE` on a batch transaction, even made to actually fire, only helps once a tree's row count is large enough to enter the planner's most-common-values list (roughly `1/default_statistics_target` of the table's distinct `treeid`s) — smaller or skewed trees stay below that and get the wrong plan regardless of how often `ANALYZE` runs. A schema-time fix such as

```sql
ALTER TABLE person SET (autovacuum_analyze_scale_factor = 0,
                        autovacuum_analyze_threshold = 20000);
```

on the affected shared tables hands this to autovacuum — background, cost-throttled, and independent of whether a given commit happens to be batched — without depending on the currently-empty batch `txn` dict at all. That's a separate, follow-up change; this PR only removes the non-functional one so it isn't shipped as if it were working.

## Test plan

- [ ] `python3 -m unittest SharedPostgreSQL.tests -v` (the removed test file is gone; no other SharedPostgreSQL tests reference `_analyze_tables`/`_get_txn_tables`)
- [ ] Confirm `shareddbapi.py` no longer references `_analyze_tables` / `_get_txn_tables`

https://claude.ai/code/session_016WGcS1xRMVRw9c5NXPThQ3


dsblank:
@GaryGriffin @DavidMStraub we can merge this that reverts #1024. I'll have a follow-up that we can further review and discuss.

GaryGriffin:
@DavidMStraub Are you okay for me to merge the revert?

DavidMStraub:
Yes, thank you both!

PR #1026, opened 2026-09-07
SharedPostgreSQL: fix analyze threshold for shared tables at schema creation

dsblank:
## Summary

Follow-up to #1025 (revert of #1024). #1024's diagnosis was correct — on a shared database, every tree's rows live in the same tables, so autovacuum's analyze threshold (a fixed count plus a fraction of the *whole* table's row count) is sized against all trees combined. A batch import for one tree can leave the planner using statistics that predate those rows and pick a badly wrong plan (e.g. a secondary index instead of the primary key) for that tree. Its fix didn't work because it tried to trigger an explicit `ANALYZE` from inside a batch `transaction_commit()`, but every `trans.add()` call site is guarded by `if not trans.batch:`, so a batch transaction's `txn` dict is never populated for it to read from (see #1025 for the full analysis).

This PR takes the declarative route instead: set a fixed, table-size-independent analyze threshold on each per-tree table at schema-creation time, and let autovacuum do the work — background, cost-throttled, and triggered by row-count alone, independent of whether the commit that pushed a table over the threshold happened to be batched.

```python
for table in (person, family, source, citation, event, media,
              place, repository, note, tag, reference,
              name_group, metadata, gender_stats):
    ALTER TABLE {table} SET (autovacuum_analyze_scale_factor = 0,
                             autovacuum_analyze_threshold = 1000)
```

## Verification

Reproduced the original report's scenario against a real PostgreSQL 16 instance (200,000 pre-existing rows, then a 500-row import for a new tree):

- **Before**: effective analyze threshold is `50 + 0.1 * 200,000 = 20,050`. A `treeid + handle` lookup for the new tree used a secondary index (`person_given_name`) and touched 502-503 buffers. `n_mod_since_analyze` stayed at 500 and no autoanalyze ran even after 30s with an accelerated (5s) `autovacuum_naptime`.

Also confirmed all 14 `ALTER TABLE ... SET (...)` statements execute without error against the addon's actual generated schema (extracted and ran the real `_create_schema()` SQL, table names included, against a scratch database).

**Update (per @DavidMStraub's review):** at the originally-proposed `autovacuum_analyze_threshold = 5000`, a 500-row import for a new tree did *not* trigger an autoanalyze after 6 naptimes (`n_mod_since_analyze=500`, lookup still 502 buffers) — the earlier "after" claim above for that exact scenario didn't hold up under a second, independent reproduction. A larger, 6,000-row import did trigger autoanalyze within 30s and dropped the lookup to 4 buffers, confirming the mechanism works once the threshold is actually crossed. The threshold has been lowered to `1000` to make autovacuum responsive to smaller, more typical per-tree batches; it does not by itself fix an import smaller than the threshold, which still waits on other combined-table activity to cross it — same as before this change, just against a much lower bar than the old scale-factor-based threshold.

## Performance impact on imports / bulk operations

Checked whether this can slow down an import or a bulk tool like "Check and Repair" — it can't, and not just in theory:

- The `ALTER TABLE` statements run exactly once, when a brand-new shared database's schema is first created — never during an import.
- `ANALYZE`'s `ShareUpdateExclusiveLock` doesn't conflict with ordinary `INSERT`/`UPDATE`/`DELETE`, so autovacuum's background worker can never block an import's writes.
- A Gramps import is normally one single transaction, committed once at the end. Autovacuum can only see committed rows, so it's structurally unable to analyze any of the newly-imported rows until *after* the import has already finished and committed.

Net effect: no slowdown of the import's or bulk operation's own duration or latency, only some small, bounded, asynchronous background `ANALYZE` work on PostgreSQL's own schedule.

*(An earlier version of this section described a theoretical "ANALYZE mid-import" cost from a long-held open transaction. @DavidMStraub's review re-tested that scenario — 6,000 rows held open for 51s, confirmed open via `pg_stat_activity` — and `n_mod_since_analyze` stayed at 0 with autoanalyze only firing after commit, as expected from PostgreSQL's transactional statistics. That paragraph didn't reproduce and has been removed.)*

## Known limitations (carried over from the #1024/#1025 discussion)

- This does not fix every case in the earlier review — a tree that is too small or too skewed relative to others can still fail to enter the planner's most-common-values list, since that's a sampling limit of `ANALYZE` itself (bounded by `default_statistics_target`), not a timing issue this threshold change can address.
- Because this only runs in `_create_schema()`, it only takes effect for newly-created shared databases. An existing deployment would need the same 14 `ALTER TABLE` statements run by hand against its database; this addon has no schema-migration mechanism to hook a retroactive fix into, and adding one felt out of scope for this change. The `ALTER TABLE` statements are now documented next to the `SHARED_TABLE_ANALYZE_*` constants in `shareddbapi.py` for operators who need to apply them by hand.
- These are per-table storage parameters, so once set they take precedence over any `autovacuum_analyze_scale_factor` / `autovacuum_analyze_threshold` configured server-wide in `postgresql.conf` for these 14 tables — an operator tuning autovacuum globally won't see it affect these tables unless they also change (or `RESET`) the table-level settings.

## Test plan

- [x] `python3 -m unittest SharedPostgreSQL.tests -v` — includes `test_create_schema_analyze.py`, which asserts `_create_schema()` issues the expected `ALTER TABLE` for every per-tree table using the module's `SHARED_TABLE_ANALYZE_*` constants. The expected table set is now derived from the `CREATE TABLE` statements `_create_schema()` actually executes (minus `trees`), rather than a hardcoded copy of the table list, so a future table added without the setting fails the test. Whether autovacuum actually acts on the storage parameter is server behavior, not addon logic, and was verified manually above rather than asserted in a unit test.
- [x] `black --check SharedPostgreSQL/shareddbapi.py SharedPostgreSQL/tests/test_create_schema_analyze.py`

https://claude.ai/code/session_016WGcS1xRMVRw9c5NXPThQ3


dsblank:
CC: @GaryGriffin and @DavidMStraub. I think this is as much as we can do in the addon without a major refactor.

DavidMStraub:
After some back-and-forth discussion with Claude Opus, here is the summary. I'll leave the final decision up to you @dsblank, I don't have a strong opinion.

---

*Review written by Claude (Anthropic's AI model). Checked against a local PostgreSQL 16 instance, with the schema created by the PR's own `_create_schema()`.*

## What the PR gets right

- **Sound direction.** Moving from an imperative ANALYZE in `transaction_commit()` to autovacuum storage parameters hands the work to PostgreSQL itself: background, cost-throttled, and independent of whether a commit was batched.
- **Correct implementation.** The branch merges cleanly into `maintenance/gramps60` without reintroducing any of #1024, the new tests pass, and all 14 per-tree tables receive `autovacuum_analyze_scale_factor=0, autovacuum_analyze_threshold=5000`, with `trees` correctly excluded.
- **Accurate reasoning on locking.** ANALYZE's `ShareUpdateExclusiveLock` does not conflict with DML, and because an import is a single transaction, autovacuum only acts on its rows after commit.
- **Clear limitations.** The PR states plainly that MCV sampling limits remain and that existing databases are unaffected. The claim that the addon has no migration hook of its own also checks out: the stored schema version is Gramps core's.
- **Well-scoped tests.** They assert only what a unit test can, and say so explicitly.

## Why this may not be the right design

The *need* for this setting follows from the addon's shared-table layout, but the *value* is deployment tuning, which fits poorly in addon code:

- **The right threshold is workload-dependent.** It depends on tree sizes, tree count and write rate, so no single constant suits both a small installation and a large multi-tenant one. The chosen value illustrates this: with a threshold of 5000, the PR's own scenario — a 500-row import for a new tree — does not trigger an autoanalyze (see below).
- **Per-table settings silently override server configuration.** Table reloptions take precedence over `postgresql.conf`, so an operator tuning autovacuum globally sees no effect on these 14 tables, with nothing to indicate why.
- **Nobody fully owns the value.** It is applied once at creation, cannot be configured, and changing the constant later has no effect on existing databases. Operators end up managing it by hand regardless, and code and database can drift apart.

An alternative is for the addon to *document* the shared-table statistics caveat, the recommended settings, and a query to check exposure, and leave applying and tuning them to operators — either globally in `postgresql.conf` or per table.

## If it is merged, suggested changes

1. **Revisit the threshold or the verification text.** Measured with the shipped values:

   ```
   500-row import, after 6 naptimes:   n_mod_since_analyze=500, no autoanalyze, lookup 502 buffers
   6000-row import, after 30 s:        autoanalyze ran, lookup 4 buffers
   ```

   Either lower the threshold (e.g. 1000; each ANALYZE is bounded at 30,000 sampled blocks, so cost stays modest) or correct the description, which reports the 500-row case as fixed.

2. **Drop the "ANALYZE mid-import" paragraph.** It did not reproduce: with 6,000 inserted rows held in an open transaction for 51 s (confirmed open via `pg_stat_activity`), `n_mod_since_analyze` stayed at 0 and autoanalyze ran only after commit, as expected from PostgreSQL's transactional statistics.

3. **Document the settings where operators will find them** — in the addon's documentation or next to the loop, not only in the PR description. State that they are defaults, that they take precedence over server-level autovacuum settings, and give the `ALTER TABLE` statements for existing databases.

4. **Derive the expected tables in the test** from the `CREATE TABLE` statements actually executed (minus `trees`) rather than a hardcoded copy of the list, so a future table added without the setting makes the test fail.

5. **Nit:** the test imports `SharedPostgreSQL.shareddbapi` a second time for the constants, despite the comment above it saying it avoids a second import. Harmless, but inconsistent.

dsblank:
Thanks for the thorough re-testing — much appreciated. Addressed the "if merged" suggestions:

1. **Threshold**: lowered `SHARED_TABLE_ANALYZE_THRESHOLD` from 5000 to 1000, since your reproduction showed 5000 didn't trigger autoanalyze for a 500-row import. Also corrected the PR description's "after" claim for that exact scenario, which didn't hold up under your independent test.
2. **Dropped the "ANALYZE mid-import" paragraph** from the PR description — your re-test (6,000 rows held open 51s, confirmed via `pg_stat_activity`) showed `n_mod_since_analyze` stays at 0 until commit, so that concern doesn't apply.
3. **Documented the settings for operators**, next to the `SHARED_TABLE_ANALYZE_*` constants in `shareddbapi.py`: that they're per-table storage parameters which take precedence over `postgresql.conf`'s server-wide autovacuum settings, plus the `ALTER TABLE` statements needed to retrofit an existing deployment.
4. **Test now derives the expected per-tree table set from a single source of truth** rather than a hardcoded copy — went a bit further than "parse the executed CREATE TABLE statements": `_create_schema()` now uses one dict (`per_tree_table_schema`, table name → column definitions) to drive both the CREATE TABLE loop and the ALTER TABLE loop, and exposes the result as `self._per_tree_tables`. So the 14 table names now appear in exactly one place in the source, and the test reads that list directly instead of reverse-engineering it from the SQL sent to the mock. Verified the generated SQL is byte-for-byte unchanged from before this refactor.
5. **Fixed the reimport nit** — reworded the comment above the `SharedDBAPI` MRO lookup so it no longer claims to avoid "a second import" right next to a second import.

All 52 `SharedPostgreSQL` tests pass, `black --check` is clean. PR description updated to match.

dsblank:
@GaryGriffin this is now ready for merge/deploy.

GaryGriffin:
Do you also want this cherry picked to 61?

dsblank:
> Do you also want this cherry picked to 61?

No, because gramps61 has some changes that we'll need to address (FamilySearch). We'll need to make a specific one for that.

PR #1029, opened 2026-09-13
Context View

ajeb78:
Add Context View, a family-context chart.

The Context View shows the selected person, their parents, their spouse, the children of the family as well as the siblings of both them and their spouse. Birth and death dates, as well as the marriage date and place, are shown if known.

This is helpful for visualizing cousin / aunt / uncle relationships: clicking on any person visible in the Context View redisplays it centered on them, so it can be used to navigate around the tree.

If the selected person had multiple spouses it is possible to switch between the different families with buttons.

The path to the tree's home person is shown in bold, or in a label describing which family needs selecting. (If the home person is visible, no bold or label is shown.)

The right-click context menu works on any person, and primary gallery photos can be shown.

### DIFF
--- a/ContextView/ContextView.gpr.py
+++ b/ContextView/ContextView.gpr.py
@@ -1,0 +1,51 @@
+# encoding:utf-8
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2026 Adrian Baugh
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <https://www.gnu.org/licenses/>.
+#
+# `_` is injected into this file's namespace by the plugin registrar, and is
+# already the addon's own translator; it must not be defined here.
+#
+from gramps.gen.plug._pluginreg import *
+
+MODULE_VERSION = "6.0"
+
+register(
+    VIEW,
+    id="contextview",
+    name=_("Context"),
+    description=_(
+        "A chart placing a person in the context of their immediate family: "
+        "the couple in the centre, each partner's parents above them, their "
+        "children below, and each partner's siblings in a column to their "
+        "own side, eldest at the top. Click any box to re-centre the chart "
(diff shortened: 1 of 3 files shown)


GaryGriffin:
Plz edit the gpr.py file and remove the following lines :

`from gramps.gen.plug._pluginreg import *`
Reason: This import is done already for GPR file

`order=END,`
Reason: Unknown command

ajeb78:
Lines removed as requested. Thanks for the feedback :)

GaryGriffin:
1. The Help URL is listed in the GPR but it does not exist. You need to create this file at https://gramps-project.org/wiki/index.php?title=Addon:Context_View

2. No marriage info on ContextView. The other charts contain this.

3. Other charts do not print the death symbol unless the person is actually dead. Compare GraphView and FamilyTreeView presentation of example tree with ContextView. I thought the people were dead but an unknown date.

Screenshot 2026-09-21 at 2 09 30 PM
Screenshot 2026-09-21 at 2 15 24 PM

Screenshot 2026-09-21 at 2 08 46 PM

ajeb78:
1. I requested a gramps wiki account on Sep 13th but hadn't received the login last time I checked. Please could you check if it's stuck in a queue? I have a wiki page ready to create once the account is made.

2. Could I push back on this item of feedback? The marriage info on this view (at least the date) is really useful especially when dealing with a person with multiple families. The fact that it's available in other views doesn't help me if I'm looking at this view. Having the marriage date visible adds context to the parents' birth and death dates as well as those of the children.

3. OK. What is the convention if someone has no death info but was born longer ago than the "maximum age probably alive" preference? In that situation should an undated skull be shown, or still nothing? Once you let me know I'll change this to align with the answer to this.

GaryGriffin:
> I requested a gramps wiki account on Sep 13th but hadn't received the login last time I checked. Please could you check if it's stuck in a queue? I have a wiki page ready to create once the account is made.

@sam-m888 can you address the wiki account?

> Could I push back on this item of feedback? The marriage info on this view is really useful especially when dealing with a person with multiple families. The fact that it's available in other views doesn't help me if I'm looking at this view. Having the marriage date visible adds context to the parents' birth and death dates as well as those of the children.

If that is by design, then fine.

> OK. What is the convention if someone has no death info but was born longer ago than the "maximum age probably alive" preference? In that situation should an undated skull be shown, or still nothing? Once you let me know I'll change this to align with the answer to this.

GraphView and FamilyTreeView both do not use probably alive. They only print a death symbol when there is a death event. I suspect being consistent is the best starting position.

ajeb78:
Thanks. It's late now so I'll watch for the wiki account and aim to make the changes later this week (though I'm away for a few days leading up to the weekend)

GaryGriffin:
Published on both gramps60 and gramps61 branches.
Updated template.pot for correct charset (utf-8).

PR #1031, opened 2026-09-20
Add KinshipSort People views for Gramps 6.0

jacek-kuznia:
**User impact:** Users can sort People by biological kinship degree relative to the Home Person, in either a flat list or surname groups.

## Summary

Add KinshipSort, an **EXPERIMENTAL** addon for Gramps 6.0.x, targeting `maintenance/gramps60`. This draft seeks review of the addon and its sorting choices. It does not change Gramps core or publish a Plugin Manager release.

## What to look at

- `kinship_calculation.py`: shortest Birth-only path up to a common ancestor, then down to a relative. A shared child does not make unrelated spouses blood relatives.
- `kinshipsort.py`: two People views, configurable column and full rebuild after person/family changes. Flat ties use generation then name; grouped ties use name. Descending reverses the entire order, including blank degrees. Surname-group rank includes filtered-out members.
- `tests/`: independent graph oracle and real SQLite/GTK model checks; remaining interactive checks are listed in `tests/README.md`.

## Root cause

The normal People views do not expose this calculated degree as a sortable column. This is a new addon proposal, not a fix for a reported upstream defect.

## Fix

Keep calculated values in memory and reuse Gramps People models and view controls. Calculations do not write genealogical records; normal editing remains available. The view integration necessarily uses `gramps.gui` and `gramps.plugins`, so this first submission targets 6.0 only. Pure graph logic is separated from those dependencies.

Include English source strings, Polish translation, GPL-2.0-or-later licensing, attribution of adapted Gramps grouped-view behavior and author/maintainer metadata for Jacek Kuznia.

## Verification

Checked: target branch `maintenance/gramps60` at `af123162e2310e0d5bab0f51c602624ce8c17249`, `make.py:259-293` (catalog compilation and package selection), `tests/__init__.py:1-40` (central GTK pins). KinshipSort is new on this branch; implementation references are therefore on the proposed branch: `KinshipSort/kinship_calculation.py:28-122`, `KinshipSort/kinshipsort.py:91-204`, `KinshipSort/tests/test_gramps_integration.py:139-261`.

On 20 September 2026, from an MSYS2 Bash shell on Windows:

- 15 calculation tests passed with Python 3.10 and 3.14, including 1,050 comparisons against an independent reference and malformed cyclic data.
- 12 integration tests passed with no skips in Gramps AIO 6.0.8, through a CLI-only validation tool in an isolated profile. They use synthetic temporary SQLite databases and real GTK models: relationships, absent/stale Home Person, read-only calculation, sort directions, moved column, normal name sort, updated models, registration and Polish messages.
- GNU `msgfmt --check` passed. Syntax and all 8 translated Python messages were checked. Local review ZIP/TGZ contents match the tested source; generated catalogs, archives and test profiles are excluded from this PR.

These runs use Windows executables launched from MSYS2, not a Linux environment. Interactive window behavior, immediate refresh, Undo/Redo, persisted settings, large-tree responsiveness, Linux/macOS and an official `make.py build` remain unverified. A wiki help page is drafted locally but not published, so no speculative `help_url` is supplied. Keep the PR in draft pending these checks and contributor review.

## Affected addon

New addon **KinshipSort**. Author and current maintainer: **Jacek Kuznia**, @jacek-kuznia, `jacek.kuznia@gmail.com`.

## Longer-term core proposal

I would also like to discuss an optional kinship-degree column in the standard People views in a future Gramps release. That would be a separate design and core-code contribution. Feedback on the reference person, tie-breaking, blank values and caching would help shape it; this addon can serve as the experimental prototype.

## AI assistance and provenance

Publication preparation, calculation extraction, tests, documentation and translations used OpenAI Codex (OpenAI; GPT-6 model family), 19-20 September 2026. The exact model build is not exposed. The request was to prepare the existing KinshipSort addon for Gramps contribution and draft a separate proposal for built-in kinship sorting. Earlier tooling for the supplied 0.2.7 prototype has not been verified and remains for the contributor to clarify. Adapted Gramps code retains attribution in `ATTRIBUTION.md` and source headers.


GaryGriffin:
Thanks for the addon. Couple thoughts

You do not need the `requires_gi` . For unittests, the equivalent of the command below is used. And the `tests/__init__.py` does the import of gi with try/except. Gramps invokes similarly when loading an addon.

 `python3 -m unittest tests.__init__ KinshipSort.tests.test_gramps_integration`

MANIFEST: Generally, the *.md files are not part of the addon tgz file, but if you really want them there, you can include. There are a few other addons that do this.

> I would also like to discuss an optional kinship-degree column in the standard People views in a future Gramps release.

You should start a Discussion at https://github.com/gramps-project/gramps/discussions 

EDIT: Why not use gen/relationship/RelationshipCalculator ?

jacek-kuznia:
Thanks, Gary, for the review and for pointing me to RelationshipCalculator.

**Correction to my earlier reply:** biological siblings recorded in the same
family must receive degree 2 even when neither parent has a person record.
Leaving their degree blank was a bug in KinshipSort, not intended behavior.
I have fixed it and corrected the separate core-feature proposal accordingly.

The calculation now represents each missing parent slot with a family-local
anonymous marker in memory, connected only to that parent's Birth children.
This gives siblings degree 2, their children degree 3 and first cousins
degree 4, and continues through further descendants. It also works when only
one parent is missing. No person records are created or changed. Birth is
checked separately for each parent, and unknown parents in different families
are not assumed to be the same person. Adoption, foster relations and a path
through a shared child to an unrelated spouse still do not create biological
kinship.

I removed `requires_gi` from both view registrations and updated the documented
test command to use `tests.__init__`. I retained the Markdown files in MANIFEST
for offline instructions, attribution and change history.

Regarding `RelationshipCalculator`: I checked `get_relationship_distance_new()`
with `all_families=True` and `only_birth=True`. It supports numerical distances
and recognizes siblings without parents. Its special `s` path has raw rank 1
in that case, so an adapter must count this as two genealogical steps and zero
generation change. This is a conversion requirement, not a reason to exclude
those siblings. Core's configurable search depth and efficient calculation
for the whole People list remain implementation questions. I am open to an
adapter around the core calculator or a shared bulk-distance utility; which
direction would you prefer?

Validation: all 16 calculation tests pass on Python 3.10 and 3.14, and all 17
integration tests pass without skips in Gramps AIO 6.0.8. The new missing-parent
regression cases failed before the fix and pass now. They cover siblings,
descendants, cousins, mixed Birth/adoption/foster links, separate families,
later parent entry, unchanged records and both People models. Tests use
temporary synthetic SQLite databases, with Windows executables launched from
MSYS2. Interactive and cross-platform checks remain pending.

Discussion 2522 in gramps-project/gramps, "Optional biological kinship degree
column in the standard People views", now explicitly includes recorded
biological siblings without parent records and their relatives.

This implementation, testing and response received assistance from OpenAI
Codex (OpenAI, GPT-6 model family; exact build unavailable).

GaryGriffin:
> Core's configurable search depth and efficient calculation
> for the whole People list remain implementation questions. I am open to an
> adapter around the core calculator or a shared bulk-distance utility; which
> direction would you prefer?

What are the comparison timings for each method? How do they scale with tree size? My tree is about 18K people and it was a couple seconds, I think. But others have over 100K people. 

I personally don't think we should have 2 different ways in core to calculate relationship.

dsblank:
I'm planning on a relationship calculator replacement for those database backends that can utilize SQL: https://github.com/dsblank/gramps-sql-extensions

A couple of seconds for 18k people is not bad, but the current algorithm doesn't scale well on larger databases. 

The SQL graph-based algorithm can compute relationships in ~600-630ms for a tree with 100k people; it takes about 9 seconds to do all 100k people, making it feasible to compute the relationship to a person for the whole tree.

jacek-kuznia:
Thanks, that helps.

I agree that having two separate relationship calculation mechanisms in
core would not be ideal.

I'm going to benchmark KinshipSort locally on synthetic trees with 100k+
people. I also want to experiment with your SQL edge-loading approach
combined with a compact cached in-memory graph.

KinshipSort has a slightly different requirement: for sorting the whole
People list it mainly needs a kinship sort key for every person, rather
than a fully formatted relationship for every target. Because of that, I
want to test whether the whole set of sort keys can be calculated in a
single graph traversal after the graph has been built.

I'll first compare the current KinshipSort implementation, your approach,
and the cached/single-pass variant. If the results are useful, I'll share
the timings and code.

Would you prefer that I post the results here, or open an issue in
gramps-sql-extensions?

wt., 22 wrz 2026 o 01:39 Douglas Blank ***@***.***>
napisał(a):

dsblank:
Here is fine. Thanks!

BTW, I understand that you'd want a different response back than the relationship string. We can add other relationship function there too. I already have a few based on the same algorithm, but returning different things (for graphing, bulk, etc).

jacek-kuznia:
Thanks, that is exactly the direction I would like to take.
I do not want KinshipSort to grow into a separate relationship calculator that duplicates your work. I would rather collaborate on a shared graph and relationship engine that both use cases can build on.
Your functions answer questions such as “how are A and B related?”, while KinshipSort needs a different bulk result: for one reference person, calculate the biological kinship degree for every person in the tree.
In KinshipSort, the degree is based only on Birth parent/child links and represents the shortest biological path through a common ancestor. The important part for the People view is to calculate all of those degrees in one bulk operation, rather than calling a pairwise relationship function for every person.
I already have a prototype of that bulk calculation working on an in-memory parent/child graph. I would be happy to adapt it to the graph/engine in gramps-sql-extensions instead of maintaining a separate implementation.
Something conceptually like:
biological_degrees_to(reference_handle)
returning a mapping of person handles to numeric degrees.
If that fits the direction you have in mind, I would like to work on it with you. I am also happy to follow whatever API/design you prefer for the shared graph so that this can eventually be useful beyond KinshipSort and potentially for Gramps core as well.
Would you prefer that I first prototype this directly in gramps-sql-extensions, or discuss the API with you before writing code?

dsblank:
> Would you prefer that I first prototype this directly in gramps-sql-extensions, or discuss the API with you before writing code?

Yes, let's work over there. Now that Gramps 6.1 addons can pip-install packages, you can rely on it directly.

Feel free to make a PR, or an issue, and we can discuss. I think you'll find the infrastructure east to adapt to your needs (and there has already been a request for a similar function--- sortable return values).

jacek-kuznia:
Thanks. That sounds good to me.
I would like to work on this in gramps-sql-extensions rather than develop a separate relationship calculator in KinshipSort.
My goal is to reuse the shared graph and relationship infrastructure, while adding a bulk operation for a different use case: calculating a sortable numeric biological kinship degree for every person relative to one reference person.
I will open an Issue in gramps-sql-extensions first so we can agree on the exact semantics and API before I start changing code. Once we agree on the design, I can work on an implementation and PR there.
Thanks for inviting me to collaborate.

